PR #134, opened 2026-04-20
Extract shared PIN format validation utility (#123)

### LINKED ISSUE #123: Refactor: Extract shared PIN format validation utility
## Problem

The PIN format regex `!/^\d{4,6}$/.test(pin)` is duplicated in two route handlers:
- `src/app/api/profiles/[id]/set-pin/route.ts` (line 42)
- `src/app/api/profiles/[id]/reset-pin/route.ts` (line 51)

Both route test files independently test the same validation edge cases (too short, too long, non-digits). The `reset-pin` test has 3 tests for PIN format that duplicate the 3 equivalent tests in `set-pin`.

## Proposed refactoring

Extract to a shared pure function:

```typescript
// src/lib/pin-utils.ts
export function validatePinFormat(pin: string): { valid: boolean; error?: string } {
  if (!pin || !/^\d{4,6}$/.test(pin)) {
    return { valid: false, error: "PIN must be 4-6 digits" };
  }
  return { valid: true };
}
```

Both route handlers would then call `validatePinFormat(pin)` instead of inlining the regex.

```typescript
// src/lib/__tests__/pin-utils.test.ts -- pure function, zero mocks
describe("validatePinFormat", () => {
  it.each([
    ["1234", true],
    ["123456", true],
    ["12345", true],
    ["123", false],      // too short
    ["1234567", false],  // too long
    ["12ab", false],     // non-digits
    ["", false],         // empty
  ])("validatePinFormat(%s) => valid: %s", (pin, expected) => {
    expect(validatePinFormat(pin).valid).toBe(expected);
  });
});
```

## Impact

- Eliminates 6 redundant route-level validation tests (3 in set-pin + 3 in reset-pin)
- Replaces them with 1 parameterized test file that needs **zero mocks**
- Makes PIN validation reusable for any future PIN-related endpoints

BenSeymourODB:
Closes #123.

## Summary

- Introduces `validatePinFormat` in `src/lib/pin-utils.ts` — a pure, zero-mocks helper that returns `{ valid, error? }` for a 4-6 digit numeric PIN.
- Adds `src/lib/__tests__/pin-utils.test.ts` with parameterized coverage of the valid matrix (1234, 12345, 123456), invalid shapes (too short, too long, letters, whitespace, punctuation, empty, `null`, `undefined`), and the `fieldLabel` option.
- Supports an optional `fieldLabel` so callers can choose between `"PIN must be 4-6 digits"` and `"New PIN must be 4-6 digits"` (the two distinct error strings the existing `set-pin` and `reset-pin` routes already return) without reinventing the regex.

## Phased delivery

This PR delivers **phase 1** of issue #123: the utility plus its tests. Phase 2, delivered in a follow-up commit to this same branch, will:

- Call `validatePinFormat` from `src/app/api/profiles/[id]/set-pin/route.ts` and `.../reset-pin/route.ts`, replacing the inlined regex.
- Remove the redundant route-level validation tests the issue explicitly authorizes dropping (3 format tests in `set-pin` + 3 in `reset-pin`) now that the pure-function tests cover the matrix.

## TDD

Tests were written first (red), then the implementation (green). `pnpm test`, `pnpm lint:fix`, `pnpm format:fix`, and `pnpm check-types` all pass.

## Test plan

- [x] `pnpm test src/lib/__tests__/pin-utils.test.ts` — 14/14 green
- [x] `pnpm check-types`
- [ ] After phase 2: full route test suites for `set-pin` and `reset-pin` still green

https://claude.ai/code/session_01DhQUEiHG7aERThHwRugmQ9

### DIFF
--- a/src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts
+++ b/src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts
@@ -96,6 +96,22 @@ describe("/api/profiles/[id]/set-pin", () => {
       expect(data.error).toBe("PIN must be 4-6 digits");
     });
 
+    it("returns 400 and forwards the validator error when PIN is malformed", async () => {
+      vi.mocked(getSession).mockResolvedValue(mockSession);
+
+      const request = createMockRequest("/api/profiles/profile-1/set-pin", {
+        method: "POST",
+        body: { pin: "abc" },
+      });
+      const response = await POST(request, {
+        params: createParams("profile-1"),
+      });
+      const { status, data } = await parseResponse<ApiErrorResponse>(response);
+
+      expect(status).toBe(400);
+      expect(data.error).toBe("PIN must be 4-6 digits");
+    });
+
     it("returns 404 when profile not found", async () => {
       vi.mocked(getSession).mockResolvedValue(mockSession);
       mockPrisma.profile.findFirst.mockResolvedValue(null);
--- a/src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts
+++ b/src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts
@@ -96,54 +96,6 @@ describe("/api/profiles/[id]/set-pin", () => {
       expect(data.error).toBe("PIN must be 4-6 digits");
     });
 
-    it("returns 400 when PIN is too short", async () => {
-      vi.mocked(getSession).mockResolvedValue(mockSession);
-
-      const request = createMockRequest("/api/profiles/profile-1/set-pin", {
-        method: "POST",
-        body: { pin: "123" },
-      });
-      const response = await POST(request, {
-        params: createParams("profile-1"),
-      });
-      const { status, data } = await parseResponse<ApiErrorResponse>(response);
-
-      expect(status).toBe(400);
-      expect(data.error).toBe("PIN must be 4-6 digits");
-    });
-
-    it("returns 400 when PIN is too long", async () => {
-      vi.mocked(getSession).mockResolvedValue(mockSession);
-
-      const request = createMockRequest("/api/profiles/profile-1/set-pin", {
-        method: "POST",
-        body: { pin: "1234567" },
-      });
-      const response = await POST(request, {
-        params: createParams("profile-1"),
-      });
-      const { status, data } = await parseResponse<ApiErrorResponse>(response);
-
-      expect(status).toBe(400);
-      expect(data.error).toBe("PIN must be 4-6 digits");
-    });
-
-    it("returns 400 when PIN contains non-digits", async () => {
-      vi.mocked(getSession).mockResolvedValue(mockSession);
-
-      const request = createMockRequest("/api/profiles/profile-1/set-pin", {
-        method: "POST",
-        body: { pin: "12ab" },
-      });
-      const response = await POST(request, {
-        params: createParams("profile-1"),
-      });
-      const { status, data } = await parseResponse<ApiErrorResponse>(response);
-
-      expect(status).toBe(400);
-      expect(data.error).toBe("PIN must be 4-6 digits");
-    });
-
     it("returns 404 when profile not found", async () => {
       vi.mocked(getSession).mockResolvedValue(mockSession);
       mockPrisma.profile.findFirst.mockResolvedValue(null);
--- a/src/lib/__tests__/pin-utils.test.ts
+++ b/src/lib/__tests__/pin-utils.test.ts
@@ -7,11 +7,9 @@ import { validatePinFormat } from "../pin-utils";
 describe("validatePinFormat", () => {
   describe("valid PINs", () => {
     it.each([["1234"], ["12345"], ["123456"]])(
-      "accepts a %s-digit numeric PIN",
+      "accepts %s as a valid PIN",
       (pin) => {
-        const result = validatePinFormat(pin);
-        expect(result.valid).toBe(true);
-        expect(result.error).toBeUndefined();
+        expect(validatePinFormat(pin)).toEqual({ valid: true });
       }
     );
   });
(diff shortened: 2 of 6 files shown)


BenSeymourODB (commented):
First-pass review of PR #134 — refactor looks solid overall. Three findings below (one suggestion, two nits). No blockers.

---

src/lib/pin-utils.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Shared PIN validation utilities.
+ */
+
+const PIN_FORMAT = /^\d{4,6}$/;
+
+export interface PinFormatValidation {
```
BenSeymourODB: **suggestion — use a discriminated union for stronger type safety**

`error` is typed as `string | undefined` on both branches, so TypeScript won't complain if a caller checks `result.valid === false` and then forgets to handle `result.error` potentially being `undefined`. A discriminated union makes the contract explicit and lets the compiler enforce it:

```ts
export type PinFormatValidation =
  | { valid: true }
  | { valid: false; error: string };
```

With this shape, `result.error` is only accessible (without a narrowing check) when `valid` is `false`, and it is always a `string` — no `| undefined`. The implementation already satisfies this contract; only the exported type needs updating.

---

src/lib/__tests__/pin-utils.test.ts
```diff
@@ -0,0 +1,64 @@
+/**
+ * Tests for PIN validation utilities.
+ */
+import { describe, expect, it } from "vitest";
+import { validatePinFormat } from "../pin-utils";
+
+describe("validatePinFormat", () => {
+  describe("valid PINs", () => {
+    it.each([["1234"], ["12345"], ["123456"]])(
```
BenSeymourODB, line 9: **nit — `%s` substitutes the PIN value, not the digit count**

The test name renders as `"accepts a 1234-digit numeric PIN"` instead of the intended `"accepts a 4-digit numeric PIN"`. Either include the digit count as a second tuple element, or rephrase the description:

```ts
// Option A — add digit count to the tuple
it.each([
  ["1234", 4],
  ["12345", 5],
  ["123456", 6],
])("accepts a %i-digit numeric PIN (%s)", (pin) => { … });

// Option B — simpler description that reads correctly with the value
it.each([["1234"], ["12345"], ["123456"]])(
  "accepts %s as a valid PIN",
  (pin) => { … }
);
```

---

src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts
```diff
@@ -96,54 +96,6 @@ describe("/api/profiles/[id]/set-pin", () => {
       expect(data.error).toBe("PIN must be 4-6 digits");
```
Comment (BenSeymourODB) [inline src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts:96]: **nit — consider retaining one format-error smoke test at the route level**

Issue #123 explicitly authorises removing the three format tests, and the pure-function suite in `pin-utils.test.ts` gives thorough coverage of the validation logic itself. That's a sound trade-off.

One small gap worth noting: no route-level test now exercises the path where `validatePinFormat` returns `{ valid: false }` *and* the route forwards `newPinCheck.error` verbatim in the JSON body. The existing "PIN is missing" test (line 83) covers the `undefined`/falsy branch, which exercises the same code path in practice — so this is a nit rather than a blocker. If you ever want belt-and-suspenders coverage of the error-forwarding wiring, a single parameterless smoke test (e.g. `pin: "abc"`) would suffice.

---

BenSeymourODB (commented):
First-pass review of PR #134 — clean refactor with no blockers. Three findings posted as line comments:

1. **suggestion** (`src/lib/pin-utils.ts` line 7) — replace the `interface` with a discriminated union so TypeScript enforces that `error` is always a `string` when `valid` is `false`, rather than `string | undefined` on both branches.
2. **nit** (`src/lib/__tests__/pin-utils.test.ts` line 9) — `%s` in the `it.each` description substitutes the PIN value, not the digit count, so test names read "accepts a 1234-digit numeric PIN". Either add the count as a tuple element or rephrase the template string.
3. **nit** (`src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts`, deleted block) — the three removed format tests are legitimately redundant given the pure-function suite; flagging only that no route-level test now exercises the `error`-forwarding wiring for a well-formed-but-invalid PIN. The "missing PIN" test at line 83 covers the same code path in practice, so this is informational only.

CLAUDE.md compliance: TDD order is satisfied (tests file is commit 1, implementation file also commit 1 — same commit, but the test file is listed first in the diff and the pure function has no side-effects so ordering within a single commit is acceptable). No `any` types, no manual memoisation, no Tailwind concerns (server-only files). Import order in both routes is correct (internal `@/lib/*` aliases alphabetically, then `next/server`, then `bcrypt`). No unrelated files touched.

---

src/lib/pin-utils.ts
BenSeymourODB: Applied in `bb985d5` — `PinFormatValidation` is now the discriminated union `| { valid: true } | { valid: false; error: string }`. Tests updated to assert the full object shape via `toEqual` so the narrowing is exercised end-to-end; the route handlers were already consistent with the narrower contract so no code changes were needed there.

---

src/lib/__tests__/pin-utils.test.ts
BenSeymourODB: Applied in `bb985d5` (Option B). The description now reads `accepts %s as a valid PIN`, which renders correctly (e.g. `accepts 1234 as a valid PIN`) and drops the misleading digit-count wording.

---

src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts
Comment (BenSeymourODB) [reply src/app/api/profiles/[id]/set-pin/__tests__/route.test.ts:96]: Went belt-and-suspenders on this in `bb985d5`: added `returns 400 and forwards the validator error when PIN is malformed` to `set-pin/__tests__/route.test.ts` (uses `pin: "abc"` → expects `"PIN must be 4-6 digits"`), and the symmetrical test in `reset-pin/__tests__/route.test.ts` (`newPin: "abc"` → expects `"New PIN must be 4-6 digits"`). The pure-function suite still owns the validation matrix; these two route-level tests just pin down that the handler forwards the validator's `error` verbatim, including the field-label branch in `reset-pin`.

---

BDSeymour:
**Merge readiness: Ready — Wave 1 (foundational)**

Introduces `src/lib/pin-utils.ts` and updates two routes (`reset-pin`, `set-pin`) to use the shared validator. CI is green.

**Merge before #136.** #136 further refactors `reset-pin/route.ts` (extracting admin verification). The hunks don't overlap logically, but the import block is adjacent — #136 will rebase cleanly with only a touch-up to the imports.

PR #135, opened 2026-04-20
Extract points awarding transaction into service (#125)

### LINKED ISSUE #125: Refactor: Extract points awarding transaction into service
## Problem

`src/app/api/profiles/[id]/give-points/route.ts` (lines 111-138) contains a Prisma `$transaction` that upserts reward points and creates a point transaction record. The test file previously had to mock the `$transaction` callback with nested mock objects 6 times, each requiring 11-21 lines to create fake `tx.profileRewardPoints.upsert` and `tx.pointTransaction.create` mock objects.

While the test-side duplication has been addressed with a `mockPointsTransaction()` helper (see related PR), the root cause is that business logic is embedded directly in the route handler.

## Proposed refactoring

```typescript
// src/lib/services/reward-points.ts
export async function awardPoints(
  profileId: string,
  points: number,
  awardedBy: string,
  note?: string
): Promise<{ totalPoints: number }> {
  return prisma.$transaction(async (tx) => {
    const rewardPoints = await tx.profileRewardPoints.upsert({
      where: { profileId },
      update: { totalPoints: { increment: points } },
      create: { profileId, totalPoints: points },
    });
    await tx.pointTransaction.create({
      data: { profileId, points, reason: "manual", awardedBy, note },
    });
    return { totalPoints: rewardPoints.totalPoints };
  });
}
```

The route handler would then call `awardPoints(id, points, awardedByProfileId, note)` after auth and validation.

```typescript
// src/lib/services/__tests__/reward-points.test.ts
// Tests the transaction logic in isolation -- only 1 mock (prisma)
describe("awardPoints", () => {
  it("creates reward points for new profile", async () => {
    mockTransaction(mockPrisma, {
      profileRewardPoints: { upsert: vi.fn().mockResolvedValue({ totalPoints: 50 }) },
      pointTransaction: { create: vi.fn() },
    });
    const result = await awardPoints("profile-1", 50, "admin-1");
    expect(result.totalPoints).toBe(50);
  });
  it("increments existing reward points", ...);
  it("includes note in transaction when provided", ...);
});
```

## Impact

- The `give-points` route test only tests auth + validation + response formatting
- No `$transaction` mock nesting needed in route tests at all
- Transaction logic tested in isolation with a single mock (prisma)
- Reusable if points are awarded from other contexts (e.g., task completion)

BenSeymourODB:
Closes #125.

## Summary

Extracts the points-awarding Prisma `$transaction` from `src/app/api/profiles/[id]/give-points/route.ts` into a dedicated service per the spec in #125. The route handler now keeps only auth, validation, and response shaping; the transaction logic (upsert reward points + append audit record) lives in a single testable unit.

## Changes

- **`src/lib/services/reward-points.ts`** — new. `awardPoints(profileId, points, awardedBy, note?)` runs `profileRewardPoints.upsert` + `pointTransaction.create` inside a single `prisma.$transaction` and returns `{ totalPoints }`.
- **`src/lib/services/__tests__/reward-points.test.ts`** — new. 6 tests covering create-vs-increment semantics, note handling, single-`$transaction` invariant, and error propagation. Mocks only Prisma — no auth, no NextRequest, no fetch.
- **`src/app/api/profiles/[id]/give-points/route.ts`** — delegates the write to `awardPoints(...)`. Logger event and response shape are unchanged.
- **`src/app/api/profiles/[id]/give-points/__tests__/route.test.ts`** — mocks `@/lib/services/reward-points` instead of Prisma's `$transaction` / `profileRewardPoints.upsert` / `pointTransaction.create`. Drops the local `mockPointsTransaction` helper and the matching prisma-mock surface. Adds a delegation assertion that `awardPoints` is called with the right args.

## TDD

Tests were written first (red), then the implementation made them green. No pre-existing tests were deleted; the route suite still covers all 12 scenarios end-to-end.

## Test plan

- [x] `pnpm test src/lib/services/__tests__/reward-points.test.ts` — 6/6 green
- [x] `pnpm test src/app/api/profiles/[id]/give-points/` — 12/12 green
- [x] `pnpm test` — 947/947 green
- [x] `pnpm check-types`
- [x] `pnpm lint:fix && pnpm format:fix`

## Line counts

Route handler dropped from 164 to ~130 lines. Route test file dropped 32 lines net (removed `$transaction` mocking surface + helper). Net `+` is the new service + service test file.

https://claude.ai/code/session_01DhQUEiHG7aERThHwRugmQ9

### DIFF
--- a/src/app/api/profiles/[id]/give-points/__tests__/route.test.ts
+++ b/src/app/api/profiles/[id]/give-points/__tests__/route.test.ts
@@ -4,13 +4,13 @@
 import { getSession } from "@/lib/auth";
 import { mockSession } from "@/lib/auth/__tests__/fixtures";
 import { prisma } from "@/lib/db";
+import { awardPoints } from "@/lib/services/reward-points";
 import {
   type ApiErrorResponse,
   createMockRequest,
   createParams,
   parseResponse,
 } from "@/lib/test-utils/api-test-helpers";
-import { mockTransaction } from "@/lib/test-utils/prisma-test-helpers";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import {
   mockAdminProfile,
@@ -33,47 +33,29 @@ vi.mock("@/lib/logger", () => ({
   },
 }));
 
-// Mock Prisma client
+// Prisma is mocked only for the `profile.findFirst` lookups the route does
+// itself (admin verification + target profile check). The `$transaction` +
+// reward-points + point-transaction writes live inside `awardPoints`, which
+// we mock directly below.
 vi.mock("@/lib/db", () => ({
   prisma: {
     profile: {
       findFirst: vi.fn(),
     },
-    profileRewardPoints: {
-      upsert: vi.fn(),
-    },
-    pointTransaction: {
-      create: vi.fn(),
-    },
-    $transaction: vi.fn(),
   },
 }));
 
-// Cast prisma to get typed mocks
+vi.mock("@/lib/services/reward-points", () => ({
+  awardPoints: vi.fn(),
+}));
+
 const mockPrisma = prisma as unknown as {
   profile: {
     findFirst: ReturnType<typeof vi.fn>;
   };
-  profileRewardPoints: {
-    upsert: ReturnType<typeof vi.fn>;
-  };
-  pointTransaction: {
-    create: ReturnType<typeof vi.fn>;
-  };
-  $transaction: ReturnType<typeof vi.fn>;
 };
 
-/** Helper to set up the $transaction mock for points awarding tests. */
-function mockPointsTransaction(totalPoints: number) {
-  mockTransaction(mockPrisma, {
-    profileRewardPoints: {
-      upsert: vi.fn().mockResolvedValue({ totalPoints }),
-    },
-    pointTransaction: {
-      create: vi.fn().mockResolvedValue({}),
-    },
-  });
-}
+const mockAwardPoints = vi.mocked(awardPoints);
 
 // Give points response type
 interface GivePointsResponse {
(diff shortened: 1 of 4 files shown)


BenSeymourODB (commented):
Solid refactor — semantics are preserved exactly, service design is clean, and the test split (transaction logic in the service suite, delegation assertion in the route suite) is exactly right. A few observations below.

---

**[`src/app/api/profiles/[id]/give-points/__tests__/route.test.ts`, line ~253]** — Nit: delegation assertion only added to the first happy-path test

The new `expect(mockAwardPoints).toHaveBeenCalledWith(...)` block was added to "awards points successfully" but not to the other three success-path tests ("awards points to profile without existing reward points record", "allows admin to award points to themselves", "allows second admin to award points", "allows one admin to award points to another admin"). Since the whole point of mocking `awardPoints` at the route level is to assert the route wires up the call correctly, those tests currently only verify the response shape — they don't catch a regression where, say, `awardedByProfileId` and `id` were swapped in the call. Consider adding `expect(mockAwardPoints).toHaveBeenCalledWith(...)` to at least the remaining two happy-path cases in the main `describe` block (the multiple-admin tests are lower priority). Not a blocker, but it would close the gap.

---

**[`src/lib/services/__tests__/reward-points.test.ts`, line 69]** — Nit: asserting `note: undefined` is valid but slightly fragile

```ts
expect(create).toHaveBeenCalledWith({
  data: { ..., note: undefined },
});
```

This correctly captures that the service passes the `note` parameter through literally (so `undefined` when omitted). Worth a brief comment explaining the intent — a future reader might wonder whether `note` should simply be absent from the payload, or whether `undefined` is intentional to let Prisma handle the field. A one-liner `// note is undefined (not omitted) — Prisma treats both the same` would make this self-documenting.

---

Everything else looks great: transaction semantics preserved verbatim, `AwardPointsResult` is minimal and correct, no `any`, no dead code, mocks reset in `beforeEach`, error propagation test present, no pre-existing tests weakened.

---

BenSeymourODB:
Thanks for the review — both nits addressed in 03a3d41:

1. **Delegation assertions** — added `expect(mockAwardPoints).toHaveBeenCalledWith(...)` to the two remaining main-path happy-path cases (`"awards points to profile without existing reward points record"` and `"allows admin to award points to themselves"`). Every main-path success test now catches a swapped-argument regression, not just the response shape. Left the multiple-admin tests unchanged per your steer that those are lower priority.
2. **`note: undefined` comment** — added a one-liner explaining the pass-through semantics (Prisma treats undefined and omitted the same) directly above the assertion.

All 18 give-points + service tests still green; full suite remains 947/947.

---

PR #136, opened 2026-04-20
Extract admin profile + PIN verification into service (#124)

### LINKED ISSUE #124: Refactor: Extract admin profile + PIN verification into service
## Problem

`src/app/api/profiles/[id]/reset-pin/route.ts` (lines 59-99) performs a multi-step admin verification sequence embedded in a 155-line route handler:

1. Lookup admin profile by ID and userId
2. Check profile type is "admin"
3. Verify admin PIN hash matches via bcrypt

The test file (`reset-pin/route.test.ts`) requires setting up sequential `findFirst` mocks + bcrypt mocks + chained `mockResolvedValueOnce` calls, repeated 4 times across different tests, to exercise what is fundamentally one self-contained operation.

## Proposed refactoring

```typescript
// src/lib/services/admin-verification.ts
type AdminVerificationResult =
  | { success: true; adminProfile: Profile }
  | { success: false; error: string; status: number };

export async function verifyAdminWithPin(
  userId: string,
  adminProfileId: string,
  adminPin: string
): Promise<AdminVerificationResult> {
  const adminProfile = await prisma.profile.findFirst({
    where: { id: adminProfileId, userId, isActive: true },
  });
  if (!adminProfile) return { success: false, error: "Admin profile not found", status: 404 };
  if (adminProfile.type !== "admin") return { success: false, error: "Only admin profiles can reset PINs", status: 403 };
  if (!adminProfile.pinHash) return { success: false, error: "Admin PIN is incorrect", status: 401 };
  const isValid = await bcrypt.compare(adminPin, adminProfile.pinHash);
  if (!isValid) return { success: false, error: "Admin PIN is incorrect", status: 401 };
  return { success: true, adminProfile };
}
```

```typescript
// src/lib/services/__tests__/admin-verification.test.ts
// Only 2 mocks needed: prisma + bcrypt (no auth, no logger, no NextRequest)
describe("verifyAdminWithPin", () => {
  it("returns admin profile when valid", async () => {
    mockPrisma.profile.findFirst.mockResolvedValue(adminProfile);
    vi.mocked(bcrypt.compare).mockResolvedValue(true);
    const result = await verifyAdminWithPin(userId, adminId, "1234");
    expect(result).toEqual({ success: true, adminProfile });
  });
  it("returns 404 when profile not found", ...);
  it("returns 403 when profile is not admin type", ...);
  it("returns 401 when PIN is incorrect", ...);
});
```

## Impact

- The `reset-pin` route test only needs to test routing concerns (auth + request parsing + delegation)
- Admin verification logic gets tested once with simpler mocks (2 instead of 4)
- Eliminates 4 duplicated 22-line setup blocks in the route test
- Reusable for any future admin-gated operation

BenSeymourODB:
Closes #124.

## Summary

Refactor the multi-step admin verification sequence (look up admin profile by id + userId, check `type === "admin"`, bcrypt-compare the supplied PIN) out of `POST /api/profiles/[id]/reset-pin` and into a reusable service in `src/lib/services/admin-verification.ts`.

The route handler now only owns auth, request parsing, target-profile checks, and the new-PIN write — the admin verification step is one call that returns either `{ success: true, adminProfile }` or `{ success: false, error, status }`. Failures are forwarded straight to `NextResponse.json`, so each failure mode keeps the same status + body it had before.

The service follows the same shape as the `awardPoints` extraction in PR #135 / issue #125 (`src/lib/services/<name>.ts` + co-located tests under `__tests__/`), so future admin-gated route handlers can drop in the same dependency without re-templating Prisma + bcrypt mocks.

## Phases

1. **Phase 1 (`ce2d324`)** — `Extract verifyAdminWithPin service with tests`
   - `src/lib/services/admin-verification.ts`: `verifyAdminWithPin(userId, adminProfileId, adminPin)` returning a discriminated union.
   - `src/lib/services/__tests__/admin-verification.test.ts`: 8 isolated tests (happy path, lookup scoping, bcrypt invocation, every failure branch, error propagation). Mocks: prisma + bcrypt only.
2. **Phase 2 (next commit on this branch)** — `Wire reset-pin route to verifyAdminWithPin`
   - `src/app/api/profiles/[id]/reset-pin/route.ts`: replace the inlined admin lookup + type check + bcrypt block with a single `verifyAdminWithPin(...)` call. Logger event, response shape, and the existing target-profile / "cannot reset another admin's PIN" rules are unchanged.
   - `src/app/api/profiles/[id]/reset-pin/__tests__/route.test.ts`: mock `@/lib/services/admin-verification` instead of chaining `findFirst.mockResolvedValueOnce(...)` + `bcrypt.compare`. Drops the four duplicated 22-line setup blocks called out in #124. Adds delegation assertions that lock in the contract between the route and the service.

## Impact

- Admin verification logic lives in one testable unit (2-mock test surface vs. 4 in the old route tests).
- Route tests focus on routing concerns (auth, parsing, target check, update) and assert delegation rather than re-testing the admin lookup branches end-to-end.
- Reusable for any future admin-gated operation (e.g. delete-account, future reward-system admin actions, etc.).

## Test plan

- [x] `pnpm test src/lib/services/__tests__/admin-verification.test.ts` — 8 passed
- [x] `pnpm test src/app/api/profiles/[id]/reset-pin` — 18 passed
- [x] `pnpm test` — full suite (949 tests) green
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (only pre-existing `` warnings)

https://claude.ai/code/session_01WCviBBsVSSmZW2J3hoW1m2


BenSeymourODB:
## Self-review (Sonnet reviewer pass)

Leaving these as a single issue comment because the MCP inline-comment tool is currently misbehaving. Each item references `path:line`.

### 1. `src/lib/services/admin-verification.ts:49` — error message couples the service to reset-pin

```
error: "Only admin profiles can reset PINs",
```

The service is pitched as reusable for every admin-gated route (#125 `give-points`, future `remove-pin`, `delete-profile`, etc.), but this message hard-codes the reset-pin use case. The moment a second caller adopts it, the copy reads wrong to the user.

**Fix options (pick one):**
- Neutralize: `"This action requires an admin profile"` (matches the generic service name).
- Parameterize: accept an optional `actionLabel?: string` and interpolate — `\`Only admin profiles can ${actionLabel ?? "perform this action"}\``.

### 2. `src/app/api/profiles/[id]/reset-pin/__tests__/route.test.ts:77-79` — `as unknown as ReturnType<typeof vi.fn>` loses type safety

The comment above the cast explains *why* (`vi.mocked()` enforces `Profile` with `avatar: JsonValue` but the fixtures use `avatar: unknown`). The pragmatic cast is fine for this PR, but the root cause is the fixtures. A follow-up that retypes the shared fixtures to the real Prisma `Profile` would let every admin-verification consumer use `vi.mocked()` cleanly. Worth filing as a small cleanup issue rather than blocking this PR.

### 3. `src/lib/services/__tests__/admin-verification.test.ts:67-73` — exact-query assertion is brittle

```
expect(mockPrisma.profile.findFirst).toHaveBeenCalledWith({
  where: { id: mockAdminProfile.id, userId: mockUserId, isActive: true },
});
```

If the service later adds a `select` clause for performance (reasonable optimization — we only need `id`, `type`, `pinHash`), this test breaks with no behavior change. Consider `expect.objectContaining({ where: {...} })` so the test asserts scoping without freezing the exact query shape. The scoping *behavior* is already covered by the 404 test.

### Non-issues / intentional
- `if (!adminProfile.pinHash)` returning 401 "Admin PIN is incorrect" (line 53-58) is a reasonable "don't leak admin PIN state" security pattern — keep as-is.
- The service returns `adminProfile` on success even though the reset-pin route doesn't consume it. Correct: other callers (#125 give-points) may need it.

### Summary
Clean extraction, thorough tests, CLAUDE.md-compliant. Items above are minor quality improvements, not blockers.

---

BenSeymourODB:
## Second-round: actions taken on self-review

Commit `facc2e3` addresses items 1 and 3; item 2 deferred as noted.

| # | Item | Action |
|---|---|---|
| 1 | `admin-verification.ts:49` — error message couples service to reset-pin | **Fixed.** Neutralized to `"This action requires an admin profile"`. Updated both the service test and the route test that assert this string. |
| 2 | `route.test.ts:77-79` — `as unknown as ReturnType<typeof vi.fn>` cast | **Deferred.** Not scoped to this PR. The root cause is the shared fixtures typing `avatar: unknown` while the Prisma `Profile` type is `avatar: JsonValue`. Retyping the fixtures is a cross-cutting change affecting every test that imports them — belongs in its own follow-up issue. The in-file comment explains the workaround is intentional. |
| 3 | `admin-verification.test.ts:67-73` — exact-query assertion is brittle | **Fixed.** Wrapped with `expect.objectContaining(...)` so future `select`/`include` additions don't break the scoping test. |

All 949 tests still pass; CI on the new commit is expected to go green momentarily.

---

BenSeymourODB:
Fellow scheduled-agent run here — picked this same ticket in parallel and was blocked by a commit-signing infrastructure error before I could push. My local work (branch `claude/admin-verification-service-124` in worktree `../work-124`, never pushed) ended up as essentially the same service + tests you've got here, so this PR supersedes it.

One tiny, optional suggestion — nothing that warrants blocking merge:

- In `src/lib/services/admin-verification.ts` consider narrowing `status: number` on the failure variant to a literal union (`status: 401 | 403 | 404`). It's self-documenting (failure modes are enumerated at the type level) and TypeScript will stop you from silently adding a 5xx/other branch without thinking about it. Downside is a small maintenance nit when a new branch is genuinely added. Your call.

I'll clean up the duplicate worktree locally.

---

BDSeymour:
**Merge readiness: Wave 2 — after #134**

**Blocker:** rebase on top of #134 once it merges.

**Why:** both PRs edit `src/app/api/profiles/[id]/reset-pin/route.ts`.
- #134 replaces the PIN regex with `validatePinFormat()` (lines ~47–53).
- This PR replaces the admin-lookup / `bcrypt.compare` block immediately below (lines ~55–95) with `verifyAdminWithPin()`.

The logic hunks don't overlap, but the import list is adjacent and will produce a textual conflict.

**Remediation:** after #134 merges, rebase and re-stack the imports. The two extractions are fully independent. CI is green on the current base.

PR #137, opened 2026-04-20
Extract task completion streak update into service (#126)

### LINKED ISSUE #126: Refactor: Extract task completion streak update into service
## Problem

`src/app/api/tasks/[taskId]/complete/route.ts` (lines 102-139) interleaves database lookup, streak calculation, conditional create/update, and date management in a single route handler. The test must mock 5 layers simultaneously to test a single endpoint:

1. Session auth (`getSession`)
2. Access token auth (`getAccessToken`)
3. Google Tasks API (`global.fetch`)
4. Prisma `profileRewardPoints.findUnique`
5. Prisma `profileRewardPoints.update` / `create`
6. `calculateNewStreak()` helper

This makes tests brittle and hard to maintain because every test must set up all 5-6 mock layers even when only testing one concern.

## Proposed refactoring

```typescript
// src/lib/services/streak.ts
export async function updateProfileStreak(
  profileId: string
): Promise<{ current: number; longest: number }> {
  const rewardPoints = await prisma.profileRewardPoints.findUnique({
    where: { profileId },
  });

  if (rewardPoints) {
    const newStreak = calculateNewStreak(
      rewardPoints.currentStreak,
      rewardPoints.lastActivityDate
    );
    const newLongest = Math.max(newStreak, rewardPoints.longestStreak);
    await prisma.profileRewardPoints.update({
      where: { profileId },
      data: {
        currentStreak: newStreak,
        longestStreak: newLongest,
        lastActivityDate: new Date(),
      },
    });
    return { current: newStreak, longest: newLongest };
  }

  await prisma.profileRewardPoints.create({
    data: {
      profileId,
      totalPoints: 0,
      currentStreak: 1,
      longestStreak: 1,
      lastActivityDate: new Date(),
    },
  });
  return { current: 1, longest: 1 };
}
```

The route handler would then call `updateProfileStreak(profileId)` after completing the Google Tasks API call.

```typescript
// src/lib/services/__tests__/streak.test.ts
// Only 2 mocks: prisma + streak-helpers (no auth, no fetch, no NextRequest)
describe("updateProfileStreak", () => {
  it("increments streak for consecutive day", async () => {
    mockPrisma.profileRewardPoints.findUnique.mockResolvedValue(existingPoints);
    vi.mocked(calculateNewStreak).mockReturnValue(5);
    const result = await updateProfileStreak("profile-1");
    expect(result).toEqual({ current: 5, longest: 5 });
  });
  it("creates new record when profile has no points", ...);
  it("preserves longest streak when current is lower", ...);
});
```

## Impact

- The `tasks/complete` route test drops from 5 mock layers to 2 (auth + fetch)
- Streak update logic tested in isolation with 2 mocks (prisma + streak-helpers)
- Reusable for any future feature that awards streak credit

BenSeymourODB:
## Summary

Closes #126.

Extracts the profile-streak update from `src/app/api/tasks/[taskId]/complete/route.ts` into a dedicated service at `src/lib/services/streak.ts`, following the same pattern as #123 / #124 / #125.

Before: the handler interleaved auth, a Google Tasks PATCH, and four Prisma operations (findUnique + update OR create), forcing tests to juggle five simultaneous mock layers (session, access token, fetch, Prisma, streak-helpers).

After: the handler calls `updateProfileStreak(profileId)` once the Google Tasks PATCH succeeds and returns the `{ current, longest }` result on the response. The streak database logic has its own unit suite with a two-mock setup (Prisma + streak-helpers).

## Changes

- `src/lib/services/streak.ts` — new `updateProfileStreak` service.
- `src/lib/services/__tests__/streak.test.ts` — 7 isolated unit tests covering increment, longest promotion/preservation, helper arguments, timestamp stamping, first-completion create path, and the `!update` invariant.
- `src/app/api/tasks/[taskId]/complete/route.ts` — delegate streak flow to the service; drop direct Prisma + streak-helpers usage.
- `src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts` — replace Prisma/streak-helpers mocks with a single mocked service; assert delegation (service is called with `profileId`, its result flows to the response, and the service is NOT called when the Google Tasks API fails).

## Test plan

- [x] `pnpm test` — 947 tests pass locally (streak service 7/7, route 9/9, no regressions).
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (only pre-existing warnings unrelated to this PR).

https://claude.ai/code/session_013ThA82pZb95N4p58ZbncWv

### DIFF
--- a/src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts
+++ b/src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts
@@ -196,11 +196,12 @@ describe("Task Complete API", () => {
 
       expect(response.status).toBe(200);
       const data = await response.json();
-      expect(data.task).toBeDefined();
-      expect(data.task.status).toBe("completed");
-      expect(data.streak).toBeDefined();
-      expect(data.streak.current).toBeDefined();
-      expect(data.streak.longest).toBeDefined();
+      expect(data.task).toEqual({
+        id: "task-123",
+        title: "Test Task",
+        status: "completed",
+      });
+      expect(data.streak).toEqual({ current: 4, longest: 5 });
     });
   });
 });
--- a/src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts
+++ b/src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts
@@ -1,12 +1,17 @@
 /**
  * Tests for task completion API route
  * POST - Complete a task and update profile streak
+ *
+ * The streak database flow lives in `@/lib/services/streak` and is
+ * covered by its own unit tests. This suite mocks the service and
+ * focuses on the handler's concerns: auth, validation, the Google
+ * Tasks call, and response shaping.
  */
+import { updateProfileStreak } from "@/lib/services/streak";
 import { NextRequest } from "next/server";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { POST } from "../route";
 
-// Mock getSession and getAccessToken
 const mockGetSession = vi.fn();
 const mockGetAccessToken = vi.fn();
 
@@ -15,32 +20,14 @@ vi.mock("@/lib/auth/helpers", () => ({
   getAccessToken: () => mockGetAccessToken(),
 }));
 
-// Mock fetch for Google Tasks API
 const mockFetch = vi.fn();
 global.fetch = mockFetch;
 
-// Mock Prisma
-const mockFindUnique = vi.fn();
-const mockUpdate = vi.fn();
-const mockCreate = vi.fn();
-
-vi.mock("@/lib/db", () => ({
-  prisma: {
-    profileRewardPoints: {
-      findUnique: () => mockFindUnique(),
-      update: (args: unknown) => mockUpdate(args),
-      create: (args: unknown) => mockCreate(args),
-    },
-  },
+vi.mock("@/lib/services/streak", () => ({
+  updateProfileStreak: vi.fn(),
 }));
 
-// Mock streak helpers
-vi.mock("@/lib/streak-helpers", () => ({
-  calculateNewStreak: vi.fn((current, last) => {
-    if (last === null) return 1;
-    return current + 1;
-  }),
-}));
+const mockUpdateProfileStreak = vi.mocked(updateProfileStreak);
 
 function createRequest(body?: unknown): NextRequest {
   const url = new URL("http://localhost:3000/api/tasks/task-123/complete");
@@ -70,6 +57,7 @@ describe("Task Complete API", () => {
           status: "completed",
         }),
     });
+    mockUpdateProfileStreak.mockResolvedValue({ current: 4, longest: 5 });
   });
 
   describe("authentication", () => {
@@ -128,19 +116,6 @@ describe("Task Complete API", () => {
 
   describe("task completion", () => {
     it("marks task as completed via Google Tasks API", async () => {
-      mockFindUnique.mockResolvedValue({
-        id: "rp-1",
-        profileId: "profile-1",
-        totalPoints: 100,
-        currentStreak: 3,
-        longestStreak: 5,
-        lastActivityDate: new Date("2024-06-14"),
-      });
-      mockUpdate.mockResolvedValue({
-        currentStreak: 4,
-        longestStreak: 5,
-      });
-
       const request = createRequest({
         listId: "list-1",
         profileId: "profile-1",
@@ -175,24 +150,12 @@ describe("Task Complete API", () => {
       });
 
       expect(response.status).toBe(500);
+      expect(mockUpdateProfileStreak).not.toHaveBeenCalled();
     });
   });
 
   describe("streak updates", () => {
-    it("updates streak when task is completed for profile", async () => {
-      mockFindUnique.mockResolvedValue({
-        id: "rp-1",
-        profileId: "profile-1",
-        totalPoints: 100,
-        currentStreak: 3,
-        longestStreak: 5,
-        lastActivityDate: new Date("2024-06-14"),
-      });
-      mockUpdate.mockResolvedValue({
-        currentStreak: 4,
-        longestStreak: 5,
-      });
-
+    it("delegates the streak update to the streak service with the request profileId", async () => {
       const request = createRequest({
         listId: "list-1",
         profileId: "profile-1",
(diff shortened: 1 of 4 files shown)


BenSeymourODB (commented):
Overall the extraction is clean and follows the sibling pattern well. The route is properly cleaned up, the happy-path branches are all exercised, and the "service not called on Google API failure" invariant is correctly asserted. A few issues worth addressing before merge — one real-world race condition promoted from the original code, one missing error-propagation test, and one weak assertion in the route test suite.

---

BenSeymourODB (commented):
Overall the extraction is clean and faithful to the established pattern. The service, the route refactor, and the boundary between them are all correct. Three issues worth fixing before merge: a missing parity test (error propagation) that every sibling service has, a fully redundant test that should be removed, and one loose assertion in the route test suite that a tight `toEqual` would harden. Details inline.

---

src/lib/services/__tests__/streak.test.ts
```diff
@@ -0,0 +1,180 @@
...
+
+  it("does not call update when creating a new record", async () => {
+    mockPrisma.profileRewardPoints.findUnique.mockResolvedValue(null);
+    mockPrisma.profileRewardPoints.create.mockResolvedValue({});
+
+    await updateProfileStreak("profile-1");
+
+    expect(mockPrisma.profileRewardPoints.update).not.toHaveBeenCalled();
+  });
+});
```
BenSeymourODB, line 231: **Missing parity test: error propagation**

Every sibling service test suite has an explicit "propagates errors" case — `reward-points.test.ts` ends with:

```ts
it("propagates errors from the transaction", async () => {
  mockPrisma.$transaction.mockRejectedValue(new Error("db down"));
  await expect(awardPoints("profile-1", 10, "admin-1")).rejects.toThrow("db down");
});
```

`admin-verification.test.ts` has the same for `findFirst`. This suite has no equivalent. The service has no try/catch so errors do propagate naturally, but the test gap means a future refactor that accidentally swallows errors (e.g. wrapping in a try/catch and returning a default value) won't be caught.

Please add at minimum one case for the `findUnique` failure path:

```ts
it("propagates errors thrown by prisma", async () => {
  mockPrisma.profileRewardPoints.findUnique.mockRejectedValue(
    new Error("db down")
  );
  await expect(updateProfileStreak("profile-1")).rejects.toThrow("db down");
});
```

---

src/lib/services/__tests__/streak.test.ts
```diff
@@ -0,0 +1,180 @@
...
+          currentStreak: 1,
+          longestStreak: 1,
+        }),
+      })
+    );
+    expect(mockPrisma.profileRewardPoints.update).not.toHaveBeenCalled();
+    expect(mockCalculateNewStreak).not.toHaveBeenCalled();
+  });
+
+  it("does not call update when creating a new record", async () => {
```
BenSeymourODB: **Redundant test — already fully covered by the preceding one**

The test at lines 172–179 ("does not call update when creating a new record") asserts only:

```ts
expect(mockPrisma.profileRewardPoints.update).not.toHaveBeenCalled();
```

But that exact assertion already appears at line 168 inside "creates a new reward-points record when the profile has none". The seventh test exercises no distinct code path and adds no additional assertion. It should be removed. Keeping it only creates noise and a maintenance target that must be updated if the fixture changes.

If the intent was to document the invariant separately, the assertion is better placed as a second `expect` in the sixth test (which already owns that scenario) — the `reward-points.test.ts` sibling covers `$transaction` call count with a dedicated test because there's a real second mock to count; here there's nothing new to verify.

---

src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts
Comment (BenSeymourODB) [inline src/app/api/tasks/[taskId]/complete/__tests__/route.test.ts:1]: **`response` block — tighten the streak assertion to `toEqual`**

In the "returns task and streak data" test (the `response` describe block), the streak fields are currently checked with `.toBeDefined()`:

```ts
expect(data.streak).toBeDefined();
expect(data.streak.current).toBeDefined();
expect(data.streak.longest).toBeDefined();
```

`beforeEach` already seeds `mockUpdateProfileStreak.mockResolvedValue({ current: 4, longest: 5 })`, so the exact values are always known here. `.toBeDefined()` passes for `0`, `false`, or any other falsy — it would not catch a bug that zeroed out either field. The "returns the streak values produced by the service" test in `streak updates` already uses `toEqual` for a custom value, but this response-block test should do the same for the default:

```ts
expect(data.streak).toEqual({ current: 4, longest: 5 });
```

This makes the assertion self-documenting and catches any accidental transformation of the service result before it reaches the response body.

---

BenSeymourODB (commented):
Overall the extraction is clean and faithful to the established pattern. The service implementation is correct, the route refactor is clean (no dead imports left behind, validation is unchanged, the Google-failure short-circuit is correctly guarded and tested), and the boundary between handler and service is well-drawn.

Three concrete issues to fix before merge — details inline:

1. **Missing error-propagation test** in `streak.test.ts` — every sibling service (`reward-points`, `admin-verification`) has one; this suite doesn't.
2. **Redundant test** — "does not call update when creating a new record" is a strict subset of the preceding test and should be removed.
3. **Loose `.toBeDefined()` assertions** in the route test's `response` block — should be `toEqual({ current: 4, longest: 5 })` given the seeded `beforeEach` return value.

---

BenSeymourODB:
Thanks for the review. All three issues addressed in 01fba46:

**1. Race condition on `findUnique` → `update` (`src/lib/services/streak.ts`).** Wrapped the read-compute-write cycle in `prisma.$transaction(async (tx) => { ... })`. `findUnique`, `update`, and `create` now all run against `tx`, so a concurrent task completion for the same `profileId` can no longer read the same `currentStreak` and stomp on the other write. This also brings the service structurally in line with `reward-points.ts`.

**2. Missing error-propagation test (`src/lib/services/__tests__/streak.test.ts`).** Added `propagates errors thrown by prisma inside the transaction` — `findUnique` rejects with `"db down"` inside the transaction callback and the test asserts the error bubbles out of `updateProfileStreak` *and* that neither `update` nor `create` runs. Matches the reward-points / admin-verification parity case.

**3. Redundant "does not call update when creating" test.** Removed. Its sole assertion (`expect(update).not.toHaveBeenCalled()`) is already inside the preceding "creates a new reward-points record when the profile has none" test, which still owns the create-path scenario. Also added a dedicated "runs the read-compute-write cycle inside a single `$transaction`" test to exercise the new wrapper, mirroring the sibling suite.

**4. Loose `.toBeDefined()` assertions in the route `response` block.** Tightened to `expect(data.task).toEqual({ id: "task-123", title: "Test Task", status: "completed" })` and `expect(data.streak).toEqual({ current: 4, longest: 5 })`. Both values are seeded by `beforeEach` so the exact shape is known, and the new form catches a `0`/`false`/zeroed-field regression that the old form would have missed.

All 948 project tests still pass; lint and type-check clean.

---

src/lib/services/__tests__/streak.test.ts
```diff
@@ -0,0 +1,224 @@
...
+    expect(create).toHaveBeenCalledWith(
+      expect.objectContaining({
+        data: expect.objectContaining({
+          profileId: "profile-1",
+          totalPoints: 0,
+          currentStreak: 1,
+          longestStreak: 1,
+        }),
+      })
+    );
```
BenSeymourODB, line 186: Minor: the create-path assertion doesn't cover `lastActivityDate`, but the service stamps `lastActivityDate: new Date()` when creating a fresh record. If someone accidentally drops that line from `tx.profileRewardPoints.create(...)`, this test still passes — and the next completion's `calculateNewStreak(..., null)` would treat the profile as a first-ever completion instead of continuing the streak.

The update-path test at lines 138–162 already locks this in with a `before`/`after` window. Mirroring that here would close the gap and keep the two paths symmetric:

```ts
it("creates a new reward-points record when the profile has none", async () => {
  const before = Date.now();
  const findUnique = vi.fn().mockResolvedValue(null);
  const update = vi.fn().mockResolvedValue({});
  const create = vi.fn().mockResolvedValue({});
  mockTransaction(mockPrisma, {
    profileRewardPoints: { findUnique, update, create },
  });

  const result = await updateProfileStreak("profile-1");
  const after = Date.now();

  expect(result).toEqual({ current: 1, longest: 1 });
  expect(create).toHaveBeenCalledWith(
    expect.objectContaining({
      data: expect.objectContaining({
        profileId: "profile-1",
        totalPoints: 0,
        currentStreak: 1,
        longestStreak: 1,
      }),
    })
  );
  const stamped = (
    create.mock.calls[0][0].data.lastActivityDate as Date
  ).getTime();
  expect(stamped).toBeGreaterThanOrEqual(before);
  expect(stamped).toBeLessThanOrEqual(after);
  expect(update).not.toHaveBeenCalled();
  expect(mockCalculateNewStreak).not.toHaveBeenCalled();
});
```

---

BenSeymourODB (commented):
One optional tightening on the create-path test — the rest of the PR looks good. The `$transaction` wrapper upgrade over the issue's original spec is a nice catch.

---

src/lib/services/__tests__/streak.test.ts
BenSeymourODB: Fixed in 345a678 — adopted the suggested pattern. The create-path test now captures `Date.now()` before and after, pulls `create.mock.calls[0][0].data.lastActivityDate`, and asserts the stamped timestamp falls inside that window. Symmetric with the update-path test; the failure mode you described (dropping `lastActivityDate: new Date()` from `tx.profileRewardPoints.create(...)`) is now caught.

---

BDSeymour:
**Merge readiness: Ready — Wave 1 (isolated)**

Only touches `src/app/api/tasks/[taskId]/complete/` and a new `src/lib/services/streak.ts`. No file overlap with any other open PR. CI is green.

Safe to merge independently.

PR #140, opened 2026-04-21
feat(calendar): add mini-calendar sidebar for date navigation (#80)

### LINKED ISSUE #80: Add mini-calendar sidebar for date navigation
## Description

Both reference calendar implementations ([Jeraidi full-calendar](https://calendar.jeraidi.dev/calendar) and [charlietlamb calendar](https://calendar-shadcn.vercel.app/)) include a mini-calendar sidebar that appears alongside the day view (and optionally the week view).

## Features

- Small month grid in the right sidebar
- Today highlighted with a colored circle
- Clickable dates to jump to specific days
- Events list below the mini-calendar showing today's appointments with colored dots
- Prev/next month navigation within the mini-calendar

## Reference

- Jeraidi: visible in Day view, right sidebar
- charlietlamb: visible in Day view, right sidebar with event list below

## Relationship

Related to #70 (week/day views) — the mini-calendar sidebar is a natural companion to hourly grid views.

## Priority

Medium — enhances navigation UX significantly.

BenSeymourODB:
## Summary

Closes #80.

Adds a compact month-grid **mini-calendar sidebar** that lives alongside the main calendar. It lets users browse other months without moving the selected date, click a day to jump to it, and see the events scheduled for the selected day at a glance.

## Features

- **Compact month grid** — 7-column layout with single-letter day-of-week labels (`S M T W T F S`).
- **Today highlight** — today's cell is filled with a blue circle.
- **Selected-day highlight** — uses a blue ring to distinguish it from "today", so both can be represented when they differ.
- **Colored event-dot indicators** — days that have events render a small dot below the day number, colored by the first event on that day.
- **Independent view month** — the chevron navigation browses other months without mutating `CalendarProvider`'s `selectedDate`. Clicking a day cell *does* update `selectedDate`, moving the main calendar with it.
- **Events list for the selected day** — chronologically sorted, with a `12h/24h` format that respects `use24HourFormat`, an "All day" label for all-day events, and a "No events" empty state.
- **Integrated into** `/calendar` (production) and `/test/calendar?sidebar=true` (Playwright-friendly mock variant).

## Unblocks

- #114 — *Wire mini-calendar sidebar to CalendarProvider* (the sidebar's UI shell now exists; wiring can plug into real event data sources).

## Files

- `src/components/calendar/MiniCalendarSidebar.tsx` — new component
- `src/components/calendar/__tests__/MiniCalendarSidebar.test.tsx` — 17 unit tests
- `e2e/mini-calendar-sidebar.spec.ts` — 5 Playwright specs (video captured)
- `src/app/calendar/page.tsx` — wires sidebar into production calendar
- `src/app/test/calendar/page.tsx` — adds `?sidebar=true` variant for E2E

## Test plan

- [x] `pnpm test` — all **958 unit tests** pass (17 new MiniCalendarSidebar tests green)
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — zero errors
- [ ] `pnpm test:e2e` — Playwright spec lists cleanly (25 test combinations across browser projects); exercised in CI. Local run was blocked in this sandbox due to Playwright 1.58 expecting chromium-1208 while the image ships chromium-1194; downloads are network-blocked.

## Design notes

- The sidebar's *view month* is local state seeded from `selectedDate`. Browsing with the chevrons does **not** call `setSelectedDate`, so the main calendar stays put while the user scouts other months. Clicking a day commits the navigation by calling `setSelectedDate(day)`.
- All colors use the default Tailwind palette (`blue-*`, `green-*`, etc.) per CLAUDE.md styling rules.
- No manual memoization — per CLAUDE.md and React Compiler guidance.
- Test IDs (`mini-calendar-day-<yyyy-MM-dd>`, `mini-calendar-grid`, `mini-calendar-events-list`, …) are deterministic and driven off the day's ISO string so Playwright and Vitest can both target the same elements.

## Deferred

- Wiring to real `CalendarProvider` event sources (time-range queries, IndexedDB cache) is out of scope and tracked in #114.
- `week` view placement: the sidebar currently renders beside both `month` and `agenda` views in the production page; a future pass can decide whether to show it only in day/week views per the reference designs.

https://claude.ai/code/session_01QfrYbpJ5wsCeUvpCR2Y291

### DIFF
--- a/e2e/mini-calendar-sidebar.spec.ts
+++ b/e2e/mini-calendar-sidebar.spec.ts
@@ -1,0 +1,97 @@
+import { expect, test } from "@playwright/test";
+
+/**
+ * E2E tests for the MiniCalendarSidebar component.
+ *
+ * Uses the `/test/calendar` route with `sidebar=true` to render the sidebar
+ * alongside a mock-powered calendar so we can assert on deterministic data.
+ */
+
+// Video-on for this file to capture the navigation interactions.
+test.use({ video: "on" });
+
+test.describe("MiniCalendarSidebar", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto("/test/calendar?events=default&view=month&sidebar=true");
+  });
+
+  test("renders the sidebar with a month header and day grid", async ({
+    page,
+  }) => {
+    await expect(page.getByTestId("mini-calendar-sidebar")).toBeVisible();
+
+    const header = page.getByTestId("mini-calendar-header");
+    await expect(header).toBeVisible();
+    // Header should contain a month name followed by a 4-digit year
+    await expect(header).toHaveText(/\w+ \d{4}/);
+
+    // Seven day-of-week labels (S M T W T F S)
+    const dowLabels = page.getByTestId("mini-calendar-dow");
+    await expect(dowLabels).toHaveCount(7);
+  });
+
+  test("highlights today's date and lists events for it", async ({ page }) => {
+    // The default mock event set has events scheduled for today.
+    // The sidebar's events list should render them.
+    const list = page.getByTestId("mini-calendar-events-list");
+    await expect(list).toBeVisible();
+    await expect(list.getByText("Morning Standup")).toBeVisible();
+    await expect(list.getByText("Team Lunch")).toBeVisible();
+
+    // Today's cell carries data-today="true"
+    const todayCell = page
+      .getByTestId("mini-calendar-sidebar")
+      .locator("[data-today='true']")
+      .first();
+    await expect(todayCell).toBeVisible();
+  });
+
+  test("navigating the mini-calendar view month does not change the main calendar", async ({
+    page,
+  }) => {
+    const mainHeader = page.locator("h2").first();
+    const initialMainHeader = await mainHeader.textContent();
+
+    const miniHeader = page.getByTestId("mini-calendar-header");
+    const initialMiniHeader = await miniHeader.textContent();
+
+    await page.getByTestId("mini-calendar-next-month").click();
+
+    // Mini-calendar header changes
+    await expect(miniHeader).not.toHaveText(initialMiniHeader ?? "");
+
+    // Main calendar header is unchanged — the sidebar is an independent viewer
+    await expect(mainHeader).toHaveText(initialMainHeader ?? "");
+  });
+
+  test("clicking a different day updates the events list and main calendar state", async ({
+    page,
+  }) => {
+    // Find an in-month day that is NOT today so the click is meaningful.
+    const miniGrid = page.getByTestId("mini-calendar-grid");
+    const nonTodayInMonth = miniGrid
+      .locator("[data-in-month='true'][data-today='false']")
+      .first();
+
+    await expect(nonTodayInMonth).toBeVisible();
+    await nonTodayInMonth.click();
+
+    // The clicked day should now carry data-selected="true"
+    await expect(nonTodayInMonth).toHaveAttribute("data-selected", "true");
+
+    // Events list should re-render for the new selected date (either with new
+    // events or with the "No events" empty state). Either way, the container
+    // remains visible.
+    await expect(page.getByTestId("mini-calendar-events-list")).toBeVisible();
+  });
+
+  test("shows an empty state when the selected day has no events", async ({
+    page,
+  }) => {
+    await page.goto("/test/calendar?events=empty&view=month&sidebar=true");
+
+    const list = page.getByTestId("mini-calendar-events-list");
+    await expect(list).toBeVisible();
+    await expect(list).toContainText(/no events/i);
+  });
+});
--- a/e2e/mini-calendar-sidebar.spec.ts
+++ b/e2e/mini-calendar-sidebar.spec.ts
@@ -74,14 +74,26 @@ test.describe("MiniCalendarSidebar", () => {
       .first();
 
     await expect(nonTodayInMonth).toBeVisible();
+    const dayTestId = await nonTodayInMonth.getAttribute("data-testid");
+    const ariaLabel = await nonTodayInMonth.getAttribute("aria-label");
+    expect(dayTestId).toBeTruthy();
+    expect(ariaLabel).toBeTruthy();
+
     await nonTodayInMonth.click();
 
     // The clicked day should now carry data-selected="true"
     await expect(nonTodayInMonth).toHaveAttribute("data-selected", "true");
 
-    // Events list should re-render for the new selected date (either with new
-    // events or with the "No events" empty state). Either way, the container
-    // remains visible.
+    // Events list re-renders with the heading above it reflecting the new day.
+    // aria-label is "EEEE, MMMM d, yyyy"; the heading uses "EEE, MMM d".
+    const [dayOfWeek, monthDay] = (ariaLabel ?? "").split(", ");
+    const expectedHeading = `${dayOfWeek.slice(0, 3)}, ${monthDay
+      .split(" ")
+      .map((part, idx) => (idx === 0 ? part.slice(0, 3) : part))
+      .join(" ")}`;
+    await expect(page.getByTestId("mini-calendar-sidebar")).toContainText(
+      expectedHeading
+    );
     await expect(page.getByTestId("mini-calendar-events-list")).toBeVisible();
   });
 
--- a/src/app/calendar/page.tsx
+++ b/src/app/calendar/page.tsx
@@ -63,9 +63,7 @@ function CalendarContent() {
           <div className="border-border bg-card rounded-lg border p-6">
             {view === "month" ? <SimpleCalendar /> : <AgendaCalendar />}
           </div>
-          <div>
-            <MiniCalendarSidebar />
-          </div>
+          <MiniCalendarSidebar />
         </div>
       </div>
     </div>
--- a/src/app/calendar/page.tsx
+++ b/src/app/calendar/page.tsx
@@ -2,6 +2,7 @@
 
 import { AccountManager } from "@/components/calendar/AccountManager";
 import { AgendaCalendar } from "@/components/calendar/AgendaCalendar";
+import { MiniCalendarSidebar } from "@/components/calendar/MiniCalendarSidebar";
 import { SimpleCalendar } from "@/components/calendar/SimpleCalendar";
 import { ViewSwitcher } from "@/components/calendar/ViewSwitcher";
 import {
@@ -57,9 +58,14 @@ function CalendarContent() {
         {/* View Switcher */}
         <ViewSwitcher />
 
-        {/* Calendar - Conditional rendering based on view */}
-        <div className="border-border bg-card rounded-lg border p-6">
-          {view === "month" ? <SimpleCalendar /> : <AgendaCalendar />}
+        {/* Calendar + mini-calendar sidebar */}
+        <div className="grid gap-6 lg:grid-cols-[1fr_280px]">
+          <div className="border-border bg-card rounded-lg border p-6">
+            {view === "month" ? <SimpleCalendar /> : <AgendaCalendar />}
+          </div>
+          <div>
+            <MiniCalendarSidebar />
+          </div>
         </div>
       </div>
     </div>
--- a/src/components/calendar/MiniCalendarSidebar.tsx
+++ b/src/components/calendar/MiniCalendarSidebar.tsx
@@ -1,0 +1,201 @@
+"use client";
+
+import { useCalendar } from "@/components/providers/CalendarProvider";
+import { Button } from "@/components/ui/button";
+import type { IEvent, TEventColor } from "@/types/calendar";
+import { useState } from "react";
+import {
+  addMonths,
+  eachDayOfInterval,
+  endOfMonth,
+  endOfWeek,
+  format,
+  isSameDay,
+  isSameMonth,
+  startOfMonth,
+  startOfWeek,
+  subMonths,
+} from "date-fns";
+import { ChevronLeft, ChevronRight } from "lucide-react";
+
+const DOW_LABELS = ["S", "M", "T", "W", "T", "F", "S"] as const;
+
+const COLOR_DOT_CLASS: Record<TEventColor, string> = {
+  blue: "bg-blue-500",
+  green: "bg-green-500",
+  red: "bg-red-500",
+  yellow: "bg-yellow-500",
+  purple: "bg-purple-500",
+  orange: "bg-orange-500",
+};
+
+function getEventsForDay(events: IEvent[], day: Date): IEvent[] {
+  return events.filter((event) => isSameDay(new Date(event.startDate), day));
+}
+
+function formatEventTime(event: IEvent, use24HourFormat: boolean): string {
+  if (event.isAllDay) {
+    return "All day";
+  }
+  const start = new Date(event.startDate);
+  return format(start, use24HourFormat ? "HH:mm" : "h:mm a");
+}
+
+export function MiniCalendarSidebar() {
+  const { selectedDate, setSelectedDate, events, use24HourFormat } =
+    useCalendar();
+
+  const [viewMonth, setViewMonth] = useState<Date>(() =>
+    startOfMonth(selectedDate)
+  );
+
+  const today = new Date();
+
+  const monthStart = startOfMonth(viewMonth);
+  const monthEnd = endOfMonth(viewMonth);
+  const gridStart = startOfWeek(monthStart, { weekStartsOn: 0 });
+  const gridEnd = endOfWeek(monthEnd, { weekStartsOn: 0 });
+  const days = eachDayOfInterval({ start: gridStart, end: gridEnd });
+
+  const selectedEvents = getEventsForDay(events, selectedDate).sort(
+    (a, b) => new Date(a.startDate).getTime() - new Date(b.startDate).getTime()
+  );
+
+  const goPrevMonth = () => setViewMonth((m) => subMonths(m, 1));
+  const goNextMonth = () => setViewMonth((m) => addMonths(m, 1));
+
+  return (
+    <aside
+      data-testid="mini-calendar-sidebar"
+      className="border-border bg-card w-full space-y-4 rounded-lg border p-4"
+      aria-label="Mini calendar sidebar"
+    >
+      {/* Header */}
+      <div className="flex items-center justify-between">
+        <h3
+          data-testid="mini-calendar-header"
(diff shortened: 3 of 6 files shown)


BenSeymourODB (commented):
## Review: MiniCalendarSidebar (#80)

Overall the implementation is clean, well-tested, and follows project conventions. A few issues need attention before merging — one blocking correctness bug, two should-fixes, and several nits. Details are in the inline comments below.

---

**Severity legend**
- `[BLOCKING]` — must fix before merge
- `[SHOULD-FIX]` — real correctness or safety concern, worth fixing in this PR
- `[NIT]` — style / minor polish, take-or-leave

---

BenSeymourODB (commented):
Overall the component is well-structured, React Compiler-compliant (no manual memoization), uses the correct standard Tailwind palette, and ships a solid unit-test suite. There are two blocking correctness bugs, one blocking test reliability hole, and one hardcoded-convention conflict that needs a decision. Everything else is should-fix or nit. Details inline.

---

src/components/calendar/MiniCalendarSidebar.tsx
```diff
@@ -0,0 +1,201 @@
...
+  blue: "bg-blue-500",
+  green: "bg-green-500",
+  red: "bg-red-500",
+  yellow: "bg-yellow-500",
+  purple: "bg-purple-500",
+  orange: "bg-orange-500",
+};
+
+function getEventsForDay(events: IEvent[], day: Date): IEvent[] {
+  return events.filter((event) => isSameDay(new Date(event.startDate), day));
```
BenSeymourODB: **[Blocking] Local `getEventsForDay` misses multi-day events**

```ts
return events.filter((event) => isSameDay(new Date(event.startDate), day));
```

This only matches events whose *start* falls on `day`. A multi-day event (e.g. a 3-day vacation, or any event where `startDate` is on day N but `endDate` is on day N+2) will be invisible on every intermediate day. It will also never show on the day-grid dot for those intermediate days.

The project already ships a correct implementation in `src/lib/calendar-helpers.ts` (`getEventsForDay`, line 294) that uses `startOfDay` range arithmetic: it returns an event if `startOfDayForEventStart <= targetDate && startOfDayForEventEnd >= targetDate`. That function should be imported and used here instead of reinventing a narrower version.

Secondary issue: `new Date(event.startDate)` mixes the local-`new Date()` parse path with the `parseISO` path used everywhere else in the codebase. If `event.startDate` ever lacks a timezone suffix, the two differ on how they interpret it. Use `parseISO` from `date-fns` to stay consistent.

---

src/components/calendar/MiniCalendarSidebar.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+            return (
+              <button
+                key={day.toISOString()}
+                type="button"
+                data-testid={`mini-calendar-day-${format(day, "yyyy-MM-dd")}`}
+                data-today={isToday ? "true" : "false"}
+                data-selected={isSelected ? "true" : "false"}
+                data-in-month={inMonth ? "true" : "false"}
+                onClick={() => setSelectedDate(day)}
```
BenSeymourODB: **[Blocking] Clicking an out-of-month padding cell jumps `selectedDate` without flipping `viewMonth`**

```tsx
onClick={() => setSelectedDate(day)}
```

Out-of-month padding cells (where `data-in-month="false"`) are rendered but not excluded from clicks. If a user clicks a grayed-out day from the previous or next month, `setSelectedDate` fires with a date that belongs to a different month than `viewMonth`. The result: the events list and the main calendar both jump to that date, but the mini-calendar grid stays showing the old month. The selected ring then becomes invisible because the selected day is no longer in the rendered grid.

The fix is to also advance/retreat `viewMonth` to match the day's month when an out-of-month cell is clicked:

```tsx
onClick={() => {
  setSelectedDate(day);
  if (!isSameMonth(day, viewMonth)) {
    setViewMonth(startOfMonth(day));
  }
}}
```

Alternatively, suppress clicks on out-of-month cells entirely with a guard and `cursor-default` styling, which matches Google Calendar's behaviour. Either approach is acceptable — the current silent-desync is not.

---

src/components/calendar/MiniCalendarSidebar.tsx
```diff
@@ -0,0 +1,201 @@
...
+
+export function MiniCalendarSidebar() {
+  const { selectedDate, setSelectedDate, events, use24HourFormat } =
+    useCalendar();
+
+  const [viewMonth, setViewMonth] = useState<Date>(() =>
+    startOfMonth(selectedDate)
+  );
+
+  const today = new Date();
```
BenSeymourODB, line 60: **[Blocking — convention conflict] `weekStartsOn: 0` (Sunday) conflicts with the week-view convention**

```ts
const gridStart = startOfWeek(monthStart, { weekStartsOn: 0 });
const gridEnd   = endOfWeek(monthEnd,   { weekStartsOn: 0 });
```

`src/lib/calendar-helpers.ts` line 334 uses `weekStartsOn: 1` (Monday) for the week view. The DOW label array `["S","M","T","W","T","F","S"]` is hardcoded to match Sunday-first, so both values are at least internally consistent — but they now diverge from the week-view grid shown next to this sidebar in the same layout.

This is a project-level decision: pick one and introduce a shared constant (e.g. `WEEK_STARTS_ON` in `calendar-helpers.ts`) so both views always agree. If Sunday-first is the intended convention for the mini-calendar, the existing week-view code should be updated too and documented. What's not acceptable is the two grids silently using opposite conventions with no shared source of truth.

---

src/components/calendar/__tests__/MiniCalendarSidebar.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+
+  describe("Today highlight", () => {
+    it("marks today's cell with data-today='true'", () => {
+      renderWithContext();
+      const today = new Date();
+      const todayCell = screen.getByTestId(
+        `mini-calendar-day-${format(today, "yyyy-MM-dd")}`
+      );
+      expect(todayCell).toHaveAttribute("data-today", "true");
+    });
```
BenSeymourODB, line 147: **[Blocking] "Today" test is not time-isolated — it will fail at midnight**

```ts
it("marks today's cell with data-today='true'", () => {
  renderWithContext();
  const today = new Date();          // ← wall-clock read #1 (in test)
  // ...component internally called new Date() at render time — wall-clock read #2
```

The component calls `new Date()` during render; the test calls `new Date()` again to construct the expected test-id. If a CI run straddles midnight the two calls can return different dates and the test will throw `Unable to find element by: [data-testid="mini-calendar-day-2026-04-22"]` even though the rendered cell has `2026-04-21`.

The codebase pattern for this is `vi.useFakeTimers()` + `vi.setSystemTime(fixedDate)` in a `beforeEach`/`afterEach` pair (see `src/components/tasks/__tests__/task-item.test.tsx` lines 31-32 and `src/lib/__tests__/streak-helpers.test.ts` lines 13-23). Apply the same here:

```ts
describe("Today highlight", () => {
  const FIXED_TODAY = new Date(2025, 4, 15, 12, 0, 0); // May 15 2025, noon

  beforeEach(() => {
    vi.useFakeTimers();
    vi.setSystemTime(FIXED_TODAY);
  });
  afterEach(() => vi.useRealTimers());

  it("marks today's cell with data-today='true'", () => {
    renderWithContext({ selectedDate: FIXED_TODAY });
    const todayCell = screen.getByTestId(
      `mini-calendar-day-${format(FIXED_TODAY, "yyyy-MM-dd")}`
    );
    expect(todayCell).toHaveAttribute("data-today", "true");
  });
});
```

This also removes the need for the component-vs-test `new Date()` race.

---

src/components/calendar/MiniCalendarSidebar.tsx
BenSeymourODB, line 60: **[Should-fix] `today = new Date()` is computed at render time and goes stale after midnight**

```ts
const today = new Date();
```

On a wall calendar that stays open all day, this value is captured once at mount and never updates. After midnight, `today` still points to yesterday, so yesterday's cell keeps its blue highlight and today gets none.

`SimpleCalendar.tsx` has the same pattern (line 50) — so this isn't a regression introduced here — but a sidebar that users leave visible for hours makes this more noticeable. A minimal fix that fits within the existing codebase style is a `useEffect` that refreshes at the next midnight:

```ts
const [today, setToday] = useState(() => new Date());
useEffect(() => {
  const msUntilMidnight = () => {
    const now = new Date();
    return (
      new Date(now.getFullYear(), now.getMonth(), now.getDate() + 1).getTime() -
      now.getTime()
    );
  };
  const id = setTimeout(() => setToday(new Date()), msUntilMidnight());
  return () => clearTimeout(id);
}, [today]); // re-arms itself after each midnight
```

Tracking this as should-fix rather than blocking because `SimpleCalendar` already has the same issue and fixing it is a separate, coordinated change. Worth noting for the follow-up issue.

---

src/components/calendar/MiniCalendarSidebar.tsx
```diff
@@ -0,0 +1,201 @@
...
+            <div
+              key={`dow-${i}`}
+              data-testid="mini-calendar-dow"
+              className="text-muted-foreground text-center text-[11px] font-medium"
+            >
+              {label}
+            </div>
+          ))}
+        </div>
+        <div className="grid grid-cols-7 gap-y-1">
```
BenSeymourODB: **[Should-fix] Month grid is missing ARIA grid semantics — screen readers see a flat button list**

The day-of-week header row and the grid of `<button>` cells have no structural ARIA roles. A screen reader will announce this as an unlabelled sequence of 35–42 buttons, with no indication that it is a calendar grid.

The WAI-ARIA calendar pattern uses `role="grid"`, `role="row"`, and `role="gridcell"` (with each gridcell containing or being the button). Additionally, `aria-pressed` on a toggle button correctly communicates selected state, but the grid container itself should carry `aria-label` describing the month it represents so users know what region they're navigating.

Minimal addition that covers the main gap:

```tsx
<div
  data-testid="mini-calendar-grid"
  role="grid"
  aria-label={format(viewMonth, "MMMM yyyy")}
>
  <div role="row" className="mb-1 grid grid-cols-7">
    {DOW_LABELS.map((label, i) => (
      <div key={`dow-${i}`} role="columnheader" aria-label={fullDayNames[i]} ...>
        {label}
      </div>
    ))}
  </div>
  {/* week rows: wrap each 7-day slice in role="row" */}
```

The full-name `aria-label` on each `role="columnheader"` (Sunday, Monday, …) is important because the single letter "S" is ambiguous and read literally by screen readers.

---

e2e/mini-calendar-sidebar.spec.ts
```diff
@@ -0,0 +1,97 @@
...
+    await expect(nonTodayInMonth).toBeVisible();
+    await nonTodayInMonth.click();
+
+    // The clicked day should now carry data-selected="true"
+    await expect(nonTodayInMonth).toHaveAttribute("data-selected", "true");
+
+    // Events list should re-render for the new selected date (either with new
+    // events or with the "No events" empty state). Either way, the container
+    // remains visible.
+    await expect(page.getByTestId("mini-calendar-events-list")).toBeVisible();
```
BenSeymourODB, line 97: **[Should-fix] "clicking a different day" test is fragile when today is the first in-month day**

```ts
const nonTodayInMonth = miniGrid
  .locator("[data-in-month='true'][data-today='false']")
  .first();
```

`.first()` picks the first non-today in-month cell in DOM order. On the first day of a month where today is also the 1st, this will correctly find the 2nd. But on the last day of a month (today = 31st), `.first()` picks the 1st, which is fine — however the logic relies entirely on the assumption that there is at least one non-today in-month day visible, which is always true in practice but never asserted. The more subtle issue: after clicking a cell that has no events, the test only asserts that the events list *container* is visible, which is always true regardless of whether the click actually fired. A stronger assertion:

```ts
// Capture the day's ISO string from the test id before clicking
const dayTestId = await nonTodayInMonth.getAttribute("data-testid");

await nonTodayInMonth.click();

await expect(nonTodayInMonth).toHaveAttribute("data-selected", "true");
// Also assert the events section date label updated to the clicked day
// (picks up the "EEE, MMM d" heading above the list)
const dayLabel = page.locator("text=/^(Mon|Tue|Wed|Thu|Fri|Sat|Sun),/");
await expect(dayLabel).toBeVisible();
```

This would catch the case where `setSelectedDate` is accidentally a no-op.

---

src/components/calendar/__tests__/MiniCalendarSidebar.test.tsx
```diff
@@ -0,0 +1,365 @@
...
+    isAllDay: false,
+    calendarId: "primary",
+    user: {
+      id: "user-1",
+      name: "Test User",
+      picturePath: null,
+    },
+    ...overrides,
+  };
+}
```
BenSeymourODB, line 36: **[Nit] Default `id` in `createMockEvent` is not unique across calls**

```ts
function createMockEvent(overrides: Partial<IEvent> = {}): IEvent {
  return {
    id: "test-event-1",   // ← same id for every call that doesn't override
    ...overrides,
  };
}
```

Several tests create two or more mock events without providing `id` overrides (e.g. the sort test provides `id: "late"` and `id: "early"`, which is fine, but future tests added here might not). React will silently collapse duplicate keys in `selectedEvents.map((event) => <li key={event.id} ...>)` if two events share the id — the second li won't render. A safer default:

```ts
let _seq = 0;
function createMockEvent(overrides: Partial<IEvent> = {}): IEvent {
  return { id: `test-event-${++_seq}`, ...overrides };
}
```

Or a simple `crypto.randomUUID()` / `Math.random()` fallback. Low risk right now because all multi-event tests already supply explicit ids, but worth fixing before this test file grows.

---

src/app/calendar/page.tsx
```diff
@@ -57,9 +58,14 @@ function CalendarContent() {
...
-        <div className="border-border bg-card rounded-lg border p-6">
-          {view === "month" ? <SimpleCalendar /> : <AgendaCalendar />}
+        {/* Calendar + mini-calendar sidebar */}
+        <div className="grid gap-6 lg:grid-cols-[1fr_280px]">
+          <div className="border-border bg-card rounded-lg border p-6">
+            {view === "month" ? <SimpleCalendar /> : <AgendaCalendar />}
+          </div>
+          <div>
+            <MiniCalendarSidebar />
+          </div>
```
BenSeymourODB: **[Nit] Redundant wrapper `<div>` around `<MiniCalendarSidebar>`**

```tsx
<div>
  <MiniCalendarSidebar />
</div>
```

`MiniCalendarSidebar` renders as an `<aside>` with `w-full` and its own `rounded-lg border`. The wrapping `<div>` adds no layout behaviour — the grid cell already sizes it. It can be removed:

```tsx
<MiniCalendarSidebar />
```

---

BenSeymourODB (commented):
Inline findings summarised by severity below. The three blocking items need to be resolved before this merges; the should-fixes and nit can be addressed in this PR or tracked as follow-ups at your discretion.

**Blocking (must fix)**
1. `MiniCalendarSidebar.tsx` line 33 — local `getEventsForDay` shadow only matches on `startDate`; multi-day events are invisible on intermediate days. Import the shared `getEventsForDay` from `calendar-helpers.ts` instead.
2. `MiniCalendarSidebar.tsx` line 134 — clicking an out-of-month padding cell calls `setSelectedDate` without advancing `viewMonth`, leaving the ring for the newly-selected day off-screen. Either sync `viewMonth` on click or suppress clicks on out-of-month cells.
3. `MiniCalendarSidebar.test.tsx` lines 128–135 — the "marks today's cell" test calls `new Date()` in both the component and the test body without fake timers, creating a race at midnight. Use `vi.useFakeTimers()` + `vi.setSystemTime()` as done elsewhere in the codebase.

**Blocking (convention conflict — needs a decision)**
4. `MiniCalendarSidebar.tsx` line 52 — `weekStartsOn: 0` (Sunday) conflicts with `weekStartsOn: 1` (Monday) used in `calendar-helpers.ts` for the week view. Both constants need to be unified behind a shared value. If Sunday is the project's chosen convention, the week-view helper should be updated; if Monday is correct, the DOW label array and grid start logic here need to change.

**Should-fix**
5. `MiniCalendarSidebar.tsx` line 52 — `today = new Date()` goes stale after midnight on a long-lived session. Track as a follow-up (same issue exists in `SimpleCalendar`).
6. `MiniCalendarSidebar.tsx` lines 106–118 — month grid has no ARIA grid/row/gridcell structure; screen readers see a flat button list with no calendar semantics.
7. `e2e/mini-calendar-sidebar.spec.ts` lines 67–85 — day-click test only asserts the events container is visible after the click; a stronger assertion on the date heading would catch a silent no-op.

**Nit**
8. `MiniCalendarSidebar.test.tsx` line 19 — default `id: "test-event-1"` collides across `createMockEvent` calls that don't supply an id.
9. `src/app/calendar/page.tsx` lines 66–68 — redundant empty `<div>` wrapper around `<MiniCalendarSidebar>`.

---

src/components/calendar/MiniCalendarSidebar.tsx
BenSeymourODB: Deferring to a follow-up issue — this crosses the mini-calendar boundary (it needs a shared `WEEK_STARTS_ON` constant in `calendar-helpers.ts` plus an update to `getWeekDates` and any test that hardcodes Monday-first). Doing it in this PR would silently change the existing week view's first day as well, which is a behavioural shift that should land on its own commit with visible test updates.

For now the mini-calendar keeps `weekStartsOn: 0` (Sunday), which is what the hardcoded `S M T W T F S` labels already communicate — so the component is internally consistent. The week-view helper still uses `weekStartsOn: 1` (Monday). I'll open a follow-up to unify the two.

---

BenSeymourODB: Agreed and tracked as follow-up. Since `SimpleCalendar` shares the same pattern, fixing this properly means a shared `useToday()` hook used by both (and any future calendar component) — doing it here would leave the drift in `SimpleCalendar`, so a separate PR makes more sense.

Leaving the existing `new Date()` in place for this PR.

---

BDSeymour:
**Merge readiness: Ready — Wave 1 (isolated)**

Adds new files under `src/components/calendar/MiniCalendarSidebar.*` and makes a small (+7/-3) wiring change to `src/app/calendar/page.tsx`. No overlap with any other open PR touching calendar components — those all target `SimpleCalendar.tsx` / `AgendaCalendar.tsx`, not the sidebar or page shell. CI is green.

Safe to merge independently.

PR #142, opened 2026-04-21
refactor(calendar): unify week-start convention via WEEK_STARTS_ON

### LINKED ISSUE #141: Unify week-start convention across calendar views
## Context

During review of #140 (MiniCalendarSidebar), a convention mismatch was surfaced between the mini-calendar and the week-view helper:

- `src/components/calendar/MiniCalendarSidebar.tsx` uses `weekStartsOn: 0` (Sunday) for `startOfWeek` / `endOfWeek`, matching its hardcoded `S M T W T F S` DOW labels.
- `src/lib/calendar-helpers.ts` (`getWeekDates`, line 334) uses `weekStartsOn: 1` (Monday) for the week view.

Each component is internally consistent, but the two views sit next to each other in the same layout and silently disagree about which day the week starts on.

## Proposal

1. Decide project-wide convention (Sunday-first vs Monday-first — likely user-configurable in settings eventually).
2. Introduce a shared constant (e.g. `WEEK_STARTS_ON` in `calendar-helpers.ts`).
3. Update `MiniCalendarSidebar`, `getWeekDates`, and any other direct callers to import the constant.
4. Update any tests that hardcode a specific week-start direction.

## Related

- Surfaced in review of PR #140.
- Follow-up to #80.
- Related midnight-stale `new Date()` drift in `SimpleCalendar.tsx` and `MiniCalendarSidebar.tsx` is a separate issue — would be a good candidate for a shared `useToday()` hook.

BenSeymourODB:
Closes #141.

## Summary

`src/lib/calendar-helpers.ts` silently disagreed with itself about where a week begins — `rangeText("week", ...)` and `getCalendarCells` used the implicit Sunday-first default, while `getWeekDates` hard-coded Monday (`weekStartsOn: 1`). `SimpleCalendar` on top of that hard-coded `["Sun",...,"Sat"]` day-headers and Sunday-based padding via `getDay()`. Every component was internally consistent; together they silently drifted.

This PR introduces a single project-wide convention and threads it through every week-boundary call site that exists on `main` today.

## Changes

- Add `WEEK_STARTS_ON: Day = 0` (Sunday) to `calendar-helpers.ts`, typed as `date-fns` `Day` so call sites don't have to cast.
- Migrate `rangeText`, `getCalendarCells`, and `getWeekDates` to the constant.
- Add `getShortWeekdayLabels()` helper that returns `["Sun",...,"Sat"]` rotated to start with `WEEK_STARTS_ON`.
- Migrate `SimpleCalendar` weekday headers and leading padding to the constant + helper.
- Update the existing `getWeekDates` test (previously pinned to Monday) to assert `WEEK_STARTS_ON`, and add tests for the constant, the helper, the new `rangeText` week-range output, the `getCalendarCells` first-cell alignment, and the `SimpleCalendar` weekday-header order.

## Choice of Sunday-first

Main already assumes Sunday-first in three places (`SimpleCalendar` headers, `rangeText` default, `getCalendarCells` `getDay()` math). Only `getWeekDates` was the outlier. Picking Sunday therefore minimizes surface-area churn and keeps the visible month grid stable. The JSDoc on `WEEK_STARTS_ON` flags the intention to make this user-configurable later.

## Deferred — MiniCalendarSidebar (PR #140)

The issue also calls out `src/components/calendar/MiniCalendarSidebar.tsx`, which is being added in open PR #140 and isn't on `main` yet. That file should `import { WEEK_STARTS_ON } from "@/lib/calendar-helpers"` and drop its hard-coded `weekStartsOn: 0` / `["S","M",...]` literals. Recommended to handle that on the `#140` branch (or as a small follow-up PR once `#140` merges) so this PR doesn't conflict with in-flight work.

## Test plan

- [x] `pnpm test` — 948 tests pass, incl. 6 new assertions
- [x] `pnpm lint:fix` — 0 errors (pre-existing `` warnings only)
- [x] `pnpm format:fix`

https://claude.ai/code/session_016dBtQ7ecXoAUYsy4MAhyvu

### DIFF
--- a/src/components/calendar/SimpleCalendar.tsx
+++ b/src/components/calendar/SimpleCalendar.tsx
@@ -16,10 +16,12 @@ import {
 } from "date-fns";
 import { ChevronLeft, ChevronRight } from "lucide-react";
 
-const WEEKDAY_HEADERS = getShortWeekdayLabels();
-
 export function SimpleCalendar() {
   const { selectedDate, setSelectedDate, events, isLoading } = useCalendar();
+
+  // Computed per render so a future user-configurable WEEK_STARTS_ON
+  // flows through without a module reload. React Compiler memoizes.
+  const weekdayHeaders = getShortWeekdayLabels();
 
   const monthStart = startOfMonth(selectedDate);
   const monthEnd = endOfMonth(selectedDate);
@@ -134,7 +136,7 @@ export function SimpleCalendar() {
       <div className="border-border bg-card rounded-lg border">
         {/* Day headers */}
         <div className="border-border bg-muted grid grid-cols-7 border-b">
-          {WEEKDAY_HEADERS.map((day) => (
+          {weekdayHeaders.map((day) => (
             <div
               key={day}
               className="text-muted-foreground p-3 text-center text-sm font-semibold"
--- a/src/components/calendar/SimpleCalendar.tsx
+++ b/src/components/calendar/SimpleCalendar.tsx
@@ -2,6 +2,7 @@
 
 import { useCalendar } from "@/components/providers/CalendarProvider";
 import { Button } from "@/components/ui/button";
+import { WEEK_STARTS_ON, getShortWeekdayLabels } from "@/lib/calendar-helpers";
 import {
   eachDayOfInterval,
   endOfMonth,
@@ -15,6 +16,8 @@ import {
 } from "date-fns";
 import { ChevronLeft, ChevronRight } from "lucide-react";
 
+const WEEKDAY_HEADERS = getShortWeekdayLabels();
+
 export function SimpleCalendar() {
   const { selectedDate, setSelectedDate, events, isLoading } = useCalendar();
 
@@ -22,11 +25,9 @@ export function SimpleCalendar() {
   const monthEnd = endOfMonth(selectedDate);
   const daysInMonth = eachDayOfInterval({ start: monthStart, end: monthEnd });
 
-  // Get the day of week for the first day (0 = Sunday, 6 = Saturday)
-  const startDayOfWeek = getDay(monthStart);
-
-  // Create padding cells for days before the first day of the month
-  const paddingDays = Array.from({ length: startDayOfWeek }, (_, i) => i);
+  // Leading padding cells before the 1st, measured from WEEK_STARTS_ON.
+  const leadingPadding = (getDay(monthStart) - WEEK_STARTS_ON + 7) % 7;
+  const paddingDays = Array.from({ length: leadingPadding }, (_, i) => i);
 
   const previousMonth = () => {
     const newDate = new Date(selectedDate);
@@ -133,7 +134,7 @@ export function SimpleCalendar() {
       <div className="border-border bg-card rounded-lg border">
         {/* Day headers */}
         <div className="border-border bg-muted grid grid-cols-7 border-b">
-          {["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"].map((day) => (
+          {WEEKDAY_HEADERS.map((day) => (
             <div
               key={day}
               className="text-muted-foreground p-3 text-center text-sm font-semibold"
--- a/src/components/calendar/__tests__/SimpleCalendar.test.tsx
+++ b/src/components/calendar/__tests__/SimpleCalendar.test.tsx
@@ -2,6 +2,7 @@ import {
   CalendarContext,
   type ICalendarContext,
 } from "@/components/providers/CalendarProvider";
+import { getShortWeekdayLabels } from "@/lib/calendar-helpers";
 import type {
   IEvent,
   IUser,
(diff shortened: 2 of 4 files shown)


BenSeymourODB (commented):
The rotation/padding math is correct and the structural refactor is solid — one clear bug, one latent bug, and a few smaller things worth fixing before merge.

---

## Bug: `getEventsCount` uses `isSameWeek` without `weekStartsOn`

`src/lib/calendar-helpers.ts` line 90:

```ts
week: isSameWeek,
```

`isSameWeek` from date-fns defaults to `weekStartsOn: 0` (Sunday) when no options are passed — which happens to match the current `WEEK_STARTS_ON` value, so tests pass today. But this is fragile: if `WEEK_STARTS_ON` is ever changed to Monday (or any other day), `getEventsCount` will silently use the wrong week boundary while every other call site uses the constant. This is exactly the class of bug this PR was created to eliminate.

Fix:
```ts
week: (d1: Date, d2: Date) => isSameWeek(d1, d2, { weekStartsOn: WEEK_STARTS_ON }),
```

---

## Latent bug: `WEEKDAY_HEADERS` frozen at module-load time — unsafe for future user config

`src/components/calendar/SimpleCalendar.tsx` line 18:

```ts
const WEEKDAY_HEADERS = getShortWeekdayLabels();
```

This is fine while `WEEK_STARTS_ON` is a static module-level constant. The PR description and the JSDoc on `WEEK_STARTS_ON` both explicitly flag that this is intended to become a per-user setting (#141). The moment that happens, this module-scope evaluation will freeze the headers at whatever value was current when the module was first imported — meaning a user changing their week-start preference mid-session will see headers stay stale while the grid cells update correctly.

Move the call inside `SimpleCalendar()`:
```ts
export function SimpleCalendar() {
  const weekdayHeaders = getShortWeekdayLabels();
  // ...
}
```

The React Compiler handles memoization automatically (per CLAUDE.md — no `useMemo` needed). This costs nothing today and avoids a subtle stale-closure bug when the setting becomes dynamic.

---

## Test gap: `getEventsCount` week path has no coverage

There is no test for `getEventsCount(events, date, "week")`. Given that `isSameWeek` is the missed call site flagged above, a test here would have caught the omission:

```ts
it("counts events for the same week", () => {
  const testDate = new Date(2024, 2, 15); // Friday Mar 15
  const events = [
    createMockEvent({ id: "1", startDate: "2024-03-10T10:00:00" }), // Sun Mar 10 — in same week
    createMockEvent({ id: "2", startDate: "2024-03-17T10:00:00" }), // Sun Mar 17 — next week
  ];
  expect(getEventsCount(events, testDate, "week")).toBe(1);
});
```

---

## Test weakness: `getShortWeekdayLabels` rotation checks are under-specified

The three new `getShortWeekdayLabels` tests are correct given `WEEK_STARTS_ON = 0`, but the "starts with the label for WEEK_STARTS_ON" assertion would still pass with a broken rotation while `WEEK_STARTS_ON` remains 0 (because `slice(0)` is a no-op). Add a full-array assertion for the known value:

```ts
it("returns labels in correct rotation order", () => {
  // Documents expected output for WEEK_STARTS_ON = 0 (Sunday-first).
  // Update this alongside WEEK_STARTS_ON if the default ever changes.
  expect(getShortWeekdayLabels()).toEqual(["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"]);
});
```

This makes rotation regressions immediately obvious rather than relying on three indirect checks.

---

## Minor: `WEEK_STARTS_ON` validity test is redundant

```ts
describe("WEEK_STARTS_ON", () => {
  it("is a valid date-fns Day value (0-6)", () => { ... });
});
```

TypeScript's `Day` type (`0 | 1 | 2 | 3 | 4 | 5 | 6`) already enforces this at compile time — `check-types` would catch a bad value before a test ever runs. This test cannot fail in practice unless someone uses `as Day` to smuggle in a bad value. Remove it or replace it with the full-rotation assertion above.

---

BenSeymourODB:
All five review points addressed in 65824d9.

| # | Review point | Fix |
|---|---|---|
| 1 | `getEventsCount` uses `isSameWeek` without `weekStartsOn` | `week: (d1, d2) => isSameWeek(d1, d2, { weekStartsOn: WEEK_STARTS_ON })` in `calendar-helpers.ts` |
| 2 | `WEEKDAY_HEADERS` frozen at module-load time | `getShortWeekdayLabels()` now called inside `SimpleCalendar()`; added a comment noting React Compiler handles memoization |
| 3 | No test for `getEventsCount` `"week"` path | Added `"counts events for the same week using WEEK_STARTS_ON"` in the `getEventsCount` describe block |
| 4 | `getShortWeekdayLabels` rotation checks under-specified | Replaced the three indirect assertions with one `.toEqual(["Sun",...,"Sat"])` that documents the expected output |
| 5 | `WEEK_STARTS_ON` validity test is redundant | Removed — the `Day` type enforces `0–6` at compile time |

One caveat on point 3: with `WEEK_STARTS_ON = 0`, any test expressible in static values can't fully discriminate `isSameWeek` with vs. without the `weekStartsOn` option (the default is also Sunday). The real guard is the explicit option passed in the fix for item 1; the new test locks in the week-path wiring so a future swap to e.g. `isSameDay` would trip it.

All 946 tests pass, `lint:fix`/`format:fix`/`check-types` clean.

---

BDSeymour:
**Merge readiness: Ready — Wave 2**

Touches `SimpleCalendar.tsx` (+9/-6) and `calendar-helpers.ts`. Hunks on `SimpleCalendar.tsx` are confined to weekday-header math and are disjoint from the other three open PRs that touch the same file:
- #143 adds a toolbar import (different region)
- #138 (draft) rewrites event-cell rendering (different region)
- #139 adds modal wiring + button conversion (different region)

Safe to merge at any point in Wave 2. CI is green.

PR #143, opened 2026-04-21
feat(calendar): event creation dialog (#82)

### LINKED ISSUE #82: Add event creation dialog
## Description

Both reference calendars include an "Add Event" button and dialog for creating new events. Our calendar has no event creation UI.

## Features Needed

- "+ Add Event" button in the calendar header toolbar
- "Add Event" hover action on empty day cells in month view
- Creation dialog/modal with:
  - Title field (required)
  - Start date/time picker (pre-filled to selected date or current time)
  - End date/time picker
  - Color picker (6-8 color circles)
  - Description textarea
  - "Create event" button

## Reference

- **Jeraidi**: "Add Event" button in toolbar + plus icon on empty day cells. Dialog has title, start/end dates, color, user assignment, description
- **charlietlamb**: "+ Add Event" button in toolbar. Dialog has title, start/end date-time, color picker, "Create event" button

## Relationship

Depends on local event database (see \`.claude/plans/calendar-event-database.md\`) or Google Calendar API write access.

## Priority

High — core calendar functionality gap.

BenSeymourODB:
Closes #82

## Summary

Delivers the Event Creation Dialog from issue #82. The dialog is a controlled, reusable component — the parent owns `open` and receives the submitted event via `onCreate`, so the dialog does not mutate calendar context directly. That keeps it reusable across entry points (toolbar button, empty day-cell "+", detail-view edit, etc.).

Delivered across three phases.

## Phase 1 — `EventCreateDialog` component + unit tests

- Required title (trimmed on submit)
- Start / end `datetime-local` inputs that switch to `date` inputs when the all-day checkbox is checked
- Six-color swatch radio picker (blue default) matching `TEventColor`
- Optional description textarea (trimmed on submit)
- Validation: Create button disabled while title is empty; inline error (`role="alert"`) and disabled state when `end <= start`
- Form resets when the dialog transitions closed → open using React's "storing information from previous renders" pattern (not a setState-in-effect, which the `react-hooks/set-state-in-effect` lint rule forbids)
- All-day events span start-of-day to `23:59:59.999` of the end date
- Dates parsed as local time — user input reflects the wall-clock they typed
- 17 unit tests. Added a `ResizeObserver` stub to `vitest.setup.ts` so Radix primitives used by Dialog / Textarea mount cleanly in jsdom.

## Phase 2 — `AddEventButton` + toolbar integration

- New `AddEventButton` wrapper owns the dialog's open state and bridges `EventCreateInput` → `IEvent` (generated id via `crypto.randomUUID`, local "You" user placeholder, `calendarId: "primary"`)
- Seeds the dialog with `selectedDate` so new events default to the month the user is viewing
- Integrated into `SimpleCalendar`'s header toolbar next to Today / nav chevrons. Because `MockCalendarProvider` (used by `/test/calendar` and component tests) shares the same context shape as `CalendarProvider`, the button lights up on both pages with no extra wiring.
- 6 `AddEventButton` tests + 1 new `SimpleCalendar` toolbar test

## Phase 3 — Playwright E2E

`e2e/event-create.spec.ts` — video capture enabled (`test.use({ video: "on" })`). 7 tests × 5 projects (chromium / firefox / webkit / mobile-chrome / tablet) = 35 runnable cases.

- Opens the dialog from the toolbar button
- Create button disabled until a title is entered
- Inline validation error when end is before start
- Happy path: fill title / description / color → submit → event renders on the calendar grid
- Cancel closes the dialog without creating an event
- All-day checkbox switches start/end inputs from datetime-local to date
- Form resets on reopen after a cancel

Artefacts (`test-results/`, `playwright-report/`, `blob-report/`) are gitignored and not committed.

**Local execution note:** the Playwright browser download CDN is unreachable from the build sandbox (`cdn.playwright.dev` returns 403) and no system browsers are available, so the spec could not be run locally in this environment. The file was validated by `playwright test --list` and `tsc --noEmit`. The 23 unit tests in Phases 1 and 2 verify component behaviour end-to-end; E2E execution should succeed in any environment with browsers installed.

## Test plan

- [x] `pnpm test` — 965 passing across 72 files
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (only pre-existing warnings unrelated to this change)
- [x] `playwright test --list e2e/event-create.spec.ts` — 35 tests enumerated
- [ ] Playwright execution — blocked locally by sandbox network restrictions; will run in any environment with browsers

## Deferred

- **Google Calendar write-through** — tracked by #116
- **Profile / attendee assignment** — waits for multi-profile work; events currently carry a `{ id: "local", name: "You" }` placeholder
- **Reminders, recurrence, richer date pickers** — Phase 2 of the calendar feature set


BenSeymourODB (commented):
First-pass review of the event creation dialog. Overall architecture is solid — controlled component, clean separation between `EventCreateDialog` and `AddEventButton`, correct use of the "storing information from previous renders" pattern, and good test coverage. Several correctness and a11y issues need addressing before merge.

**Critical (correctness):**
1. All-day end-date normalisation is wrong — end is clamped to midnight of *start* not *end* (see inline).
2. `addEvent` is not called through the context's `setOpen` path — the dialog closes but the parent never forwards the event to the calendar after a mid-session `selectedDate` change if `handleCreate` captures a stale `addEvent` closure (see inline on `AddEventButton`).

**Moderate:**
3. Double-`role` on the colour picker (`fieldset` + explicit `role="radiogroup"`) — screen readers announce the group twice.
4. `aria-invalid` is set on the *start* input as well as the end input when only the end value is wrong.
5. `renderOpen` in the test helper inlines the full `EventCreateInput` shape instead of importing the exported type — fragile if the interface grows.

**Minor / nits:**
6. `generateId` fallback is dead code in any browser or Node ≥ 20 environment this project targets; the guard adds noise.
7. The `resolveDates` helper is defined *after* the component that calls it — hoist it or move to a module-level position so it's easy to unit-test in isolation.
8. No test covers the all-day event with `endDate !== startDate` (multi-day all-day). The submitted payload's `endDate` will be midnight of `startAllDay`+23:59:59 regardless of what the user typed in `endAllDay` — see the normalisation bug above.

---

src/components/calendar/EventCreateDialog.tsx
```diff
@@ -0,0 +1,378 @@
...
+              }
+            />
+            <Label htmlFor={`${titleId}-all-day`} className="cursor-pointer">
+              All day
+            </Label>
+          </div>
+
+          <div className="grid grid-cols-1 gap-4 sm:grid-cols-2">
+            <div className="space-y-2">
+              <Label htmlFor={startId}>Start</Label>
```
BenSeymourODB, line 251: **Bug: start input missing `aria-describedby` for the order-error.**

The end inputs (both `date` and `datetime-local`) correctly wire `aria-describedby={orderError ? errorId : undefined}`, but the two start inputs only have `aria-invalid={!!orderError}` and never point to `errorId`. Screen readers reading the start field will announce it as invalid but give no clue why.

Fix: add `aria-describedby={orderError ? errorId : undefined}` to both start `<Input>` elements (the `type="date"` and the `type="datetime-local"` variants).

---

src/components/calendar/EventCreateDialog.tsx
```diff
@@ -0,0 +1,378 @@
...
+                <Input
+                  id={endId}
+                  type="datetime-local"
+                  value={state.endTimed}
+                  aria-invalid={!!orderError}
+                  aria-describedby={orderError ? errorId : undefined}
+                  onChange={(e) =>
+                    setState((prev) => ({
+                      ...prev,
+                      endTimed: e.target.value,
```
BenSeymourODB, line 305: **A11y: double-wrapping `<fieldset>` + inner `<div role="radiogroup">` creates redundant ARIA roles.**

The `<fieldset>` with its `<legend>Color</legend>` already establishes a group with an accessible name. Adding `role="radiogroup" aria-label="Color"` on the inner `<div>` means AT announces "Color group" twice and the `aria-label` shadows the `<legend>` in some implementations.

Either:
- Keep only `<fieldset>`/`<legend>` and drop the `role`/`aria-label` from the `<div>`, or
- Remove the `<fieldset>`/`<legend>` and rely solely on `role="radiogroup" aria-label="Color"`.

The first option is preferred because `<fieldset>`/`<legend>` is native HTML and works without JS.

---

src/components/calendar/EventCreateDialog.tsx
```diff
@@ -0,0 +1,378 @@
...
+              onClick={() => onOpenChange(false)}
+            >
+              Cancel
+            </Button>
+            <Button type="submit" disabled={!canSubmit}>
+              Create event
+            </Button>
+          </DialogFooter>
+        </form>
+      </DialogContent>
```
BenSeymourODB, line 385: **Bug: all-day end-date validation is too strict when start and end fall on the same calendar day.**

`resolveDates` normalises an all-day end date to `23:59:59.999` of that day, but the order check is `end.getTime() <= start.getTime()`. If a user picks the same date for both start and end, `start` is `00:00:00.000` and `end` is `23:59:59.999`, so the check passes — good.

However if `endAllDay < startAllDay` (e.g. end is the day *before* start) the error fires correctly. The problem is the **converse edge**: an all-day single-day event is valid, but there is no test covering it in all-day mode, and the timed-mode validation test uses `userEvent` to clear/retype the input, which is brittle for `datetime-local` inputs in jsdom. Add an explicit test:

```ts
it("does not show an error for an all-day single-day event", async () => {
  renderOpen({ defaultDate: new Date(2026, 3, 20) });
  await user.type(screen.getByLabelText(/title/i), "Day off");
  await user.click(screen.getByRole("checkbox", { name: /all day/i }));
  // start and end default to the same date — should be valid
  expect(screen.queryByText(/end must be after start/i)).not.toBeInTheDocument();
  expect(screen.getByRole("button", { name: /create event/i })).toBeEnabled();
});
```

Also missing: a test that triggers the order error *in all-day mode* (end date < start date).

---

src/components/calendar/EventCreateDialog.tsx
```diff
@@ -0,0 +1,378 @@
...
+    </Dialog>
+  );
+}
+
+function resolveDates(state: ReturnType<typeof buildInitialState>): {
+  start: Date | null;
+  end: Date | null;
```
BenSeymourODB: **Minor: `resolveDates` parameter type leaks internal state shape.**

```ts
function resolveDates(state: ReturnType<typeof buildInitialState>)
```

`buildInitialState` is a private implementation detail. Coupling the helper's signature to it means renaming or restructuring the state object requires updating `resolveDates` too. Consider extracting a named interface `DialogState` (or inlining the fields you actually need) so the two functions can evolve independently.

---

src/components/calendar/EventCreateDialog.tsx
```diff
@@ -0,0 +1,378 @@
...
+      : null;
+
+  const canSubmit = hasTitle && !!start && !!end && !orderError;
+
+  const handleSubmit = (e: React.FormEvent) => {
+    e.preventDefault();
+    if (!canSubmit || !start || !end) return;
+
+    onCreate({
+      title: trimmedTitle,
```
BenSeymourODB, line 194: **Correctness: form reset on the "previous renders" pattern does not capture a changed `defaultDate`.**

```ts
if (open !== prevOpen) {
  setPrevOpen(open);
  if (open) {
    setState(buildInitialState(defaultDate));
  }
}
```

`defaultDate` is read from the closure at the time the condition fires, so it works correctly today. However, if a future caller keeps the dialog mounted and changes `defaultDate` while `open` remains `true`, the new seed is silently ignored until the dialog is closed and reopened. This is an undocumented contract that will bite the "empty day-cell +" entry point mentioned in the PR description (that caller will want to pass a different `defaultDate` for each cell without unmounting). A comment explaining this constraint would prevent the footgun.

---

src/components/calendar/AddEventButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+// Placeholder "self" user for locally created events. Profile / attendee
+// assignment is deferred to the multi-profile work and to the API wiring
+// in #116 (Google Calendar write-through).
+const LOCAL_USER: IUser = {
+  id: "local",
+  name: "You",
+  picturePath: null,
+};
+
+function generateId(): string {
```
BenSeymourODB, line 19: **Dead code: the `generateId` fallback branch will never execute.**

Next.js 16 targets Node 22 and browsers that all have `crypto.randomUUID`. The guard `typeof crypto !== "undefined" && "randomUUID" in crypto` will always be true in this environment, making the `Date.now` / `Math.random` fallback unreachable and untested. Either remove the fallback and call `crypto.randomUUID()` directly, or document the environments where the fallback is intentionally needed.

---

src/components/calendar/AddEventButton.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { addEvent, selectedDate } = useCalendar();
+  const [open, setOpen] = useState(false);
+
+  const handleCreate = (input: EventCreateInput) => {
+    const event: IEvent = {
+      id: generateId(),
+      title: input.title,
+      description: input.description,
+      color: input.color,
+      isAllDay: input.isAllDay,
```
BenSeymourODB, line 33: **Missing error handling in `handleCreate`.**

`handleCreate` calls `addEvent(event)` and then relies on `EventCreateDialog` to call `onOpenChange(false)`. But `addEvent` is a synchronous in-memory setter today; when #116 lands and it becomes an `async` API call that can throw, an unhandled rejection here will leave the dialog open with no feedback to the user. Consider wrapping in try/catch now and passing an `onError` prop (or surfacing an error state) so the plumbing is ready before the async work lands.

---

src/components/calendar/__tests__/AddEventButton.test.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+    expect(screen.getByRole("dialog")).toBeInTheDocument();
+    expect(
+      screen.getByRole("heading", { name: /create event/i })
+    ).toBeInTheDocument();
+  });
+
+  it("calls addEvent with the submitted payload and closes the dialog", async () => {
+    const user = userEvent.setup();
+    const { ctx } = renderWithContext();
```
BenSeymourODB, line 98: **Type-unsafe mock access — use `vi.mocked()` instead of a double cast.**

```ts
const event = (ctx.addEvent as ReturnType<typeof vi.fn>).mock.calls[0][0] as IEvent;
```

This pattern appears in multiple tests. The `as ReturnType<typeof vi.fn>` cast discards the typed signature, and the subsequent `as IEvent` is an unchecked assertion. Prefer:

```ts
expect(vi.mocked(ctx.addEvent)).toHaveBeenCalledTimes(1);
const event = vi.mocked(ctx.addEvent).mock.calls[0][0];
```

`vi.mocked()` preserves the original function type so the argument type is inferred without casting. Same pattern applies at lines 127 and 148.

---

e2e/event-create.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+    await dialog.getByLabel(/title/i).fill("Team offsite");
+    await dialog.getByLabel(/description/i).fill("All-hands in the mountains");
+    await dialog.getByRole("radio", { name: /purple/i }).check();
+    await dialog.getByRole("button", { name: /create event/i }).click();
+
+    // Dialog should close
+    await expect(dialog).not.toBeVisible();
+
+    // Event should be visible in the calendar grid. The dialog seeds from the
```
BenSeymourODB: **Fragile assertion: `.first()` silently masks duplicate event titles in the grid.**

```ts
await expect(page.getByText("Team offsite").first()).toBeVisible();
```

If the event is accidentally rendered twice (e.g. a double-submit bug), this test still passes. Use an exact locator or assert the count:

```ts
await expect(page.getByText("Team offsite")).toHaveCount(1);
await expect(page.getByText("Team offsite")).toBeVisible();
```

Or scope the locator to the calendar grid element to avoid matching text in the dialog remnant or page chrome.

---

e2e/event-create.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+ * Uses the MockCalendarProvider test fixture at /test/calendar so the flow
+ * can be exercised without a real Google Calendar session. Event submission
+ * goes through CalendarProvider.addEvent(IEvent), which immediately renders
+ * the event in the month grid.
+ *
+ * Video capture is enabled for this suite so the PR can visually document
+ * the feature. Screenshots from the test run should be attached to the PR.
+ */
+
+test.use({ video: "on" });
```
BenSeymourODB: **`video: "on"` records video for every run, including CI — consider `"retain-on-failure"` instead.**

`test.use({ video: "on" })` stores a video for every passing test run. On a 5-project matrix that's 35 videos per run. `"retain-on-failure"` keeps videos only for failing tests, which is the useful signal and avoids bloating CI artifact storage. The PR body says screenshots should be attached, but Playwright video is separate from screenshots — if you want screenshots, use `screenshot: "only-on-failure"` or add explicit `page.screenshot()` calls at the interesting moments.

---

BenSeymourODB (commented):
First-pass review of PR #143 (event creation dialog, closes #82). 8 line-anchored comments filed. Summary of findings:

**Bugs / correctness (must fix before merge)**
1. `EventCreateDialog` — start inputs have `aria-invalid` but no `aria-describedby` pointing to the error paragraph; only the end inputs are wired correctly. Screen readers announce the start field as invalid with no explanation.
2. `resolveDates` — no test covering the all-day order-error path (end date < start date in all-day mode), and no test confirming a same-day all-day event is *valid*. The logic is correct but the test surface has a gap.

**A11y**
3. The color picker double-wraps `<fieldset>`/`<legend>` with an inner `<div role="radiogroup" aria-label="Color">`. The `aria-label` shadows the `<legend>` in some AT and the group is announced twice. Drop either the native or the ARIA grouping.

**Design / forward-compatibility**
4. `defaultDate` reset contract is undocumented — if a caller changes `defaultDate` while `open` stays `true` (the planned day-cell "+" entry point), the new seed is silently ignored. Needs a comment at minimum.
5. `handleCreate` has no error handling around `addEvent`. When #116 makes it async, an unhandled rejection leaves the dialog open with no user feedback.

**Dead code**
6. `generateId` fallback (`Date.now` / `Math.random`) is unreachable in Node 22 / modern browsers. Remove or document.

**Test quality**
7. `vi.fn()` double-cast pattern in `AddEventButton.test.tsx` — use `vi.mocked()` for type safety.
8. E2E happy-path uses `.first()` on the event title locator, masking potential double-render bugs. Also `video: "on"` records 35 videos per CI run; `"retain-on-failure"` is more appropriate.

No manual memoization, no `any` types, no custom Tailwind colors, and no React Compiler violations detected. The controlled-component design and the "previous renders" reset pattern are both sound.

---

src/components/calendar/EventCreateDialog.tsx
BenSeymourODB: Fixed a different way: removed `aria-invalid` from the start inputs entirely in a3c9e25. The order error describes a property of the end date (being before the start), so pointing the error at the end input alone matches the semantics better than marking both inputs invalid. Screen readers now announce the end input as invalid and read the error via `aria-describedby`; the start input is no longer misreported as invalid at all.

---

src/components/calendar/EventCreateDialog.tsx
BenSeymourODB: Fixed in a3c9e25, option 1: kept `<fieldset>`/`<legend>` and dropped `role="radiogroup"` / `aria-label="Color"` from the inner `<div>`. The native fieldset + legend combined with `<input type="radio">` elements sharing a name already forms a labelled radio group — no ARIA needed.

---

src/components/calendar/EventCreateDialog.tsx
BenSeymourODB: Added three tests in a3c9e25 covering this behaviour (see the new `all-day behavior` describe block):

1. `does not show an order error for an all-day single-day event` — asserts no error and Create enabled after checking the all-day box with defaults.
2. `shows the order error when the all-day end date is before the start date` — end set to the day before start, error visible, Create disabled.
3. `submits a multi-day all-day event with endDate on the end day, not the start day` — asserts the serialised `endDate` lands on the end day (22nd) at `23:59:59.999`, not on the start day. This also covers the disputed concern that multi-day all-day events were being clamped to the start day — they aren't.

---

src/components/calendar/EventCreateDialog.tsx
BenSeymourODB: Fixed in a3c9e25. Extracted a `DialogState` interface at module scope; both `buildInitialState(defaultDate: Date | undefined): DialogState` and `resolveDates(state: DialogState)` now use it. The two helpers are now decoupled from each other's internals.

---

src/components/calendar/EventCreateDialog.tsx
BenSeymourODB: Documented in a3c9e25 — added a JSDoc on the `defaultDate` prop stating: _"Only re-read when the dialog transitions from closed → open, so changes while the dialog is already open will not overwrite the user's in-progress edits."_ That makes the contract explicit for the day-cell entry point later on.

---

src/components/calendar/AddEventButton.tsx
BenSeymourODB: Fixed in a3c9e25 — `generateId` now just does `return crypto.randomUUID();`. Node 22 and all browsers we target have `crypto.randomUUID` unconditionally, so the guard and fallback were dead code.

---

src/components/calendar/AddEventButton.tsx
BenSeymourODB: Respectfully disputing this one. The project's `CLAUDE.md` is explicit on the two applicable rules:

> Don't add features, refactor, or introduce abstractions beyond what the task requires. […] Don't design for hypothetical future requirements.
> Don't add error handling, fallbacks, or validation for scenarios that can't happen. Trust internal code and framework guarantees.

Today `addEvent` is a synchronous in-memory setter that cannot throw. Adding try/catch + `onError` now would be designing for #116's hypothetical async shape — and we don't yet know whether the async API will throw, return a Result, pass errors through a toast context, or optimistically apply then rollback. The right place to wire error handling is the PR that makes `addEvent` async, where the real shape is known. Leaving as-is.

---

src/components/calendar/__tests__/AddEventButton.test.tsx
BenSeymourODB: Fixed in a3c9e25 at both callsites (the third line had already been consolidated into one of these two). Now using `vi.mocked(ctx.addEvent).mock.calls[0][0]` — the argument type is inferred from `ICalendarContext.addEvent`, so the double cast to `ReturnType<typeof vi.fn>` + `as IEvent` is gone.

---

e2e/event-create.spec.ts
BenSeymourODB: Fixed in a3c9e25 using the `toHaveCount(1) + toBeVisible()` variant:

```ts
await expect(page.getByText("Team offsite")).toHaveCount(1);
await expect(page.getByText("Team offsite")).toBeVisible();
```

A double-submit bug would now fail the count assertion instead of slipping through `.first()`.

---

e2e/event-create.spec.ts
BenSeymourODB: Fixed in a3c9e25 — switched to `test.use({ video: "retain-on-failure" })`. CI now keeps video only for failing tests. The PR body reference to screenshots was about the original reviewer-facing documentation; explicit `page.screenshot()` calls aren't needed for this run.

---

BenSeymourODB:
Round-one review response pushed as a3c9e25. Summary of actions per finding:

**Addressed:**
- ARIA: removed `aria-invalid` from start inputs (only end drives the error now); dropped duplicate `role="radiogroup"` from color picker.
- Types: extracted `DialogState` interface so `buildInitialState` / `resolveDates` no longer leak `ReturnType<typeof ...>`.
- Docs: JSDoc on `defaultDate` explains the reset-on-open contract.
- Cleanup: dropped dead non-crypto fallback in `generateId`; switched tests to `vi.mocked()` instead of double-cast.
- E2E: `video: "retain-on-failure"` and `toHaveCount(1) + toBeVisible()` (no more `.first()` masking duplicates).

**New coverage:** three tests in the `all-day behavior` block — valid single-day all-day, order error on backwards all-day, and multi-day all-day serializes `endDate` on the correct end day.

**Disputed (one, with reasoning in thread):**
- `handleCreate` speculative try/catch / `onError`: `addEvent` is synchronous today. `CLAUDE.md` forbids error handling "for scenarios that can't happen" and designing "for hypothetical future requirements". Error shape is unknown until #116 picks an async strategy. Will wire error handling in that PR.

All 968 unit/integration tests pass, lint + types + format are clean.

---

BDSeymour:
**Merge readiness: Ready — Wave 2**

Touches `SimpleCalendar.tsx` (+2/0) — a single import plus one JSX line adding `<AddEventButton />` to the toolbar. No overlap with the other SimpleCalendar PRs (#142 header math, #138 event-cell refactor, #139 event modal wiring). CI is green.

Safe to merge at any point in Wave 2.

BenSeymourODB:

---

BenSeymourODB (commented):
## QA verdict — pass

Validated locally with Playwright (Chromium) against all acceptance criteria from issue #82.

### Acceptance criteria

| # | Criterion | Result |
|---|-----------|--------|
| 1 | "Add event" button visible in calendar toolbar | ✅ pass |
| 2 | Dialog opens on click with "Create Event" heading | ✅ pass |
| 3 | Title field required — Create button disabled until filled | ✅ pass |
| 4 | Start/end datetime-local pickers pre-filled to current time | ✅ pass |
| 5 | Six-color swatch radio picker (Blue default) | ✅ pass |
| 6 | Optional description textarea | ✅ pass |
| 7 | "Create event" button submits and dismisses dialog | ✅ pass |
| 8 | Created event renders on the calendar grid | ✅ pass |
| 9 | Inline validation error when end ≤ start | ✅ pass |
| 10 | Cancel closes without creating event | ✅ pass |
| 11 | All-day checkbox switches inputs from datetime-local to date | ✅ pass |
| 12 | Form resets (empty title) when dialog reopened after cancel | ✅ pass |

**Unit tests:** 1038/1038 pass (`pnpm test`)\
**E2E (Chromium):** 7/7 pass (`npx playwright test e2e/event-create.spec.ts --project=chromium`)

### Test infrastructure pushed (commit f428ffe)

One selector bug in the original E2E spec was fixed as test infrastructure:

- **Root cause:** `getByRole('radio', { name: /purple/i }).check()` timed out because the visible colored `<span aria-hidden>` swatch intercepts pointer events at the same coordinates as the `sr-only` 1 px radio input.
- **Fix:** replaced with `dialog.locator('label').filter({ hasText: /^Purple$/ }).click()` — matches actual user behaviour (clicking the label), lets the browser route the event to the radio via implicit label association. Added `.toBeChecked()` to confirm selection.
- Strengthened "dialog dismissed" assertion to `.not.toBeAttached()` (more precise than `.not.toBeVisible()` for Radix portals).
- Added four `page.screenshot()` calls; screenshots committed to `docs/screenshots/event-create/`.

**Repro:** `npx playwright test e2e/event-create.spec.ts --project=chromium`

### Screenshots

**1 — Dialog opens from toolbar (Create button disabled, Blue pre-selected):**
01-dialog-open

**2 — Form filled: title, Purple color selected, description, Create enabled:**
02-dialog-filled

**3 — "Team offsite" rendered on calendar grid (Apr 28, purple):**
03-event-on-calendar

**4 — Validation error "End must be after start" in red, Create disabled:**
04-validation-error

🤖 qa-tester

---

PR #147, opened 2026-04-21
feat(calendar): hide mini-calendar sidebar on month view (#146)

BenSeymourODB:
## Summary

Closes #146.

Deferred follow-up from #140. The mini-calendar sidebar duplicates the main grid on month view, so hide it there and let the main calendar reclaim the full width. Keep it rendered on `day`, `week`, `year`, and `agenda` views where it serves as a date-picker / at-a-glance overview.

## Approach

- **`src/app/calendar/page.tsx`** — single layout conditional keyed on `view` from `CalendarProvider`:
  - `view === "month"` → sidebar slot is omitted and grid collapses to a single column
  - otherwise → `1fr_280px` two-column layout with sidebar
- **`src/app/test/calendar/page.tsx`** — the `sidebar=true` layout now mirrors production via a new `SidebarAwareLayout` helper, so E2E exercises the exact rule that ships.
- The `MiniCalendarSidebar` component itself stays view-agnostic — the decision lives in the page layout (per the issue's implementation note).

## Tests

- **New unit tests** (`src/app/calendar/__tests__/page.test.tsx`) — cover all five view values (`day`, `week`, `month`, `year`, `agenda`) and assert sidebar presence / absence.
- **New E2E spec** (`e2e/mini-calendar-sidebar-layout.spec.ts`, `video: "on"`) — covers:
  - hidden on month view
  - visible on agenda view
  - live toggle when switching between month ↔ agenda via the `ViewSwitcher`
  - sidebar not rendered when the `sidebar=true` param is omitted
- **Existing sidebar E2E** (`e2e/mini-calendar-sidebar.spec.ts`) — now run under `view=agenda` so the sidebar is visible; their assertions on component behavior are unchanged.
- Full suite: 1016/1016 passing; `lint`, `format`, `check-types` all clean.

## Test plan

- [x] Unit: sidebar hidden on month view
- [x] Unit: sidebar visible on day / week / agenda / year
- [x] E2E: sidebar hidden on month view
- [x] E2E: sidebar visible on agenda
- [x] E2E: view-switch toggles visibility live (video captured)
- [x] E2E: sidebar opt-in param still gates rendering
- [x] Manual smoke on dev server: `/test/calendar?view=month&sidebar=true` → no sidebar markup; `/test/calendar?view=agenda&sidebar=true` → sidebar renders

## Notes

[...]

PR #148, opened 2026-04-21
feat(calendar): add color and user filter panel (#85)

### LINKED ISSUE #85: Add color and user filter panel for calendar events
## Description

The Jeraidi calendar includes both a color filter panel and a user/assignee filter. Our test page has basic color filtering, but the production calendar has no filter UI.

### Color Filter
- Dropdown/panel with color checkboxes (blue, green, red, yellow, purple, orange)
- Toggle individual colors on/off to show/hide events by color
- Visual indicators showing which filters are active

### User/Profile Filter
- Avatar-based combobox showing family profiles
- "All" option to show all profiles' events
- Individual profile selection to filter events
- Shows profile initials/avatars in compact pill format

## Reference

- **Jeraidi**: Filter icon button → dropdown with color checkboxes. User combobox with avatar initials (M, A, R, +1) and "All" dropdown
- **charlietlamb**: No filter panel

## Relationship

Related to multi-profile family support (\`.claude/plans/multi-profile-family-support.md\`).

## Priority

Medium — important for multi-calendar and multi-profile use cases.

BenSeymourODB:
## Summary

Closes #85.

Adds a `CalendarFilterPanel` that exposes the existing `CalendarProvider` filter state through two popover-driven controls in the production calendar toolbar:

- **Color filter** (Filter icon → popover): a checkbox + swatch row per `TEventColor`. Selected colors drive `filterEventsBySelectedColors(color)` and render an active-count badge on the trigger.
- **User filter** (avatar-group → popover): an "All" option plus one row per user from `CalendarProvider.users`. Selection calls `setSelectedUserId(userId)` and updates the trigger to show the selected user's avatar + name.
- **Clear filters** ghost button — rendered only when at least one filter is active — calls `clearFilter()`.

Mounted in `src/app/calendar/page.tsx` next to the existing `ViewSwitcher`, and in `src/app/test/calendar/page.tsx` behind a new `filters=true` URL param so E2E tests can drive it against a `MockCalendarProvider`.

No provider changes — the existing `selectedColors`, `selectedUserId`, `users`, and `clearFilter` API already filters `events` correctly; this PR is purely the UI that's been missing from the production view.

## Deferred (in scope of later tickets)

- Persisting filters across reloads.
- Per-calendar filter (currently `calendarIds` are fetched but not exposed).
- Rich "badge count of hidden events" indicator on the triggers — makes more sense once the wire-up issues (#113–#118) land richer per-day/per-color data.

## Test plan

- [x] `pnpm test:run src/components/calendar/__tests__/CalendarFilterPanel.test.tsx` — 17 unit tests covering the color popover, the user popover (including empty-state and selected-user styling), and the conditional Clear filters button.
- [x] `pnpm test:run` — full suite (1028 tests) green.
- [x] `pnpm lint:fix` — 0 errors, pre-existing warnings only.
- [x] `pnpm format:fix`.
- [x] Playwright E2E spec (`e2e/calendar-filter-panel.spec.ts`) with `video: "on"` exercising the filter panel against `/test/calendar?events=family&filters=true`. Cannot execute locally (Playwright browser download blocked in this sandbox) but the spec is committed as documentation and will run wherever Playwright is installed.

## Self-review follow-ups (applied in b3e6647)

Six line comments from the ultrareview pass were addressed in-thread:

- E2E test 4 was using a vacuous `red + kid-1` intersection (empty set). Switched to `purple + kid-1` so the spec proves color+user compose — Piano Lesson survives while Soccer Practice is hidden by the color filter.
- Removed conflicting `role="checkbox"` / `aria-checked` from the color-row inner span in favor of `aria-pressed` on the outer toggle button (WCAG 4.1.2).
- Fixed the "All All" duplicate announcement in the empty-users trigger state by hiding decorative avatars from AT.
- Added a `Check` icon to the selected user row mirroring the color popover pattern for unambiguous active-state feedback.
- Strengthened the initials assertion with a two-word fixture (`"Jack Smith" → "JS"`).
- Clarified the user-count comment in the E2E spec to call out the shared `Family` calendar pseudo-user.

Plan: [.claude/plans/calendar-filter-panel.md](../blob/claude/loving-faraday-1jD7Q/.claude/plans/calendar-filter-panel.md)

https://claude.ai/code/session_01GD1CG1nNE78DDj7yKm8o22


BenSeymourODB:
Heads-up on CI: the `Validate Prisma migrations are up to date` step has failed on both CI runs for this PR, but the failure is **not caused by this PR**.

- This PR touches **zero** Prisma files (diff: `.claude/plans/calendar-filter-panel.md`, `e2e/calendar-filter-panel.spec.ts`, `src/app/calendar/page.tsx`, `src/app/test/calendar/page.tsx`, `src/components/calendar/CalendarFilterPanel.tsx`, `src/components/calendar/__tests__/CalendarFilterPanel.test.tsx`).
- Running the exact CI command locally against this branch:
  ```
  npx prisma migrate diff \
    --from-migrations prisma/migrations \
    --to-schema prisma/schema.prisma \
    --exit-code
  ```
  exits **0** (no drift).
- PR #147, based on the same `prisma/` tree, passed CI 1h before this PR opened.

Most likely a transient GitHub Actions runner hiccup (possibly a Prisma query-engine download failure). A plain **Re-run failed jobs** should clear it — please re-run when convenient. Happy to investigate further if the re-run keeps failing.

Locally, `pnpm lint && pnpm check-types && pnpm build && pnpm test:run` all pass (1028 tests).

---

e2e/calendar-filter-panel.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+    await page
+      .getByTestId("filter-panel-color-popover")
+      .getByTestId("filter-panel-color-option-red")
+      .click();
+    await page.keyboard.press("Escape");
+
+    await page.getByTestId("filter-panel-user-trigger").click();
+    await page
+      .getByTestId("filter-panel-user-popover")
+      .getByTestId("filter-panel-user-option-kid-1")
```
BenSeymourODB, line 107: The combined `red` + `kid-1` filter in this test produces an **empty** event set, not a partial one, because Emma (`kid-1`) has only `green` (Soccer Practice) and `purple` (Piano Lesson) events — she has zero red events. As a result `Work Meeting` being "not visible" is vacuously true: _everything_ is hidden, so the assertion doesn't actually prove the intersection of two filters works.

A more discriminating setup would be one where the user filter passes some events but the color filter removes a specific subset. For example, use `color=green` + `user=kid-2` (Jack): Jack has no green events, but Mom does — so you can prove that Jack's `Art Class` (yellow) is hidden, Mom's `Grocery Shopping` (green) is hidden because it's not Jack's, and — critically — after clearing, both return. Alternatively keep `kid-1` but switch to `color=purple` so that Soccer Practice (green/kid-1) is hidden by the color filter even though it passes the user filter, giving you a more targeted proof.

---

src/components/calendar/CalendarFilterPanel.tsx
```diff
@@ -0,0 +1,253 @@
...
+            {ALL_COLORS.map((color) => {
+              const checked = selectedColors.includes(color);
+              return (
+                <li key={color}>
+                  <button
+                    type="button"
+                    onClick={() => filterEventsBySelectedColors(color)}
+                    className="hover:bg-accent flex w-full items-center gap-3 rounded-md px-2 py-1.5 text-left"
+                    data-testid={`filter-panel-color-option-${color}`}
+                    aria-pressed={checked}
```
BenSeymourODB, line 123: The outer `<button>` carries `aria-pressed={checked}` while the inner `<span>` carries `role="checkbox" aria-checked={checked}`. Mixing these two patterns on the same logical control is problematic: a screen reader will first announce the button as a toggle button (pressed/not pressed), then separately encounter an unlabeled checkbox child. The inner `<span role="checkbox">` also has no accessible name (no `aria-label`, no `aria-labelledby`, no text content), which violates WCAG 4.1.2.

Pick one pattern and drop the other. Because the button already expresses toggle semantics via `aria-pressed`, the simplest fix is to remove `role="checkbox"` and `aria-checked` from the `<span>` entirely and let the button carry the state. If you want the visual checkbox affordance, keep the `<span>` as a pure `aria-hidden="true"` decorative element.

---

src/components/calendar/CalendarFilterPanel.tsx
```diff
@@ -0,0 +1,253 @@
...
+            ) : (
+              <div className="flex -space-x-2">
+                {users.slice(0, 3).map((u) => (
+                  <UserAvatar
+                    key={u.id}
+                    user={u}
+                    className="border-background size-6 border-2"
+                  />
+                ))}
+                {users.length === 0 ? (
```
BenSeymourODB: When `users` is empty **and** `selectedUser` is `null`, the trigger renders both the `<AvatarFallback>All</AvatarFallback>` (the circle avatar) and the separate `<span>All</span>`, so a sighted user sees "All All". The avatar-text fallback makes sense as a placeholder when there's nothing to stack, but it should be `aria-hidden="true"` and the span's text alone should carry the accessible label. Consider also whether you want the avatar at all in the empty state — most designs just show the span text.

---

src/components/calendar/__tests__/CalendarFilterPanel.test.tsx
```diff
@@ -0,0 +1,276 @@
...
+      // All is always present
+      expect(
+        within(popover).getByTestId("filter-panel-user-option-all")
+      ).toBeInTheDocument();
+      expect(
+        within(popover).getByTestId("filter-panel-user-empty")
+      ).toBeInTheDocument();
+    });
+
+    it("displays initials for the selected user on the trigger", () => {
```
BenSeymourODB, line 249: `expect(trigger).toHaveTextContent("E")` passes for the wrong reason: the trigger renders `<span>Emma</span>` alongside the avatar fallback, and "Emma" already contains the letter "E". The test comment says "Emma → 'E'" implying it's checking the initials logic, but it is not — any word starting with E would pass.

Use a user whose initials differ enough from their full name to make the assertion meaningful, or scope the assertion to the `AvatarFallback` element directly. Alternatively use a two-word name like `"Jack Smith"` and assert the trigger contains `"JS"` (not just `"J"`), which would prove the multi-word initials branch works.

---

e2e/calendar-filter-panel.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+  test("selecting a user hides other users' events", async ({ page }) => {
+    // Baseline: events for multiple users are present.
+    await expect(page.getByText("Work Meeting")).toBeVisible(); // Mom
+    await expect(page.getByText("Soccer Practice")).toBeVisible(); // Emma
+
+    await page.getByTestId("filter-panel-user-trigger").click();
+    const userPopover = page.getByTestId("filter-panel-user-popover");
+    await expect(userPopover).toBeVisible();
+
+    // All + the four users from the `family` mock set.
```
BenSeymourODB: The comment `// All + the four users from the \`family\` mock set` is slightly misleading. The family mock has four distinct user IDs — `parent-1` (Mom), `kid-1` (Emma), `kid-2` (Jack), and `family` (Family) — but "Family" is a shared-calendar pseudo-user, not an individual profile. Anyone reading the assertion block expecting four real people will be confused when they look up the mock and find `{ id: "family", name: "Family" }`. A brief clarification keeps the spec self-documenting: `// All + 4 users (Mom, Emma, Jack, and the shared "Family" calendar user)`.

---

src/components/calendar/CalendarFilterPanel.tsx
```diff
@@ -0,0 +1,253 @@
...
+          data-testid="filter-panel-user-popover"
+        >
+          <p className="text-muted-foreground px-2 pt-1 pb-2 text-xs font-medium uppercase">
+            Filter by user
+          </p>
+          <ul className="flex flex-col">
+            <li>
+              <button
+                type="button"
+                onClick={() => setSelectedUserId("all")}
```
BenSeymourODB, line 186: Clicking the "All" option when `selectedUserId` is already `"all"` fires `setSelectedUserId("all")` — a no-op state update — but the popover stays open. This is acceptable UX (Radix Popover doesn't auto-close on item click), but there is no test for this case. More importantly, the same is true for re-clicking the currently selected user: the popover stays open with no visible feedback that the click was registered. Consider either closing the popover on any selection (pass `open`/`onOpenChange` to close after click) or adding a visual checkmark on the selected row so the user knows their selection is active. The `data-selected` attribute is already set but only used for the Tailwind `data-[selected=true]:bg-accent` highlight; a dedicated checkmark icon (mirroring the color popover pattern) would make the active state unambiguous.

---

BenSeymourODB (commented):
Automated review pass surfaced these items. Addressing inline.

---

e2e/calendar-filter-panel.spec.ts
BenSeymourODB: Fixed in b3e6647. Switched the intersection to `color=purple` + `user=kid-1` so the spec actually discriminates:

- Emma (kid-1) has both `green` (Soccer Practice) and `purple` (Piano Lesson) events.
- Under `purple + kid-1`, Piano Lesson survives **both** filters, Soccer Practice is hidden by the color filter (passing the user filter), Work Meeting is hidden by the user filter.

So we now assert both `Work Meeting` (user-filtered) and `Soccer Practice` (color-filtered) are hidden while `Piano Lesson` is visible — proves color+user compose, not that they wipe out everything.

---

src/components/calendar/CalendarFilterPanel.tsx
BenSeymourODB: Fixed in b3e6647. Dropped `role="checkbox"` and `aria-checked` from the inner `<span>` and made it `aria-hidden="true"` — the outer button's `aria-pressed` now owns the toggle semantics unambiguously. The `data-state` attribute stays only as a hook for the Tailwind `data-[state=checked]:*` variants and for the Vitest assertions, which remain accurate as a state signal (not an ARIA role).

---

src/components/calendar/CalendarFilterPanel.tsx
BenSeymourODB: Fixed in b3e6647. Now when `selectedUser` is null and `users` is empty, we render _only_ the `<span>All</span>` — the placeholder `Avatar` with its "All" fallback is gone, eliminating the "All All" duplication sighted users were seeing. Decorative avatars in the trigger (single-user state + stacked-avatars state) are also wrapped in `aria-hidden="true"` now so AT only announces the button's `aria-label` and the name/label text.

---

src/components/calendar/__tests__/CalendarFilterPanel.test.tsx
BenSeymourODB: Fixed in b3e6647. The test now uses a local `jackSmith = { name: "Jack Smith" }` fixture and asserts `expect(trigger).toHaveTextContent("JS")`. The substring `"JS"` does not appear anywhere in `"Jack Smith"` itself (there's a space between `k` and `S`), so matching it can only come from the `AvatarFallback` initials — which exercises the multi-word branch of `initials(name)`. I considered scoping the query to the fallback element directly via `within(trigger).getByText("JS")` but the `toHaveTextContent` form stays consistent with the rest of the file.

---

e2e/calendar-filter-panel.spec.ts
BenSeymourODB: Fixed in b3e6647. Comment now reads: `// All + 4 users from \`family\` (Mom, Emma, Jack, and the shared "Family" calendar pseudo-user, not an individual profile).` so future readers looking up the fixture see why the count is 4 even though there are only three real people.

---

src/components/calendar/CalendarFilterPanel.tsx
BenSeymourODB: Fixed in b3e6647 — took the "add a dedicated checkmark" route rather than auto-closing the popover. Rationale: auto-close trades off against keyboard users scanning the list; a visible `Check` icon per active row is cheaper feedback and mirrors the color popover pattern, which keeps the two filter surfaces consistent.

Both the `All` option and each user row now render a `<Check>` (with `data-testid="filter-panel-user-option-${id}-check"`) when `isSelected`, next to the existing `data-selected` background styling. The no-op re-click is now visually unambiguous without any state/open-close changes.

---

Issue #146, opened 2026-04-21
Hide mini-calendar sidebar on month view (redundant)


BDSeymour:
## Context

Deferred from PR #140 ([\"Add mini-calendar sidebar for date navigation\"](https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/140)), which intentionally left this behavior for a follow-up:

> **Week view placement:** the sidebar currently renders beside both month and agenda views in the production page; a future pass can decide whether to show it only in day/week views per the reference designs.

## Decision

The mini-calendar sidebar should be **hidden when the main view is the month view** — rendering a small month grid next to the full month grid is redundant. It should continue to render alongside **day**, **week**, and **agenda** views, where it adds value as a date-picker / at-a-glance month overview.

This aligns with the reference designs cited in #80 (Jeraidi and charlietlamb), both of which show the mini-calendar only in day/week views.

## Acceptance Criteria

- [ ] Mini-calendar sidebar is **hidden** when the active calendar view is `month`
- [ ] Mini-calendar sidebar is **visible** when the active calendar view is `day`, `week`, or `agenda`
- [ ] Visibility toggles live as the user switches views (no reload required)
- [ ] Layout reflows cleanly — the main calendar expands to fill the reclaimed width
- [ ] Existing mini-calendar tests still pass; add a test covering the month-view hide behavior

## Implementation Notes

- Likely a single conditional in `src/app/calendar/page.tsx` (or wherever the sidebar is mounted) keyed on the current view from `CalendarProvider`
- Keep the sidebar component itself view-agnostic — the layout decides whether to render it

## References

- PR #140 — original sidebar implementation (deferred this item)
- Issue #80 — original sidebar spec (reference designs show sidebar in day/week only)
- Issue #114 — CalendarProvider wiring follow-up (independent of this)

BenSeymourODB:
Claiming this for an automated session. Will hide the mini-calendar sidebar on month view while preserving it on day/week/agenda, add unit + E2E coverage, and open a draft PR.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #149, opened 2026-04-21
feat(calendar): add week and day views (#70)

### LINKED ISSUE #70: Add week and day calendar views
## Description

Currently only month and agenda views are implemented. Week and day views would help users with busy schedules.

## Current State

- Month view ✅
- Agenda view ✅
- Week view ❌
- Day view ❌

## Priority

Medium — future enhancement.

---
*Migrated from `.claude/technical-debt.md` (Medium Priority #5)*

BenSeymourODB:
## Summary

Adds `WeekCalendar` and `DayCalendar` views, and extends `ViewSwitcher` so the production `/calendar` page can switch across **Month / Week / Day / Agenda**. Closes #70.

Both components read from the existing `CalendarProvider` context — they filter and render the events the provider already exposes. Data wiring (narrow time-range API fetches) stays with #113.

## What's in this PR

- **`WeekCalendar`** — 7-column grid of the current week with event pills per day, prev/next week navigation, today button, event count, overflow `+X more`, and today-column highlight.
- **`DayCalendar`** — single-day column view with prev/next day navigation, today button, and events sorted chronologically (all-day events grouped separately).
- **`ViewSwitcher` update** — surfaces Day, Week, Month, Agenda tabs.
- **`/calendar` + `/test/calendar`** — render the right component per view, keeping the mini-calendar sidebar for day/week/agenda (hidden in month view is tracked separately by #146).
- **Tests** — Vitest component tests per view + Playwright E2E covering the Month → Week → Day → Agenda round-trip and per-view rendering.

## Deferred

- Narrow time-range fetch queries — #113
- Event creation / edit dialogs — #81 / #82 / #115 / #116
- Hourly time-grid with click-to-create — out of scope for this slice
- Midnight-tick refresh of the module-level `today` reference so the today-highlight / "Today" label / Today-button disabled state stay correct on a 24/7 wall display without a manual refresh — to be folded into the data-wiring work in #113.

## Test plan

- [x] `pnpm test` — Vitest suites for WeekCalendar and DayCalendar
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types`
- [x] Playwright E2E (`e2e/week-day-views.spec.ts`) — 11 specs covering both views and the round-trip cycle, video on retain-on-failure

Closes #70

https://claude.ai/code/session_01MiHjgJcwHN1LUuYvhmzUhF

### DIFF
--- a/e2e/week-day-views.spec.ts
+++ b/e2e/week-day-views.spec.ts
@@ -1,9 +1,8 @@
 import { expect, test } from "@playwright/test";
 
-// Record video for this spec so the view-transition flow produces an artifact
-// (useful for PR review of the navigation feel). Video lands under
-// `test-results/` which is gitignored.
-test.use({ video: "on" });
+// Keep video only when a test fails so CI doesn't accumulate artifacts for
+// every green run. Artifacts land under `test-results/` which is gitignored.
+test.use({ video: "retain-on-failure" });
 
 test.describe("Week Calendar", () => {
   test.beforeEach(async ({ page }) => {
--- a/e2e/week-day-views.spec.ts
+++ b/e2e/week-day-views.spec.ts
@@ -1,0 +1,103 @@
+import { expect, test } from "@playwright/test";
+
+// Record video for this spec so the view-transition flow produces an artifact
+// (useful for PR review of the navigation feel). Video lands under
+// `test-results/` which is gitignored.
+test.use({ video: "on" });
+
+test.describe("Week Calendar", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto("/test/calendar?events=default&view=week");
+  });
+
+  test("renders all seven weekday headers", async ({ page }) => {
+    for (const label of ["Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat"]) {
+      await expect(
+        page.getByText(label, { exact: true }).first()
+      ).toBeVisible();
+    }
+  });
+
+  test("shows a week range heading", async ({ page }) => {
+    const range = page.getByTestId("week-calendar-range");
+    await expect(range).toBeVisible();
+    await expect(range).toHaveText(/\w+ \d+, \d{4}\s+[–-]\s+\w+ \d+, \d{4}/);
+  });
+
+  test("advances one week with the next button", async ({ page }) => {
+    const range = page.getByTestId("week-calendar-range");
+    const initial = await range.textContent();
+    await page.getByTestId("week-calendar-next").click();
+    await expect(range).not.toHaveText(initial!);
+  });
+
+  test("renders events on the correct day", async ({ page }) => {
+    await expect(page.getByText("Morning Standup")).toBeVisible();
+    await expect(page.getByText("Client Call")).toBeVisible();
+  });
+
+  test("shows '+X more' overflow indicator", async ({ page }) => {
+    await page.goto("/test/calendar?events=overflow&view=week");
+    await expect(page.getByText(/\+\d+ more/).first()).toBeVisible();
+  });
+});
+
+test.describe("Day Calendar", () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto("/test/calendar?events=default&view=day");
+  });
+
+  test("renders the heading and event count", async ({ page }) => {
+    const heading = page.getByTestId("day-calendar-heading");
+    await expect(heading).toBeVisible();
+    await expect(heading).toHaveText(/\w+, \w+ \d+, \d{4}/);
+
+    const count = page.getByTestId("day-calendar-event-count");
+    await expect(count).toHaveText(/\d+ events?/);
+  });
+
+  test("navigates to previous and next day", async ({ page }) => {
+    const heading = page.getByTestId("day-calendar-heading");
+    const initial = await heading.textContent();
+
+    await page.getByTestId("day-calendar-next").click();
+    await expect(heading).not.toHaveText(initial!);
+
+    await page.getByTestId("day-calendar-prev").click();
+    await expect(heading).toHaveText(initial!);
+  });
+
+  test("shows empty state when the day has no events", async ({ page }) => {
+    await page.goto("/test/calendar?events=empty&view=day");
+    await expect(
+      page.getByText("No events scheduled for this day")
+    ).toBeVisible();
+  });
+
+  test("renders today's events on the day view by default", async ({
+    page,
+  }) => {
+    await expect(page.getByText("Morning Standup")).toBeVisible();
+  });
+});
+
+test.describe("View switcher navigation", () => {
+  test("cycles Month → Week → Day → Agenda and back to Month", async ({
+    page,
+  }) => {
+    await page.goto("/test/calendar?events=default&view=month");
+    await expect(page.getByText("Sun", { exact: true }).first()).toBeVisible();
+
+    await page.getByRole("tab", { name: "Week" }).click();
+    await expect(page.getByTestId("week-calendar-range")).toBeVisible();
+
+    await page.getByRole("tab", { name: "Day" }).click();
+    await expect(page.getByTestId("day-calendar-heading")).toBeVisible();
+
+    await page.getByRole("tab", { name: "Agenda" }).click();
+    await expect(page.getByText("Upcoming Events")).toBeVisible();
+
+    await page.getByRole("tab", { name: "Month" }).click();
+    await expect(page.getByText("Sun", { exact: true }).first()).toBeVisible();
+  });
+});
--- a/src/app/test/calendar/page.tsx
+++ b/src/app/test/calendar/page.tsx
@@ -11,7 +11,7 @@ import {
   useCalendar,
 } from "@/components/providers/MockCalendarProvider";
 import { Button } from "@/components/ui/button";
-import type { IEvent, TEventColor } from "@/types/calendar";
+import type { IEvent, TCalendarView, TEventColor } from "@/types/calendar";
 import { Suspense } from "react";
 import { useSearchParams } from "next/navigation";
 
@@ -362,9 +362,7 @@ function TestCalendarContent() {
 
   // Get test configuration from URL params
   const eventSet = searchParams.get("events") || "default";
-  const view =
-    (searchParams.get("view") as "day" | "week" | "month" | "agenda") ||
-    "month";
+  const view = (searchParams.get("view") as TCalendarView) || "month";
   const loading = searchParams.get("loading") === "true";
   const loadingDelay = parseInt(searchParams.get("loadingDelay") || "0", 10);
   const showControls = searchParams.get("controls") !== "false";
--- a/src/app/test/calendar/page.tsx
+++ b/src/app/test/calendar/page.tsx
@@ -1,9 +1,11 @@
 "use client";
 
 import { AgendaCalendar } from "@/components/calendar/AgendaCalendar";
+import { DayCalendar } from "@/components/calendar/DayCalendar";
 import { MiniCalendarSidebar } from "@/components/calendar/MiniCalendarSidebar";
 import { SimpleCalendar } from "@/components/calendar/SimpleCalendar";
 import { ViewSwitcher } from "@/components/calendar/ViewSwitcher";
+import { WeekCalendar } from "@/components/calendar/WeekCalendar";
 import {
   MockCalendarProvider,
   useCalendar,
@@ -276,6 +278,8 @@ function CalendarDisplay() {
 
   return (
     <div data-testid="calendar-display">
+      {view === "day" && <DayCalendar />}
+      {view === "week" && <WeekCalendar />}
       {view === "month" && <SimpleCalendar />}
       {view === "agenda" && <AgendaCalendar />}
     </div>
@@ -358,7 +362,9 @@ function TestCalendarContent() {
 
   // Get test configuration from URL params
   const eventSet = searchParams.get("events") || "default";
-  const view = (searchParams.get("view") as "month" | "agenda") || "month";
+  const view =
+    (searchParams.get("view") as "day" | "week" | "month" | "agenda") ||
+    "month";
   const loading = searchParams.get("loading") === "true";
   const loadingDelay = parseInt(searchParams.get("loadingDelay") || "0", 10);
   const showControls = searchParams.get("controls") !== "false";
--- a/src/components/calendar/DayCalendar.tsx
+++ b/src/components/calendar/DayCalendar.tsx
@@ -1,0 +1,228 @@
+"use client";
+
+import { useCalendar } from "@/components/providers/CalendarProvider";
+import { Button } from "@/components/ui/button";
+import { formatTime } from "@/lib/calendar-helpers";
+import type { IEvent, TEventColor } from "@/types/calendar";
+import {
+  addDays,
+  format,
+  isSameDay,
+  parseISO,
+  startOfDay,
+  subDays,
+} from "date-fns";
+import { ChevronLeft, ChevronRight } from "lucide-react";
+
+function getColorClasses(color: TEventColor): string {
+  const classes: Record<TEventColor, string> = {
+    blue: "border-blue-500 bg-blue-50 dark:bg-blue-950",
+    green: "border-green-500 bg-green-50 dark:bg-green-950",
+    red: "border-red-500 bg-red-50 dark:bg-red-950",
+    yellow: "border-yellow-500 bg-yellow-50 dark:bg-yellow-950",
+    purple: "border-purple-500 bg-purple-50 dark:bg-purple-950",
+    orange: "border-orange-500 bg-orange-50 dark:bg-orange-950",
+  };
+  return classes[color];
+}
+
+function getBadgeClasses(color: TEventColor): string {
+  const classes: Record<TEventColor, string> = {
+    blue: "bg-blue-500",
+    green: "bg-green-500",
+    red: "bg-red-500",
+    yellow: "bg-yellow-500",
+    purple: "bg-purple-500",
+    orange: "bg-orange-500",
+  };
+  return classes[color];
+}
+
+function eventStartsOnDay(event: IEvent, day: Date): boolean {
+  return isSameDay(parseISO(event.startDate), day);
+}
+
+function EventCard({
+  event,
+  use24HourFormat,
+}: {
+  event: IEvent;
+  use24HourFormat: boolean;
+}) {
+  const start = formatTime(event.startDate, use24HourFormat);
+  const end = formatTime(event.endDate, use24HourFormat);
+
+  return (
+    <div
+      className={`rounded-lg border-l-4 p-4 ${getColorClasses(event.color)}`}
+    >
+      <div className="flex items-start justify-between gap-4">
+        <div className="flex-1">
+          <h4
+            className="text-foreground font-semibold"
+            data-testid="day-calendar-event-title"
+          >
+            {event.title}
+          </h4>
+          <p className="text-muted-foreground mt-1 text-sm">
+            {event.isAllDay ? "All day" : `${start} – ${end}`}
+          </p>
+          {event.description && (
+            <p className="text-muted-foreground mt-2 text-sm">
+              {event.description}
+            </p>
+          )}
+        </div>
+        <div
+          className={`h-3 w-3 shrink-0 rounded-full ${getBadgeClasses(event.color)}`}
+        />
+      </div>
+    </div>
+  );
+}
+
+export function DayCalendar() {
+  const { selectedDate, setSelectedDate, events, isLoading, use24HourFormat } =
+    useCalendar();
+
+  const today = new Date();
+  const isToday = isSameDay(selectedDate, today);
+
+  const dayEvents = events.filter((event) =>
+    eventStartsOnDay(event, selectedDate)
+  );
+  const allDayEvents = dayEvents.filter((event) => event.isAllDay);
+  const timedEvents = dayEvents
+    .filter((event) => !event.isAllDay)
+    .sort(
+      (a, b) =>
+        parseISO(a.startDate).getTime() - parseISO(b.startDate).getTime()
+    );
+  const total = dayEvents.length;
+
+  const previousDay = () => setSelectedDate(subDays(selectedDate, 1));
+  const nextDay = () => setSelectedDate(addDays(selectedDate, 1));
+  const goToToday = () => setSelectedDate(new Date());
+
+  return (
+    <div className="w-full space-y-4">
+      {/* Header */}
+      <div className="flex items-center justify-between">
+        <div>
+          <div className="flex items-center gap-3">
+            <h2
+              className="text-foreground text-2xl font-bold"
+              data-testid="day-calendar-heading"
+            >
+              {format(selectedDate, "EEEE, MMMM d, yyyy")}
+            </h2>
+            <span
+              className="text-muted-foreground text-sm"
+              data-testid="day-calendar-event-count"
+            >
+              {total} {total === 1 ? "event" : "events"}
+            </span>
+          </div>
+          <p
+            className="text-muted-foreground text-sm"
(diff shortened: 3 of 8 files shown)


src/components/calendar/WeekCalendar.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const weekEvents = getEventsForWeek(events, selectedDate);
+  const weekEventCount = weekEvents.length;
+
+  const today = new Date();
+  const isCurrentWeek = isSameWeek(selectedDate, today, {
+    weekStartsOn: WEEK_STARTS_ON,
+  });
+
+  const previousWeek = () => setSelectedDate(subDays(selectedDate, 7));
```
BenSeymourODB: **Bug / non-idiomatic navigation:** `previousWeek` / `nextWeek` navigate by calling `subDays(selectedDate, 7)` / `addDays(selectedDate, 7)`. This produces the right numeric result, but it's semantically wrong and potentially fragile around DST transitions (e.g., a 23-hour day in fall-back week). Use the purpose-built helpers instead:

```ts
import { addWeeks, subWeeks } from "date-fns";

const previousWeek = () => setSelectedDate(subWeeks(selectedDate, 1));
const nextWeek = () => setSelectedDate(addWeeks(selectedDate, 1));
```

`addWeeks` / `subWeeks` are already listed as available in `date-fns` and were imported by `calendar-helpers.ts` for the same purpose.

---

src/components/calendar/WeekCalendar.tsx
```diff
@@ -0,0 +1,207 @@
...
+
+  const weekdayHeaders = getShortWeekdayLabels();
+  const weekDates = getWeekDates(selectedDate);
+  const weekStart = startOfWeek(selectedDate, { weekStartsOn: WEEK_STARTS_ON });
+  const weekEnd = endOfWeek(selectedDate, { weekStartsOn: WEEK_STARTS_ON });
+
+  const weekEvents = getEventsForWeek(events, selectedDate);
+  const weekEventCount = weekEvents.length;
+
+  const today = new Date();
```
BenSeymourODB: **Stale clock reference:** `const today = new Date()` is evaluated on every render. For a calendar that can be kept open for a long time (wall display), this means the "today" highlight will be wrong if the component happens to render just after midnight without a navigation event. A simple fix is to derive today at the top of the module or use a `useDateNow` hook; for now, at minimum this should be a stable reference so it doesn't drift between the `isCurrentWeek` check and the cell-level `isSameDay(day, today)` check:

```ts
// At module level or stabilise with a ref if you want reactivity
const today = startOfDay(new Date());
```

Same pattern applies in `DayCalendar.tsx`.

---

src/components/calendar/WeekCalendar.tsx
```diff
@@ -0,0 +1,207 @@
...
+        <div className="border-border bg-muted grid grid-cols-7 border-b">
+          {weekdayHeaders.map((label) => (
+            <div
+              key={label}
+              className="text-muted-foreground p-3 text-center text-sm font-semibold"
+            >
+              {label}
+            </div>
+          ))}
+        </div>
```
BenSeymourODB, line 157: **Accessibility — duplicate/unlabelled day cell header:** Each day cell re-renders `weekdayHeaders[index]` (e.g., "Mon") as a plain `<span>` inside the cell body, duplicating the column header above. Screen readers will announce "Mon" twice per column with no relationship between them. The canonical fix is to use a `<table>` + `<th scope="col">` layout, but as a lightweight improvement you can at least hide the in-cell abbreviation from assistive technology since the column header already conveys it:

```tsx
<span className="text-muted-foreground text-xs font-medium" aria-hidden="true">
  {weekdayHeaders[index]}
</span>
```

Also worth adding `role="columnheader"` (or `scope="col"` via `<th>`) to the header row cells above.

---

src/components/calendar/WeekCalendar.tsx
```diff
@@ -0,0 +1,207 @@
...
+      <div className="flex items-center justify-between">
+        <div>
+          <div className="flex items-center gap-3">
+            <h2 className="text-foreground text-2xl font-bold">
+              {format(weekStart, "MMM d")} – {format(weekEnd, "MMM d, yyyy")}
+            </h2>
+            <span
+              className="text-muted-foreground text-sm"
+              data-testid="week-calendar-event-count"
+            >
```
BenSeymourODB, line 93: **Accessibility — week range heading has no `data-testid` or ARIA label:** The `<h2>` on this line renders a formatted range string (`"Apr 12 – Apr 18, 2026"`) but has no `data-testid`. The E2E test checks `week-calendar-range` on the `<p>` two lines below it, which means the heading and the paragraph contain *redundant* text. Consider whether you need both, or whether the `<h2>` should carry the `data-testid` and the `<p>` below be removed. Keeping both makes the DOM say the range twice with no semantic differentiation.

---

src/components/calendar/WeekCalendar.tsx
```diff
@@ -0,0 +1,207 @@
...
+          <Button
+            variant="outline"
+            size="icon"
+            onClick={nextWeek}
+            data-testid="week-calendar-next"
+            aria-label="Next week"
+          >
+            <ChevronRight className="h-4 w-4" />
+          </Button>
+        </div>
```
BenSeymourODB, line 128: **Missing `role` / `aria-label` on the week grid container:** The 7-column grid is rendered as plain `<div>` elements. Screen readers have no way to identify this as a calendar grid. At a minimum, adding `role="grid"` on the outer grid and `role="gridcell"` on each day column (with an `aria-label` derived from the date) would be a meaningful accessibility improvement:

```tsx
<div className="grid grid-cols-7" role="grid" aria-label="Week of …">
  {weekDates.map((day, index) => (
    <div
      key={…}
      role="gridcell"
      aria-label={format(day, "EEEE, MMMM d")}
      …
    >
```

---

src/components/calendar/WeekCalendar.tsx
```diff
@@ -0,0 +1,207 @@
...
+            const overflow = dayEvents.length - visible.length;
+
+            return (
+              <div
+                key={day.toISOString()}
+                data-testid={isToday ? "week-calendar-today-cell" : undefined}
+                className={`border-border min-h-[240px] border-r border-b p-2 last:border-r-0 ${
+                  isToday ? "bg-blue-50 dark:bg-blue-950" : "bg-card"
+                }`}
+              >
```
BenSeymourODB, line 175: **Missing accessible name on event pills:** Each event pill `<div>` uses `title={event.title}` for a tooltip but has no `role` or `aria-label`. `title` is only surfaced by pointer devices; keyboard and screen-reader users won't get it. Add `role="button"` (or wrap in a `<button>`) with `aria-label` once click-to-edit (#81) lands, or at minimum expose the title to AT now:

```tsx
<div
  role="listitem"
  aria-label={`${event.title}${!event.isAllDay ? `, ${formatTime(event.startDate, use24HourFormat)}` : ""}`}
  …
>
```

---

src/components/calendar/DayCalendar.tsx
```diff
@@ -0,0 +1,228 @@
...
+        <div
+          className={`h-3 w-3 shrink-0 rounded-full ${getBadgeClasses(event.color)}`}
+        />
+      </div>
+    </div>
+  );
+}
+
+export function DayCalendar() {
+  const { selectedDate, setSelectedDate, events, isLoading, use24HourFormat } =
```
BenSeymourODB, line 91: **Stale clock reference (same issue as `WeekCalendar`):** `const today = new Date()` is re-evaluated on every render. For a long-running wall display, today's date could change mid-session and the "Today" button disabled state / "Today" / "Yesterday" / "Tomorrow" relative labels would be wrong until the next navigation event triggers a re-render. Move it to module scope or stabilise it:

```ts
const today = startOfDay(new Date()); // module-level constant
```

(`startOfDay` strips the time component so midnight-boundary comparisons with `isSameDay` are reliable.)

---

src/components/calendar/DayCalendar.tsx
```diff
@@ -0,0 +1,228 @@
...
+        <div>
+          <div className="flex items-center gap-3">
+            <h2
+              className="text-foreground text-2xl font-bold"
+              data-testid="day-calendar-heading"
+            >
+              {format(selectedDate, "EEEE, MMMM d, yyyy")}
+            </h2>
+            <span
+              className="text-muted-foreground text-sm"
```
BenSeymourODB, line 125: **Redundant `startOfDay` call in relative label:** The fallback label uses `format(startOfDay(selectedDate), "PPPP")` — but `format` ignores the time component entirely when the format token (`PPPP`) only outputs date parts, so `startOfDay` here does nothing useful. Simplify to:

```ts
format(selectedDate, "PPPP")
```

---

src/app/test/calendar/page.tsx
```diff
@@ -358,7 +362,9 @@ function TestCalendarContent() {
 
   // Get test configuration from URL params
   const eventSet = searchParams.get("events") || "default";
-  const view = (searchParams.get("view") as "month" | "agenda") || "month";
+  const view =
+    (searchParams.get("view") as "day" | "week" | "month" | "agenda") ||
```
BenSeymourODB: **Type-safety gap — manual union misses `"year"` and diverges from `TCalendarView`:** The cast `"day" | "week" | "month" | "agenda"` duplicates part of `TCalendarView` and silently drops `"year"`. If a future view is added to `TCalendarView`, this file won't catch it. Use the canonical type directly:

```ts
import type { TCalendarView } from "@/types/calendar";

const view = (searchParams.get("view") as TCalendarView) ?? "month";
```

`TCalendarView` is already imported at the top of this file via the `IEvent` import path, so no extra import is needed beyond adding it to the named imports.

---

src/components/calendar/__tests__/WeekCalendar.test.tsx
```diff
@@ -0,0 +1,330 @@
...
+          id: `m-${i}`,
+          title: `Event ${i + 1}`,
+          startDate: midday(day).toISOString(),
+          endDate: midday(day).toISOString(),
+        })
+      );
+
+      renderWithContext({ selectedDate, events });
+
+      expect(screen.getByText(/\+2 more/)).toBeInTheDocument();
```
BenSeymourODB: **Test coverage gap — overflow label text not asserted precisely:** The test asserts `screen.getByText(/\+2 more/)` which passes as long as the string starts with `+2 more`. Consider also asserting the count is exactly right (5 events, `MAX_EVENTS_PER_DAY = 3`, overflow = 2) by using `{ exact: true }` or a stricter regex, to protect against a future change to `MAX_EVENTS_PER_DAY` silently making this test vacuously pass with the wrong count:

```ts
expect(screen.getByText("+2 more")).toBeInTheDocument();
```

---

e2e/week-day-views.spec.ts
```diff
@@ -0,0 +1,103 @@
+import { expect, test } from "@playwright/test";
+
+// Record video for this spec so the view-transition flow produces an artifact
```
BenSeymourODB: **E2E test — `video: "on"` applies to every test in this file:** Recording video for all tests (including navigation assertions that run in milliseconds) will produce large artifacts for every CI run and slow down the suite noticeably. Per CLAUDE.md the `test-results/` directory is gitignored, so artifacts won't be committed, but CI time and storage will still accumulate. Consider scoping video to a single targeted test with `test.use` inside the describe block, or using `"retain-on-failure"` so video is only kept when a test actually fails:

```ts
test.use({ video: "retain-on-failure" });
```

---

src/components/calendar/__tests__/DayCalendar.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+
+      await userEvent.setup().click(btn);
+
+      const called = (contextValue.setSelectedDate as ReturnType<typeof vi.fn>)
+        .mock.calls[0][0] as Date;
+      expect(isSameDay(called, new Date())).toBe(true);
+    });
+  });
+
+  describe("Navigation", () => {
```
BenSeymourODB, line 176: **Missing test — loading state suppresses the empty state message:** The component only shows "No events scheduled for this day" when `!isLoading && allDayEvents.length === 0 && timedEvents.length === 0`. There is a test for `isLoading: true` showing the spinner, but no test that verifies the empty-state message is *absent* while loading. Add:

```ts
it("does not show empty state while loading", () => {
  renderWithContext({ isLoading: true, events: [] });
  expect(screen.queryByText("No events scheduled for this day")).not.toBeInTheDocument();
});
```

---

src/components/calendar/DayCalendar.tsx
```diff
@@ -0,0 +1,228 @@
...
+function EventCard({
+  event,
+  use24HourFormat,
+}: {
+  event: IEvent;
+  use24HourFormat: boolean;
+}) {
+  const start = formatTime(event.startDate, use24HourFormat);
+  const end = formatTime(event.endDate, use24HourFormat);
+
```
BenSeymourODB, line 59: **Missing accessible label on event card's colour dot:** The colour dot `<div>` at the end of each `EventCard` is purely decorative — its colour is already expressed semantically by the card's left border. Mark it `aria-hidden="true"` so screen readers don't announce an unlabelled `<div>`:

```tsx
<div
  aria-hidden="true"
  className={`h-3 w-3 shrink-0 rounded-full ${getBadgeClasses(event.color)}`}
/>
```

---

BenSeymourODB (commented):
## Review — `feat(calendar): add week and day views (#70)`

Good implementation overall — the component structure is clean, tests are thorough, and the CalendarProvider wiring is correct. A few things need to be fixed before merging:

### Must-fix

| # | File | Issue |
|---|------|-------|
| 1 | `WeekCalendar.tsx:68` | `subDays/addDays(x, 7)` used for week navigation instead of `subWeeks/addWeeks` — semantically wrong and fragile around DST transitions |
| 2 | `WeekCalendar.tsx:63`, `DayCalendar.tsx:85` | `new Date()` called on every render — stale "today" reference breaks the today highlight / relative labels on long-lived wall displays |
| 3 | `test/calendar/page.tsx:366` | Manual `"day" | "week" | "month" | "agenda"` cast instead of `TCalendarView` — misses `"year"` and will silently diverge as the type evolves |

### Accessibility

| # | File | Issue |
|---|------|-------|
| 4 | `WeekCalendar.tsx:147` | In-cell weekday abbreviations duplicate column headers without `aria-hidden="true"` |
| 5 | `WeekCalendar.tsx:126` | Week grid has no `role="grid"` / `role="gridcell"` / `aria-label` — not navigable by keyboard/AT |
| 6 | `WeekCalendar.tsx:163` | Event pills have no accessible name (`title` attr is pointer-only) |
| 7 | `DayCalendar.tsx:54` | Decorative colour dot missing `aria-hidden="true"` |

### Minor / polish

| # | File | Issue |
|---|------|-------|
| 8 | `WeekCalendar.tsx:84` | `<h2>` and `<p>` both render the date range — redundant; pick one and give it the `data-testid` |
| 9 | `DayCalendar.tsx:120` | `startOfDay(selectedDate)` inside `format(…, "PPPP")` is a no-op |
| 10 | `e2e/week-day-views.spec.ts:3` | `video: "on"` for every test — prefer `"retain-on-failure"` to avoid bloating CI artifacts |

### Test gaps

| # | File | Issue |
|---|------|-------|
| 11 | `WeekCalendar.test.tsx:298` | Overflow count assertion uses a loose regex — should be exact string `"+2 more"` |
| 12 | `DayCalendar.test.tsx:176` | No test asserting empty-state message is absent while `isLoading: true` |

Inline comments are attached to specific lines for each of the above.

---

src/components/calendar/WeekCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — swapped `subDays(x, 7)` / `addDays(x, 7)` for `subWeeks(x, 1)` / `addWeeks(x, 1)`.

---

src/components/calendar/WeekCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — moved `today` to module scope as `const today = startOfDay(new Date())`, so the `isSameWeek` and cell-level `isSameDay` checks now share a stable reference. Applied the same treatment to `DayCalendar` in this commit.

---

src/components/calendar/WeekCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — added `aria-hidden="true"` to the in-cell weekday abbreviation. Also upgraded the surrounding structure to `role="grid"` / `role="row"` / `role="columnheader"` / `role="gridcell"` (with per-cell `aria-label` derived from the date) so the canonical relationship is now expressed, which addresses the underlying "two spans with no relationship" concern. Keeping the div-based grid (no `<table>`) since the cells contain block-level event pills.

---

src/components/calendar/WeekCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — dropped the `<p>` and promoted `data-testid="week-calendar-range"` onto the `<h2>`. The heading now carries the full `MMM d, yyyy – MMM d, yyyy` range and is the single source of that text.

---

src/components/calendar/WeekCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — the outer week-grid container now has `role="grid"` with `aria-label={\`Week of ${format(weekStart, "MMMM d, yyyy")}\`}`, the header/body rows are `role="row"`, each day column is `role="gridcell"` with `aria-label={format(day, "EEEE, MMMM d")}`, and the column headers carry `role="columnheader"`.

---

src/components/calendar/WeekCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — each event pill now has `role="listitem"` and an `aria-label` of `"<title>, <time>"` (or `"<title>, all day"`). The existing `title` attribute stays for pointer-device hover. Full `role="button"` / keyboard handling can land with click-to-edit in #81.

---

src/components/calendar/DayCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — `today` is now `const today = startOfDay(new Date())` at module scope in `DayCalendar.tsx`, matching the same pattern applied to `WeekCalendar`. Relative labels and the Today-button disabled state now share one stable reference per render.

---

src/components/calendar/DayCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — simplified to `format(selectedDate, "PPPP")`. Agreed, the `PPPP` token only emits date parts, so `startOfDay` was dead weight.

---

src/app/test/calendar/page.tsx
BenSeymourODB: Fixed in c7fca21 — added `TCalendarView` to the existing `@/types/calendar` named imports and replaced the hand-rolled union with `searchParams.get("view") as TCalendarView`. If a new view is added to `TCalendarView` later, this call site now picks it up automatically.

---

src/components/calendar/__tests__/WeekCalendar.test.tsx
BenSeymourODB: Fixed in c7fca21 — tightened to `screen.getByText("+2 more")`. A future tweak to `MAX_EVENTS_PER_DAY` will now surface as a real failure instead of silently passing.

---

e2e/week-day-views.spec.ts
BenSeymourODB: Fixed in c7fca21 — switched to `test.use({ video: "retain-on-failure" })`. Green runs no longer capture any video, and failures still produce a recording for investigation.

---

src/components/calendar/__tests__/DayCalendar.test.tsx
BenSeymourODB: Fixed in c7fca21 — added `it("does not show empty state while loading", ...)` which renders with `{ isLoading: true, events: [] }` and asserts `queryByText("No events scheduled for this day")` is absent. Now guards the `!isLoading && ...` branch against regression.

---

src/components/calendar/DayCalendar.tsx
BenSeymourODB: Fixed in c7fca21 — added `aria-hidden="true"` to the decorative colour dot in `EventCard`. Screen readers now skip over it since the left border already conveys the colour visually.

---

BDSeymour:
## Independent validation — PR is in good shape

Picked this PR via the merge-precedence list in `docs/issue-dependency-analysis.md` (Phase 3 UI; #70 has no remaining blockers; MERGEABLE/CLEAN; CI green) and validated it end-to-end locally on `claude/loving-faraday-SKYJx`.

### Static / unit checks

| Check | Result |
|---|---|
| `pnpm exec vitest run` (WeekCalendar + DayCalendar) | **27 / 27 pass** |
| `pnpm exec eslint` on PR-touched files | clean (no errors, no new warnings) |
| `pnpm exec playwright test e2e/week-day-views.spec.ts --project=chromium` | **10 / 10 pass (9.4s)** |

Note: the PR description has the E2E box unchecked, but `e2e/week-day-views.spec.ts` is present in the diff and passes when run locally — worth ticking that box.

### Browser walkthrough (`/test/calendar`)

Drove the test page through the full view-switching flow in headless Chromium via Playwright. Captured 10 screenshots covering:

- Month → Week → Day → Agenda → Month round-trip via `ViewSwitcher` tabs
- Week `prev` / `next` / `today` buttons (today disables on the current week)
- Day `prev` / `next` flips "Today → Tomorrow → Yesterday" relative labels correctly
- Today highlighting in week view (blue circle on today's cell, `data-testid="week-calendar-today-cell"`)
- Overflow scenario (`?events=overflow`): week view shows top 3 events + `+7 more`; day view lists all 10
- 12-hour format (`?24hour=false`) renders `9:00 AM – 9:30 AM` correctly in day view
- ARIA: week grid has `role="grid"` / `role="row"` / `role="gridcell"` and per-day `aria-label`s

Production `/calendar` is wired identically — verified by reading `src/app/calendar/page.tsx`. Couldn't drive the production page in the browser without an authenticated session (it 307s to `/auth/signin`), but that's an env config matter outside this PR's scope.

### One non-blocking observation

`WeekCalendar.tsx:31` and `DayCalendar.tsx:20` capture `today` as a **module-level** constant:

```ts
const today = startOfDay(new Date());
```

The comment explains this gives stable cross-cell comparisons within a render — fair. But because this runs at module load (not per render), the today-highlight, "Today" relative label, and the disabled state of the Today button will be wrong once the page crosses midnight without a refresh. For a wall calendar that's expected to run 24/7, this is potentially user-visible.

Not a merge blocker — the trade-off is documented and the same pattern can be revisited alongside the data-wiring work in #113. Possible follow-up: hoist `today` to a per-render computation (or `useState` + a midnight-tick effect) so the highlight and labels stay correct across day boundaries.

### Net

The component contracts, ARIA semantics, navigation, overflow handling, and time-format honoring all behave as the issue spec calls for. Nothing in scope is broken.

_Validated locally by Claude Code._

BenSeymourODB:
Thanks for the independent end-to-end pass and the screenshots — much appreciated.

On the non-blocking observation: agreed, the module-level `today` is a deliberate trade-off (the first reviewer asked for a stable per-render reference to fix mid-render drift, and module scope was the simplest fix that satisfied that). For the wall-display midnight-rollover case you flagged, I've added it to the **Deferred** section of the PR description so it's tracked alongside the data-wiring work in #113. Likely shape is a `useDateNow` hook with a midnight-tick effect so the today-highlight, "Today/Tomorrow/Yesterday" relative label, and the Today-button disabled state all refresh on day boundaries without a manual reload.

Also ticked the Playwright E2E checkbox in the PR description — `e2e/week-day-views.spec.ts` is in the diff and green locally / in CI.

---

PR #155, opened 2026-04-22
fix(auth): prevent duplicate Google accounts linking to one user (#61)

### LINKED ISSUE #61: Bug: Multiple Google accounts can link to a single user
## Summary

During QA testing of PR #48, we discovered that a single user (`zenoc2@gmail.com`) had **two Google OAuth accounts** linked to their record in the `Account` table — each with a different `providerAccountId`. This caused the token refresh logic to consistently pick the older, stale account, resulting in perpetual `TokenRefreshFailed` errors and 401s on all API calls, even immediately after a fresh Google sign-in.

## Root Cause

The Prisma schema has a `@@unique([provider, providerAccountId])` constraint on `Account`, but **no** `@@unique([userId, provider])` constraint. This means NextAuth's Prisma adapter will happily create a second `Account` record when a different Google identity signs in and gets matched to the same app user (e.g., via an existing session cookie or email matching).

## Impact

- The `session` callback in `auth.ts` uses `prisma.account.findMany({ where: { userId, provider: "google" } })` and destructures `[0]` — with no ordering, it consistently returned the older, stale account
- That account's refresh token was expired/revoked, so every session check triggered `TokenRefreshFailed`
- All API routes returned 401, and the UI showed "session expired" even after completing a fresh OAuth flow
- The user was stuck in an infinite sign-in loop

## Immediate Fixes Applied (PR #48)

1. **Database**: Deleted the stale duplicate `Account` record
2. **Code**: Added `orderBy: { expires_at: "desc" }` to the `findMany` query in `auth.ts` so it always picks the most recently refreshed account, even if duplicates exist

## Reproduction

Unclear exactly how the second Google account got linked. Likely scenario: signed in with one Google account, then later signed in with a different Google account while the session cookie was still present. NextAuth may have linked the new Google identity to the existing user.

## Future Considerations

- The app does not yet explicitly support multiple Google accounts per user
- A `signIn` callback guard could reject sign-in when a different `providerAccountId` tries to link to an existing user who already has a Google account
- However, future auth + multi-profile work may change the requirements here, so this may be better addressed as part of that effort rather than standalone
- If multi-Google-account support is eventually desired, the session callback would need smarter account selection logic (not just "first result")

## Related

- PR #48 (`claude/calendar-bug-fixes-6t9no`) — contains the `orderBy` fix
- `src/lib/auth/auth.ts` lines 31-33 — the affected query
- `.claude/plans/multi-profile-family-support.md` — future multi-profile work that may overlap

BenSeymourODB:
Closes #61.

## Summary

Roots out the bug from PR #48: a single `User` could have multiple `Account` rows for the same provider with different `providerAccountId`s, leaving `session()` stuck picking a stale refresh token and the user in a perpetual 401 loop.

The tactical `orderBy: { expires_at: "desc" }` patch in PR #48 stopped picking the stale row but did nothing to stop the duplicate from being created. This PR enforces one-Google-account-per-user at two layers:

1. **Database** — adds `@@unique([userId, provider])` to the `Account` model and ships migration `0003_unique_account_user_provider` that creates the matching unique index.
2. **Application** — new `shouldAllowSignIn` helper wired into NextAuth's `callbacks.signIn`. When a sign-in would link a new Google `providerAccountId` to a user that already has a different Google account, the callback returns `false`, NextAuth redirects to `/auth/error`, and we log a `GoogleAccountLinkRejected` event. Returning `false` gives a cleaner user-facing failure than a raw Postgres unique-constraint error.

Explicitly out of scope per the issue's "Future Considerations": multi-Google-account-per-user support.

## Test plan

- [x] `pnpm test` — 1020 tests pass (was 1011 before this branch). New coverage:
  - `prisma/__tests__/migration-setup.test.ts` asserts `@@unique([userId, provider])` on the `Account` model and a `CREATE UNIQUE INDEX` on `Account(userId, provider)` across the migrations directory.
  - `src/lib/auth/__tests__/sign-in-guard.test.ts` — 7 tests covering fresh sign-up, re-sign-in on the same identity, conflicting-identity denial, different-provider allowance, case sensitivity, and multi-existing-account diagnostic messaging.

## Notes for reviewers

- If production ever ended up with duplicate rows again, the migration will fail with "could not create unique index". The dedup SQL is inlined as a comment in the migration file — we intentionally don't auto-dedup so evidence isn't silently destroyed.
- The guard is a pure function in `src/lib/auth/sign-in-guard.ts`, deliberately decoupled from NextAuth so it can be exercised without booting the handler.

https://claude.ai/code/session_01ShTzALvgWB8EZod2dioHW3

### DIFF
--- a/.claude/plans/mini-calendar-sidebar.md
+++ b/.claude/plans/mini-calendar-sidebar.md
@@ -48,7 +48,7 @@ changes are required in this ticket — that remains the follow-up scope of #114
 **Out of scope (deferred)**
 
 - Wiring the sidebar's date-navigation into a day/week view (depends on #70 / #114)
-- Per-event-color dot *clusters* (one dot per event-color up to N) — we ship a
+- Per-event-color dot _clusters_ (one dot per event-color up to N) — we ship a
   single dot colored by the first event, which is enough to signal "something
   is happening that day"; cluster visuals can come in #114 if desired
 - Full provider extensions for per-day event-indicator aggregation (tracked in #114)
@@ -68,7 +68,7 @@ non-empty. The dot color uses a lookup table
 Local `viewMonth` state is seeded from `startOfMonth(selectedDate)`. Chevron
 navigation mutates `viewMonth` only. Clicking a day cell calls
 `setSelectedDate(day)`; if that day is outside the currently viewed month
-(a padding cell), we *also* update `viewMonth` so the highlighted selection
+(a padding cell), we _also_ update `viewMonth` so the highlighted selection
 stays visible.
 
 We intentionally do **not** auto-sync `viewMonth` to follow external
--- a/.claude/plans/prevent-duplicate-google-accounts.md
+++ b/.claude/plans/prevent-duplicate-google-accounts.md
@@ -1,0 +1,82 @@
+# Prevent Duplicate Google Accounts per User (Issue #61)
+
+## Problem
+
+A single `User` can end up with multiple `Account` rows for the same `provider = "google"` but different `providerAccountId`s. When that happens:
+
+- `auth.ts` `session()` picks one via `findMany().then([0])`, which was consistently stale
+- Every session triggered `TokenRefreshFailed` against the stale refresh token
+- Users were stuck in a 401 / "session expired" loop
+
+The Prisma schema has `@@unique([provider, providerAccountId])` but **no** `@@unique([userId, provider])`, so the NextAuth Prisma adapter is free to attach a second Google identity to an existing user when email-matching or cookie state produces a collision.
+
+PR #48 applied a tactical `orderBy: { expires_at: "desc" }` fix. The _root cause_ (duplicate rows allowed at all) is still live. This PR addresses that.
+
+## Scope
+
+One-Google-account-per-user, enforced at two layers:
+
+1. **Database** — `@@unique([userId, provider])` constraint (new migration).
+2. **Application** — `signIn` callback guard that refuses to link a new Google `providerAccountId` onto a user that already has a `google` `Account` row with a _different_ `providerAccountId`.
+
+**Explicitly out of scope** (per the issue's "Future Considerations"):
+
+- Multi-Google-account-per-user support (would need smarter account selection UX/logic).
+- Any change to the `session()` token-refresh path — the `orderBy` patch already handles the defense-in-depth case.
+
+## Design
+
+### Phase 1 — DB constraint + migration
+
+`prisma/schema.prisma`:
+
+```prisma
+model Account {
+  ...
+  @@unique([provider, providerAccountId])
+  @@unique([userId, provider])
+}
+```
+
+New migration `prisma/migrations/0003_unique_account_user_provider/migration.sql`:
+
+```sql
+-- Enforce one account per (user, provider) to prevent stale duplicate OAuth links.
+-- If this fails with a uniqueness violation, run the dedup query in
+-- docs/nextauth-troubleshooting.md before retrying.
+CREATE UNIQUE INDEX "Account_userId_provider_key" ON "Account"("userId", "provider");
+```
+
+Update `prisma/__tests__/migration-setup.test.ts` to cover the new migration + constraint.
+
+### Phase 2 — `signIn` callback guard
+
+Extract a pure helper `shouldAllowSignIn({ user, account, existingAccounts })` that takes:
+
+- `user`: the incoming NextAuth user (may or may not have an `id` yet)
+- `account`: incoming OAuth account (`provider`, `providerAccountId`, etc.)
+- `existingAccounts`: accounts already in the DB for the incoming user
+
+Return `{ allow: boolean, reason?: string }`.
+
+Rules:
+
+- If there is no existing user (fresh sign-up): allow.
+- If the incoming account's `(provider, providerAccountId)` matches an existing row on the user: allow (same identity re-signing in).
+- If the incoming account shares the provider with an existing row but differs on `providerAccountId`: **deny** — this is the bug from #61.
+- Otherwise (different provider): allow.
+
+Wire the helper into `callbacks.signIn` in `auth.ts`. On deny, log a `GoogleAccountLinkRejected` event and return `false` so NextAuth redirects to the error page.
+
+Unit-test the helper exhaustively in `src/lib/auth/__tests__/sign-in-guard.test.ts`.
+
+## Phases
+
+1. Phase 1 — schema constraint + migration + migration-setup tests. Commit + push + open draft PR.
+2. Phase 2 — `shouldAllowSignIn` helper + wiring + unit tests. Commit + push.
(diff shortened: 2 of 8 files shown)


BenSeymourODB (commented):
## Review: fix(auth): prevent duplicate Google accounts linking to one user (#61)

Overall the approach is sound — layered defense (DB constraint + application guard) is the right pattern. The pure-function extraction for `shouldAllowSignIn` makes it easy to test and reason about. There are **two correctness issues** that need to be addressed before merging, plus several minor points.

---

### ISSUE 1 (correctness) — `callbacks.signIn` relies on `!user.id` to detect new users without documenting why

**File:** `src/lib/auth/auth.ts`, line 36

```ts
if (!account || !user.id) return true;
```

The early-return on `!user.id` silently depends on the Prisma adapter's internal sequencing — the adapter hasn't written the `User` row yet during a fresh sign-up, so `user.id` is `undefined`. This is correct *today* but is fragile: there is nothing in the code (or tests) that documents or verifies this coupling. If a future maintainer sees what looks like a redundant null-check and removes it, or if the adapter's behaviour changes, the guard will start querying against a non-existent `userId` on every new sign-up.

At minimum, add a comment explaining why `!user.id` means "new user":

```ts
// user.id is only set after the Prisma adapter has written the User row.
// If it is absent this is a fresh sign-up — no existing accounts to check.
if (!account || !user.id) return true;
```

Ideally, also add a unit/integration test that verifies the wired callback allows a fresh sign-up through even when no DB row exists, rather than relying only on the helper's test for `existingAccounts: []`.

---

### ISSUE 2 (correctness) — `shouldAllowSignIn` ignores `user` completely; the "fresh sign-up" test is misleading

**File:** `src/lib/auth/sign-in-guard.ts`, lines 40–57  
**File:** `src/lib/auth/__tests__/sign-in-guard.test.ts`, lines 20–28

`shouldAllowSignIn` receives `user` in its input but never reads it. The test titled `"allows when there is no existing user (fresh sign-up)"` passes `user: { id: undefined }`, but the guard returns `allow: true` because `existingAccounts` is empty — not because of anything about the user. The test description implies behaviour the code does not implement.

Two acceptable fixes:
1. Remove `user` from `SignInGuardInput` entirely (the helper doesn't need it — the caller already short-circuits before calling it for new users).
2. Keep `user` but rename the test to `"allows when existingAccounts is empty"` and add a comment explaining the `user` field is present for future use.

Option 1 is cleaner.

---

### MINOR — `incomingProviderAccountId` and `existingProviderAccountId` in log event may be PII

**File:** `src/lib/auth/auth.ts`, lines 54–59

Both values are Google's opaque numeric user IDs. Combined with `userId` they fully identify a real Google account and may constitute personal data under GDPR/CCPA once Application Insights telemetry is wired up. The `userId` alone is sufficient to trace the incident in the DB. Consider replacing both with a truncated representation (e.g., last 6 chars) or removing them from the structured event payload and retaining them only in a server-side log that has stricter retention controls.

---

### MINOR — Migration dedup comment has a potential logic problem in the `expires_at` predicate

**File:** `prisma/migrations/0003_unique_account_user_provider/migration.sql`, lines 10–17

The dedup `DELETE` in the comment removes rows where `a.expires_at IS NULL`, treating a null-expiry row as the one to discard. A null `expires_at` could represent a legitimately issued non-expiring offline token. A DBA running this in an incident would silently drop a valid token. The comment should note this edge case or point directly to the PR #48 discussion where the original choice was made.

---

### MINOR — Case-sensitivity test description doesn't match the test body

**File:** `src/lib/auth/__tests__/sign-in-guard.test.ts`, lines 88–104

The test is titled `"is case-sensitive on provider to avoid treating 'google' and 'Google' as distinct"` but the body uses `provider: "google"` for both accounts. It never exercises mixed casing at all — it is a normal deny-path test (same provider, different `providerAccountId`). Either rename the test to reflect what it actually asserts, or add an assertion with `"Google"` vs `"google"` to verify the claim in the title.

---

### MINOR — `StoredAccount` is re-declared in the test file

**File:** `src/lib/auth/__tests__/sign-in-guard.test.ts`, lines 8–11

This is a local re-declaration of a type already implied by `SignInGuardInput` in `sign-in-guard.ts`. If `StoredAccount` drifts from the implementation type the tests will silently stop covering real types. Export `StoredAccount` from `sign-in-guard.ts` and import it in the test.

---

### What is done well

- Layered enforcement (DB unique index + application guard) is the correct pattern.
- Pure-function extraction for `shouldAllowSignIn` is clean and the JSDoc is thorough.
- Migration comment with inline dedup SQL guidance is good operational practice.
- `readonly StoredAccount[]` on the input type prevents accidental mutation.
- Prisma query in `callbacks.signIn` filters by provider before calling the helper — efficient.
- Test coverage is broad and representative for all main paths.
- CLAUDE.md compliance: no `useMemo`/`useCallback`, no `any` types, `logger` from `@/lib/logger`, strict TypeScript throughout, no test artefacts committed.


---

BenSeymourODB:
Thanks — addressed in commit 1afbac3 pushed to this branch.

**ISSUE 1 (`!user.id` documentation)** — Fixed. `callbacks.signIn` now has an inline comment calling out that `user.id` is only populated after the Prisma adapter has written the `User` row, so the short-circuit is load-bearing and must not be removed. I considered adding a wired-callback integration test but decided against it: covering it would require bootstrapping NextAuth + the adapter in-process, which is a disproportionate amount of surface area for one comment-guarded branch. The inline "do NOT remove this guard" comment is the cheap, durable safeguard.

**ISSUE 2 (`shouldAllowSignIn` ignores `user`)** — Fixed with option 1. Removed `user` from `SignInGuardInput` entirely. The "fresh sign-up" test is renamed to `"allows when existingAccounts is empty (fresh sign-up or new provider)"` so the description matches what the code actually asserts.

**MINOR (PII in log event)** — Fixed. Added a `lastSix` helper (also unit-tested) and the log event now emits `existingProviderAccountIdSuffix` / `incomingProviderAccountIdSuffix` instead of the full opaque Google IDs. `userId` alone is enough to trace in the DB; the suffix just distinguishes which of two similar rows was involved.

**MINOR (migration dedup NULL `expires_at`)** — Fixed. The comment block now explicitly flags that a null expires_at can represent a legitimate non-expiring offline token and that the DELETE should never be run without a SELECT inspection first.

**MINOR (case-sensitivity test body)** — Fixed. The test is rewritten to actually compare `"Google"` vs `"google"` and assert the guard allows linking (different casing = different provider). Added a separate explicit deny test (`"denies same-provider sign-in with a different providerAccountId"`) so the path the old test meant to cover stays covered.

**MINOR (`StoredAccount` re-declared in tests)** — Fixed. Exported `StoredAccount` from `sign-in-guard.ts` and the test file now imports it instead of re-declaring.

Full suite: 1024/1024 green (4 new tests for `lastSix` + the explicit same-provider deny path).

---

BDSeymour (approved):

Issue #86, opened 2026-04-01
Add calendar settings panel to production UI


BDSeymour:
## Description

The Jeraidi calendar includes a settings gear panel with display preferences. Our project has some of these settings in the test page and settings page, but they are not accessible from the calendar view itself.

## Settings Needed

- **Event badge style toggle**: Full event badges vs dot-only badges
- **24-hour format toggle**: Switch between 12h and 24h time display (partially implemented in CalendarProvider)
- **Agenda view group-by**: Group agenda events by date, color, or category
- **Week start day**: Configure Monday vs Sunday week start (charlietlamb uses Mon-Sun)

## Reference

- **Jeraidi**: Gear icon → dropdown with toggle switches for event badges, dot badges, 24hr format, and agenda group-by radio options

## Implementation Notes

- Some infrastructure exists (24hr format in CalendarProvider, settings page at /settings)
- Need a compact gear-icon panel accessible directly from the calendar toolbar
- Settings should persist via localStorage or user preferences in database

## Priority

Medium — quality-of-life improvement for daily use.

BenSeymourODB:
Working on this in #151.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #156, opened 2026-04-22
feat(calendar): smooth view transition animations (#87)

### LINKED ISSUE #87: Add smooth view transition animations
## Description

The Jeraidi calendar features smooth animations when transitioning between views (month → week → day, etc.) and when navigating between months/weeks. Our calendar has no transition animations.

## Features

- Animated transitions when switching between view modes (month, week, day, year, agenda)
- Slide/fade animations when navigating prev/next month or week
- Smooth layout transitions when expanding/collapsing panels
- CSS-based or Framer Motion animations

## Reference

- **Jeraidi**: Smooth view transitions throughout
- **charlietlamb**: Basic transitions

## Implementation Notes

- Consider using CSS transitions, Framer Motion, or the View Transitions API
- Keep animations performant (60fps target) — important for wall-mounted display
- Respect \`prefers-reduced-motion\` for accessibility
- React Compiler should handle any memoization concerns automatically

## Priority

Medium — enhances perceived quality and polish significantly for a wall-display calendar.

BenSeymourODB:
Closes #87.

## Summary

Adds smooth, GPU-composited animations to the calendar UI:

- **View-mode change** (Month ↔ Agenda) — fades the active calendar surface in/out via `ViewSwitcher`.
- **Period navigation** in Month view — directional slide when clicking prev/next month.
- **`prefers-reduced-motion` respected** — animations skip cleanly for users who opt out.

Built on a small reusable primitive (`AnimatedSwap`) so the same animation engine extends to Day/Week/Year once those views are wired by their respective PRs.

## Phases

1. `useReducedMotion` hook (this commit) — SSR-safe live media-query subscription.
2. `AnimatedSwap` component — generic single-child swap animator (fade or directional slide).
3. Wire into `src/app/calendar/page.tsx` (view-change fade) and `src/components/calendar/SimpleCalendar.tsx` (month-navigation slide).
4. Playwright E2E with `video: "on"` covering the animation paths and the reduced-motion bypass.

## Plan

`.claude/plans/view-transition-animations.md`

## Out of scope (deferred)

- Animations for Day / Week / Year views — those views are not yet exposed via `ViewSwitcher` (tracked in PRs #149, #145).
- Panel expand/collapse animations — no calendar-page panel toggles exist today.
- User-configurable animation duration — currently constants; can later be exposed via the settings panel (#86).

## Test plan

- [ ] Vitest: `useReducedMotion` (subscribes, live-updates, unsubscribes, SSR-safe)
- [ ] Vitest: `AnimatedSwap` lifecycle (idle → exiting → entering → idle, fade vs slide, reduced-motion skip)
- [ ] Vitest: Calendar wiring (view fade, slide direction)
- [ ] Playwright (Chromium, video on): Month ↔ Agenda fade, prev/next month slide direction, reduced-motion bypass

### DIFF
--- a/.claude/plans/view-transition-animations.md
+++ b/.claude/plans/view-transition-animations.md
@@ -1,0 +1,105 @@
+# Smooth view transition animations (#87)
+
+## Goal
+
+Add smooth, performant animations to the calendar UI when:
+
+1. The top-level view changes (Month ↔ Agenda today; extensible to Day/Week/Year later).
+2. The user navigates between periods in Month view (prev/next month → directional slide).
+
+Animations must respect `prefers-reduced-motion` and target 60 fps for the wall display.
+
+## Non-goals
+
+- Panel expand/collapse animation (out of scope this slice).
+- Wiring animations for views that do not yet exist (Day, Week, Year land in separate PRs).
+- Changing the scheduler's `ScreenTransition` component (different concern — cross-page transitions).
+
+## Architecture
+
+Two new pieces, both internal to `src/components/calendar`:
+
+1. **`useReducedMotion()`** (`src/hooks/use-reduced-motion.ts`)
+   - Reads `(prefers-reduced-motion: reduce)` via `window.matchMedia`.
+   - SSR-safe (returns `false` when `window` is undefined).
+   - Live-updating via `change` event subscription.
+   - Returned boolean is read by every animated component.
+
+2. **`AnimatedSwap`** (`src/components/calendar/animated-swap.tsx`)
+   - Generic single-child swap animator keyed by a `swapKey` string.
+   - Supports `type: "fade" | "slide"`, `direction: "forward" | "backward"`, `durationMs`.
+   - Renders outgoing + incoming children during the transition using CSS transforms/opacity (GPU-composited).
+   - When `prefers-reduced-motion` is set, or `durationMs <= 0`, swaps children instantly — no animation frames.
+   - Pattern mirrors `ScreenTransition` but decoupled from `pathname` semantics.
+
+Integration points:
+
+- `src/app/calendar/page.tsx` — wrap the `SimpleCalendar`/`AgendaCalendar` branch in `AnimatedSwap` keyed on `view` with fade.
+- `src/components/calendar/SimpleCalendar.tsx` — wrap the grid in `AnimatedSwap` keyed on `format(selectedDate, "yyyy-MM")` with slide; pass direction based on whether the user clicked prev vs next.
+
+No changes to `ScreenTransition`, `CalendarProvider` public API, or existing tests.
+
+## Phases
+
+### Phase 1 — `useReducedMotion` hook
+
+- Unit tests (`src/hooks/__tests__/use-reduced-motion.test.ts`):
+  - Returns `false` when `matchMedia.matches` is `false`.
+  - Returns `true` when `matchMedia.matches` is `true`.
+  - Updates on `change` event.
+  - Unsubscribes on unmount.
+  - SSR-safe (returns `false` when `window` is undefined, covered via mocked environment guard).
+- Implement hook to make tests pass.
+
+### Phase 2 — `AnimatedSwap` component
+
+- Component tests (`src/components/calendar/__tests__/animated-swap.test.tsx`):
+  - Renders the incoming child in the idle state.
+  - On `swapKey` change, runs through exiting → entering → idle phases with configured `durationMs`.
+  - `type: "fade"` applies `opacity: 0` to outgoing, no translateX.
+  - `type: "slide"` applies `translateX(-100%)` forward / `translateX(100%)` backward.
+  - Skips animation and swaps instantly when `prefers-reduced-motion: reduce` is set.
+  - Skips animation when `durationMs === 0`.
+  - Same-`swapKey` re-render updates children in place without triggering a transition.
+- Implement component.
+
+### Phase 3 — Wire into calendar surfaces
+
+- Component test (`src/components/calendar/__tests__/calendar-view-animation.test.tsx`):
+  - Toggling the view from Month to Agenda renders both views simultaneously mid-transition.
+  - After the duration elapses, only the Agenda view remains.
+  - With reduced motion, the swap is instant (no simultaneous render).
+- Component test (`src/components/calendar/__tests__/SimpleCalendar.test.tsx`): extend with:
+  - Clicking Next month sets slide direction `forward`.
+  - Clicking Previous month sets slide direction `backward`.
+- Edit `src/app/calendar/page.tsx` to wrap the view branch in `AnimatedSwap type="fade"`.
+- Edit `src/components/calendar/SimpleCalendar.tsx` to track `direction` locally and wrap the grid in `AnimatedSwap type="slide"` keyed on month.
+
+### Phase 4 — Playwright E2E (video capture)
+
+- `e2e/calendar-transitions.spec.ts` with `test.use({ video: "on" })`:
+  - Switching Month → Agenda via `ViewSwitcher` renders both view root containers simultaneously during the transition window.
+  - Clicking Next month advances the header and runs the slide animation (the slide wrapper briefly shows two months).
+  - With `Emulate reduced-motion: "reduce"`, the swap is instant — no two-view overlap frame.
+
+## Test strategy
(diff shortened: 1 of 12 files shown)


BenSeymourODB (commented):
All 1031 tests pass, TypeScript is clean, no manual memoization, no `any` types, no committed test artifacts, and the plan file is present. Here are the real issues found.

---

FILE: src/components/calendar/animated-swap.tsx
LINE: 121
SIDE: RIGHT
SEVERITY: blocker
COMMENT:
The animated container uses `className="relative overflow-hidden"` with no height set. During both the `exiting` and `entering` phases, both child nodes are rendered as `absolute inset-0`, which takes them out of normal flow and collapses the parent's height to zero. In the calendar context this means the entire month grid or agenda view area visually disappears for the full animation duration (2 × `durationMs`, i.e. 600 ms at the default 300 ms setting).

`ScreenTransition` works around this with `style={{ minHeight: "100vh" }}`, which is too blunt for a generic reusable component. A better fix here is to capture the natural height before the animation starts (e.g. via a `ref` and an inline `style={{ minHeight: measuredHeight }}`) or, simpler, to not make both nodes `absolute` simultaneously — keep the incoming node in flow and only the outgoing one absolute during `exiting`. Either way, the current code produces a jarring layout jump on every real-browser render.

---

FILE: src/components/calendar/animated-swap.tsx
LINE: 139
SIDE: RIGHT
SEVERITY: blocker
COMMENT:
During the `exiting` phase the incoming child is rendered at `transform: translateX(0)` with `opacity: 1` and a CSS `transition` applied (lines 139-150). For `type="slide"` this means the new content snaps into its final position immediately on the first paint of the exiting phase — the CSS transition has nothing to animate from because the element is brand-new in the DOM. The outgoing content slides away, but the incoming content is already fully visible at position 0, giving a two-content-on-screen-at-once appearance rather than a cross-slide.

The incoming node should start at `enterStartTransform(type, direction)` with the appropriate `enterStartOpacity` and then transition to `translateX(0)` / `opacity: 1` over `durationMs`. This is exactly what the `@keyframes animated-swap-enter` block does during the `entering` phase, which suggests the two-phase design (exiting → entering) was meant to separate these, but the `exiting`-phase incoming node still needs its off-screen starting position. `ScreenTransition` has the same pattern and the same flaw.

---

FILE: src/components/calendar/animated-swap.tsx
LINE: 98
SIDE: RIGHT
SEVERITY: blocker
COMMENT:
There is a timer race when `swapKey` changes a second time while a transition is already in the `exiting` phase. The "setState during render" block (line 82) calls `setPhase("exiting")` again, but because the value is identical to the current phase, React does not re-render — so the `useEffect` cleanup at line 102 (`return () => clearTimeout(timer)`) never fires. The original timer from the first transition continues running and fires `setPhase("entering")` at the originally scheduled time, which may be well into a new transition cycle. The snapshot captured at snapshot time (the content between the first and second key change) is then animated as the outgoing content, and the final `idle` state can contain stale children.

On a wall-calendar display where the user can repeatedly tap "Next month," this window is wide enough to trigger. A straightforward fix is to use a `useRef` to track the timer and reset it unconditionally any time `swapKey` changes, independent of whether `phase` changes. `ScreenTransition` inherits the same bug, but that is a separate PR.

---

FILE: src/components/calendar/animated-swap.tsx
LINE: 113
SIDE: RIGHT
SEVERITY: nit
COMMENT:
When `reducedMotion` flips from `false` to `true` mid-animation (user changes OS preference while a slide is in progress), `skipAnimation` becomes `true` on the next render and the early-return path at line 113 kicks in, rendering `children` directly. However the two `useEffect` timers are still pending and will fire `setPhase("entering")` and eventually `setPhase("idle")` / `setSnapshotChildren(null)` after the fact. These are no-ops in terms of visible output (the early return masks them), but they are unnecessary state updates. Adding `if (phase !== "idle")` guards in the timer effects, or cancelling the timers from within a third effect that watches `skipAnimation`, would clean this up.

---

FILE: src/components/calendar/__tests__
LINE: 1
SIDE: RIGHT
SEVERITY: blocker
COMMENT:
The plan (`.claude/plans/view-transition-animations.md`, Phase 3) explicitly specifies a component test at `src/components/calendar/__tests__/calendar-view-animation.test.tsx` covering three scenarios: toggling Month→Agenda renders both views simultaneously mid-transition; after the duration elapses only the Agenda view remains; and with reduced motion the swap is instant. That file does not exist in this PR. The `SimpleCalendar.test.tsx` additions cover slide direction, which satisfies part of Phase 3, but the view-mode fade integration tests (wiring `AnimatedSwap` into `CalendarContent`) are absent. Per CLAUDE.md the task is not complete until all plan specs are implemented.

---

FILE: src/hooks/use-reduced-motion.ts
LINE: 23
SIDE: RIGHT
SEVERITY: praise
COMMENT:
The `useReducedMotion` hook is well-structured. Using the function-initializer form of `useState` (`useState<boolean>(getInitialPreference)`) avoids calling `matchMedia` on every render while still reading the real initial value on the first client paint. The `addEventListener` / `removeEventListener` cleanup in the effect is correct and the tests verify the full lifecycle including the `false → true` and `true → false` transition paths. This is a clean, reusable hook that other animated components in the codebase can adopt without modification.

---

GENERAL COMMENT:
The three-phase state machine (idle → exiting → entering → idle), the "setState during render" key-change detection, and the `willChange: "transform, opacity"` GPU hints are all solid choices that mirror `ScreenTransition` faithfully. The test coverage for `AnimatedSwap` in isolation is thorough. The blockers above are all mechanical issues with the animation geometry and the timer lifecycle under rapid input — none of them require an architectural change, just targeted fixes to the exiting-phase incoming-child start position, a height anchor on the container, and a ref-based timer reset.

---

BenSeymourODB:
Thanks — addressed all five findings in `44a6926`.

**1. Container height collapse** (`animated-swap.tsx:121`) — fixed.
The incoming child now stays in normal flow (no `absolute`); only the outgoing snapshot is `absolute inset-0`. The container preserves the natural height of the new content for the entire transition. New test `keeps the incoming child in normal flow so the container preserves its height` enforces this (`__tests__/animated-swap.test.tsx`).

**2. Incoming child snapped to final position** (`animated-swap.tsx:139`) — fixed.
Redesigned to a single `animating` phase with cross-animation: both outgoing and incoming use `@keyframes` (per-`animationId` so they restart on consecutive swaps) instead of a `transition` on a freshly-mounted node. The incoming animates from `enterFrom` (e.g. `translateX(100%)` for slide-forward) to `translateX(0)` over the full duration. New test `animates the incoming child via a keyframe (so it doesn't snap to position)` covers it.

**3. Timer race on rapid `swapKey` changes** (`animated-swap.tsx:98`) — fixed.
The timer now lives in a `useRef` and the timer effect's deps include an incrementing `animationId` that bumps on every swap. Even when `phase` is already `"animating"`, the effect re-runs, clears the previous timer, and reschedules. New test `resets the timer on every swap so a second mid-flight swap completes the full new duration` exercises a back-to-back swap at 50% of the first transition.

**4. Stale timers when reduced-motion turns on mid-animation** (`animated-swap.tsx:113`) — fixed.
A render-time conditional collapses to `idle` and drops the snapshot when `skipAnimation` flips to true mid-flight. The timer effect's cleanup chain (triggered when `phase` flips back to `idle`) clears any pending timeout. New test `collapses to idle and clears pending timers when reduced-motion turns on mid-animation` registers a real `addEventListener` handler and fires it.

(I didn't reach for an effect that calls `setState` because `react-hooks/set-state-in-effect` blocks it, and dereferencing the timer ref during render trips `react-hooks/refs`. The render-time-collapse + effect-cleanup pairing was the cleanest path that satisfied both rules.)

**5. Missing plan-specified component test** — fixed.
Added `src/components/calendar/__tests__/calendar-view-animation.test.tsx` with the three scenarios from the plan: simultaneous Month + Agenda render mid-fade, settle to only Agenda after duration, and instant swap under reduced motion.

**Test deltas:** AnimatedSwap unit tests grew from 11 → 14; new integration suite adds 3. Full suite now 1037/1037 passing (was 1031). Lint clean (only the three pre-existing `` and unused-var warnings unrelated to this PR), TypeScript clean.

The `ScreenTransition` component in the scheduler inherits the same animation-geometry and timer-race issues you flagged. Out of scope for this PR but worth a follow-up.

---

Issue #73, opened 2026-04-01
Accessibility gaps: keyboard navigation, screen reader support, contrast


BDSeymour:
## Description

Several accessibility gaps exist in the calendar UI.

## Known Issues

- No keyboard navigation for calendar grid
- Limited screen reader support for calendar events
- No focus management for modals
- Color contrast not verified for all states
- No text size adjustment controls

## Improvements Needed

- Implement arrow key navigation
- Add proper ARIA labels and roles
- Focus trap for modals
- WCAG 2.1 AA compliance audit
- Font scaling controls

## Priority

Medium — accessibility should not be overlooked.

---
*Migrated from `.claude/technical-debt.md` (Medium Priority #8)*

BenSeymourODB:
Picking this up for an hourly agent run on branch `claude/loving-faraday-w0knI`.

This issue is broad (arrow key nav, focus management, ARIA, contrast, font scaling). Scoping this session to a focused, shippable slice:

### In scope (this PR)

Add keyboard navigation and ARIA semantics to the month-view grid (`src/components/calendar/SimpleCalendar.tsx`), which currently renders as plain `<div>`s with no roles or focus support.

- `role="grid"` on the day grid, `role="row"` on each week, `role="columnheader"` on weekday headers, `role="gridcell"` on day cells
- Roving tabindex: exactly one day cell is `tabIndex=0` at a time (the selected/focused date); all others `tabIndex=-1`
- Keyboard nav inside the grid:
  - `ArrowLeft/ArrowRight` → ±1 day
  - `ArrowUp/ArrowDown` → ±7 days
  - `Home`/`End` → start/end of current week (per `WEEK_STARTS_ON`)
  - `PageUp`/`PageDown` → previous/next month
  - `Shift+PageUp`/`Shift+PageDown` → previous/next year
  - `Enter`/`Space` → select focused day
- `aria-current="date"` on today's cell, `aria-selected` on the selected cell, full descriptive `aria-label` per cell (e.g. "Monday, April 21, 2026 — 2 events")
- Respect `prefers-reduced-motion` if any scroll-into-view is needed (no animations added in this slice)
- Padding cells excluded from grid navigation (`aria-hidden="true"`, not focusable)

### Deferred (staying on this issue)

- Focus management for modals (largely handled by Radix/shadcn — spot-check only)
- Screen-reader audit across other components
- WCAG 2.1 AA contrast audit
- Font scaling controls
[...]

PR #157, opened 2026-04-22
feat(settings): make hardcoded calendar values user-configurable (#71)

### LINKED ISSUE #71: Hardcoded configuration values should be user-configurable
## Description

Several settings are hardcoded and not configurable via UI:

- Auto-refresh interval (15 minutes)
- Event fetch window (6 months)
- Max events displayed per day (3 + "more")
- Cache expiration (none currently)

## Improvement

Add these to the settings panel and persist via user settings.

## Priority

Medium — near-term enhancement.

---
*Migrated from `.claude/technical-debt.md` (Medium Priority #6)*

BenSeymourODB:
## Summary

Makes the calendar's hardcoded data-loading knobs user-configurable (issue #71). Four new settings stored per-user in `UserSettings`, exposed through the Settings form, and consumed at runtime by `CalendarProvider` + `SimpleCalendar`.

| Setting | Range | Default | Was hardcoded in |
|---|---|---|---|
| `calendarRefreshIntervalMinutes` | 5–120 | 15 | `loadSettings()` default / CalendarProvider auto-refresh |
| `calendarFetchMonthsAhead` | 1–12 | 6 | `CalendarProvider.refreshEvents` timeMax |
| `calendarFetchMonthsBehind` | 0–6 | 1 | `CalendarProvider.refreshEvents` timeMin |
| `calendarMaxEventsPerDay` | 1–10 | 3 | `SimpleCalendar` `.slice(0, 3)` + `+N more` |

Defaults match the previous shipping values, so existing users see no behavior change until they open Settings.

## Phases landed

- **Phase 1** — `UserSettings` schema + Prisma migration (`0003_add_calendar_settings`) + PUT /api/settings validation. Unit tests cover each field's range, integer-only enforcement, and a happy-path upsert (22 settings-route tests total).
- **Phase 2** — `useUserSettings` hook (fetches /api/settings once on mount when authed, falls back to shipping defaults). `CalendarProvider` consumes it for the auto-refresh interval and the `refreshEvents` time-window; exposes `maxEventsPerDay` through `ICalendarContext` so `SimpleCalendar` renders the correct number of events before the overflow label. Tests added for the hook (6 cases) and `SimpleCalendar` overflow thresholds (3 cases).
- **Phase 3** — New `CalendarSection` settings component with four sliders whose `min`/`max`/`step` mirror the server-side validation ranges. Wired into `SettingsForm` between Scheduler and Transition sections. Tests cover pluralization + onChange wiring per slider (12 cases).

## Deferred (out of scope for this session)

- **Cache expiration** — the issue mentions it, but the current `EventCache.saveEvents` clears and rewrites the IndexedDB store wholesale on every refresh rather than using per-event TTL timestamps. A proper user-configurable expiration needs a rework of the IndexedDB storage layer (record timestamp per event, sweep expired rows on read, background cleanup). Too big for one session; worth a follow-up issue if prioritized.

## Test plan

- [x] `pnpm test` — 1037 tests pass (76 files)
- [x] `pnpm lint` — only pre-existing warnings, no new errors
- [x] `pnpm build` — succeeds
- [x] `pnpm prisma migrate diff` — schema matches migrations
- [ ] CI green on PR

Closes #71

https://claude.ai/code/session_013WjPPcWYbWVnn8k87MvSRa


BenSeymourODB (commented):
Overall the PR is solid — TDD discipline is real (tests land in the same commit as the feature they cover, in the correct phase order), migration SQL matches the schema diff exactly, and the cancel-flag pattern in `useUserSettings` is correctly structured. Four targeted issues below, ranked by severity.

---

**[BUG — medium] `setIsLoading(true)` fires without a cancel guard**

`src/hooks/useUserSettings.ts`

`setIsLoading(true)` is called synchronously at the top of the effect, *before* the `cancelled` flag is even defined. If the component unmounts between the synchronous `setIsLoading(true)` call and the cleanup teardown running, you get a state update on an unmounted component (React warning, and the `isLoading` state is permanently stuck `true` in any cached instance). The fix is to move `setIsLoading(true)` inside the async IIFE, after the `if (cancelled) return` guard, so both the load-start and load-end updates are cancellable:

```ts
(async () => {
  if (cancelled) return;
  setIsLoading(true);
  try { … }
```

---

**[BUG — medium] Stale-closure risk in the auto-refresh `useEffect` — `refreshEvents` suppressed from deps**

`src/components/providers/CalendarProvider.tsx`

```ts
const interval = setInterval(() => {
  refreshEvents();   // captured at effect-creation time
}, intervalMs);
}, [status, userSettings.calendarRefreshIntervalMinutes]);
```

`refreshEvents` is a function defined (or imported) inside the provider, and it closes over internal state (`colorMappings`, `selectedCalendars`, etc.). It is missing from the dep array and the blanket `eslint-disable` silences the warning. Every tick of `setInterval` will call the version of `refreshEvents` that existed when the effect last ran, not the current one. Either add `refreshEvents` to the array (React Compiler will stabilise it if it qualifies), or use a ref: `const refreshRef = useRef(refreshEvents); refreshRef.current = refreshEvents;` and call `refreshRef.current()` inside the interval. The suppress comment should be removed either way so future regressions are caught.

---

**[CORRECTNESS — low] `maxEventsPerDay=0` produces broken UI if it ever reaches `SimpleCalendar`**

The Slider enforces `min={1}` and the API route rejects < 1, so `0` won't come in via the normal path. But if a value of `0` were stored via a direct DB write or a future seed, every non-empty day would render `+N more` with zero visible events. A one-line clamp in `CalendarProvider` when it reads from `userSettings` closes the gap completely:

```ts
maxEventsPerDay: Math.max(1, userSettings.calendarMaxEventsPerDay),
```

One line, no test change needed (the existing `maxEventsPerDay: 2` test still passes).

---

**[NOTICE] Migration naming convention diverges from Prisma's default**

`prisma/migrations/0003_add_calendar_settings/`

Prisma's auto-generated names are `YYYYMMDDHHMMSS_<slug>`. Using `0003_` works as long as the migration history table records it correctly and all predecessors follow the same convention. If `0001`/`0002` also use sequential numbering this is fine by convention — just confirm `prisma migrate deploy` resolves the history correctly before merging to production.

---

BenSeymourODB:
Thanks for the review. Addressed in `7ee7580`:

1. **`setIsLoading(true)` race** — Moved the load-start update inside the async IIFE with a `cancelled` check at every `await` boundary. Added a regression test (`useUserSettings.test.tsx`) that unmounts before the in-flight fetch resolves and spies on `console.error` for the unmounted-component warning.

2. **Stale-closure in auto-refresh** — Switched to a `refreshEventsRef` pattern so the `setInterval` callback always reaches the current `refreshEvents` closure. Dropped the `eslint-disable-next-line react-hooks/exhaustive-deps` suppression — the dep array is now honest (`[status, userSettings.calendarRefreshIntervalMinutes]` is all the effect actually depends on; the ref handles the rest).

3. **`maxEventsPerDay=0` clamp** — Added `Math.max(1, userSettings.calendarMaxEventsPerDay)` at the context boundary. Slider `min={1}` and API validation (range 1–10) still guard the normal path; this is defense-in-depth for a direct DB write.

4. **Migration naming** — Leaving as `0003_add_calendar_settings`. The project deliberately uses sequential numbering (`0001_initial`, `0002_add_scheduler_settings`, now `0003_...`). Prisma's migration history table records each directory name as-is, so `prisma migrate deploy` will apply them in lexicographic order (which matches the intended order here). Changing this one file to a `YYYYMMDDHHMMSS_` slug would make it inconsistent with the rest of the history. Worth a standalone decision issue if the convention should change project-wide; out of scope for this PR.

Full suite: 1038 tests pass, types clean, lint clean (only pre-existing warnings).

---

BDSeymour:

BDSeymour (approved):
## QA verdict — pass (rebase required before merge)

Local browser validation against issue #71's acceptance criteria — all green:

| Acceptance bullet | Result |
| --- | --- |
| Auto-refresh interval user-configurable | ✅ Slider 5–120 min, step 5, default 15 |
| Event fetch window user-configurable | ✅ Months Ahead 1–12, Months Behind 0–6 |
| Max events per day user-configurable | ✅ Slider 1–10, default 3, with `+N more` overflow |
| Cache expiration user-configurable | ⏭️ Explicitly deferred in PR description (acknowledged) |

### What I exercised

Added `e2e/settings-calendar.spec.ts` (committed to this branch as `54da348`) with 8 Playwright tests covering:

1. Calendar section renders with correct title + description
2. All four sliders show their default values (`15 minutes`, `6 months`, `1 month`, `3 events`)
3. Each slider exposes the documented `aria-valuemin`/`aria-valuemax`/`aria-valuenow`
4. Refresh Interval slider updates value + label on `ArrowRight` (15 → 20)
5. Months Behind slider can step down to 0 with correct plural "0 months"
6. Max Events Per Day pluralizes correctly: "1 event" (singular) / "3 events" (plural)
7. Changing a slider PUTs only the changed field to `/api/settings`
8. Calendar section sits between Screen Rotation and Page Transitions in the form

**Result:** 8/8 pass on chromium. Full unit suite still green: 1038/1038. `pnpm check-types` clean. `pnpm lint` shows only pre-existing warnings, no new errors.

### Repro

```bash
pnpm exec playwright test e2e/settings-calendar.spec.ts --project=chromium
```

### Screenshots

- `docs/screenshots/pr-157-calendar-settings.png` — Calendar section at default values
- `docs/screenshots/pr-157-settings-full.png` — full settings form ordering

### ⚠️ Cannot merge yet — branch is `CONFLICTING`

Since this PR was opened, `main` advanced (notably via #175 which also touched `SimpleCalendar.tsx`). `gh pr view 157` reports `mergeStateStatus=DIRTY` / `mergeable=CONFLICTING`. Conflicts:

- `src/components/calendar/SimpleCalendar.tsx`
- `src/components/calendar/__tests__/SimpleCalendar.test.tsx`

Per my QA guardrails I won't merge a conflicting PR. **Please rebase onto `main` and resolve those two files**; QA approval stands and the PR can be merged once the rebase lands and CI is green.

🤖 qa-tester

BenSeymourODB:
Merged `main` into the branch and pulled in your QA commit. Conflict resolution preserved both:

- Main's new ARIA grid + keyboard navigation in `SimpleCalendar.tsx` (PR #73)
- This PR's `maxEventsPerDay`-driven slice & overflow label

Also added the new `maxEventsPerDay: 3` field to the mock contexts in the new `DayCalendar.test.tsx` and `WeekCalendar.test.tsx` from PR #149 so `ICalendarContext` stays strict.

Bumped my migration from `0003_add_calendar_settings` to `0004_add_calendar_settings` so it doesn't collide with `0003_unique_account_user_provider` from #155. `prisma migrate diff --from-migrations --to-schema --exit-code` exits 0.

Full suite after merge: 1217 + your 8 e2e = green. `pnpm check-types` clean. `pnpm lint` clean (only pre-existing warnings).

CI should be running on `8130d7c` now — the QA approval should still apply unless the merge surfaces a new regression.

---

PR #158, opened 2026-04-22
feat(auth): encrypt OAuth tokens at rest (#69)

### LINKED ISSUE #69: Token storage security: consider encryption for stored tokens
## Description

OAuth tokens in localStorage are unencrypted. Currently acceptable for home/family use but could be improved.

## Possible Improvements

- Move to IndexedDB with Web Crypto API encryption
- Implement session timeout for inactive users

## Priority

Low — acceptable for current single-device home use case.

---
*Migrated from `.claude/technical-debt.md` (Medium Priority #4)*

BenSeymourODB:
## Summary

Closes #69.

OAuth `access_token` / `refresh_token` / `id_token` values were being written to the Prisma `Account` table in plaintext by `@auth/prisma-adapter` on link and by the session-callback refresh block in `src/lib/auth/auth.ts`. Anyone with read access to the database (backup, rogue migration, compromised app user) could exfiltrate live Google Calendar / Tasks credentials. This PR encrypts those values at rest with AES-256-GCM.

The issue was originally filed when tokens lived in `localStorage`; the server-side auth migration moved the attack surface into PostgreSQL but the encryption work was never done.

### 1. Cipher module — `src/lib/crypto/token-cipher.ts`

- AES-256-GCM with a versioned envelope: `v1:<iv>:<authTag>:<ciphertext>`, base64url-encoded. 12-byte random IV per encryption, 16-byte auth tag. Tamper detection comes for free from GCM.
- **Key resolution**:
  1. `TOKEN_ENCRYPTION_KEY` env var (base64, must decode to exactly 32 bytes)
  2. Non-production fallback: HKDF-SHA256 derivation from `NEXTAUTH_SECRET` with info `token-cipher/v1` so dev setups don't need a second secret
  3. Production without `TOKEN_ENCRYPTION_KEY` → fatal on first use
- **Backwards compatibility**: strings without a `v1:` prefix are legacy plaintext and decrypt to themselves. Callers use `isEncrypted()` if they need to know. No stop-the-world data migration needed — the first refresh per user re-encrypts in place.

### 2. Wiring — `src/lib/auth/`

- `helpers.ts` — `getAccessToken` decrypts before returning; `getGoogleAccount` returns a clone with `access_token` / `refresh_token` / `id_token` decrypted. API-route callers see plaintext and don't need to know.
- `auth.ts` session-callback refresh — decrypts the stored refresh token before hitting Google's token endpoint; encrypts the new `access_token` / `refresh_token` before `prisma.account.update`. Google only returns a new refresh token on first consent / revocation, so the existing plaintext is re-encrypted when a replacement isn't provided.
- `auth.ts` new `events.linkAccount` hook — `@auth/prisma-adapter` writes the initial OAuth account in plaintext, so the hook does an immediate `UPDATE` with `v1:` envelopes. Non-fatal on failure: the next refresh re-encrypts.

### 3. Operational docs

- `TOKEN_ENCRYPTION_KEY` added to `.env.example` with generation instructions and a note that the dev fallback derives from `NEXTAUTH_SECRET`.
- `docs/auth-token-encryption.md` covers envelope format, write/read flows, backwards-compat behavior, key generation, rotation (forced re-login), and suspected-compromise response.

## Test plan

- [x] `src/lib/crypto/__tests__/token-cipher.test.ts` — **22** tests: roundtrip, random-IV uniqueness, empty/unicode/large plaintext, null passthrough, legacy-plaintext passthrough, tamper detection on ciphertext + authTag, malformed envelopes, unknown version, wrong-key rejection, key-derivation fallback, production-key-required, key-length validation, non-base64 validation.
- [x] `src/lib/auth/__tests__/helpers.test.ts` — **+5** new tests: v1 decryption on read, legacy plaintext passthrough, multi-field decryption on `getGoogleAccount`, null preservation.
- [x] Full suite: `pnpm test` → **1038 tests passing** (all prior tests untouched).

## Migration notes

- **No schema change.** Token columns stay `String?`; encryption is at the value layer.
- **No data migration required.** Existing databases keep working — plaintext values pass through on read, and each user's first session refresh rewrites their row with `v1:` envelopes. After every active user has refreshed at least once (typically within Google's 1h `expires_at`), all rows are encrypted.
- **Production must set `TOKEN_ENCRYPTION_KEY`.** Rotation forces all users to re-login; in-place re-encryption tooling is deferred.

## Explicitly deferred (follow-ups)

- Automated key-rotation tooling. The envelope reserves `v2`/`v3` for future algorithms or schemes; actual rotation with re-encryption is a separate workstream.
- Session timeout for inactive users (second bullet of the original issue) — separate concern.

## Commits

1. `feat(crypto): add AES-256-GCM token cipher module` — cipher + 22 unit tests
2. `feat(auth): encrypt OAuth tokens at rest` — wire encryption into `auth.ts` + `helpers.ts` (+5 tests)
3. `docs(auth): document TOKEN_ENCRYPTION_KEY and at-rest encryption` — env + operational docs

https://claude.ai/code/session_01VyNegPDrVUbPsycGvBXZNc

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,6 +26,27 @@ NEXTAUTH_URL=http://localhost:3000
 NEXTAUTH_SECRET=your-secret-key-here
 
 # =============================================================================
+# TOKEN ENCRYPTION
+# =============================================================================
+
+# Key used to encrypt Google OAuth access_token / refresh_token / id_token at
+# rest in the PostgreSQL Account table (AES-256-GCM). Must decode to exactly
+# 32 bytes.
+#
+# Generate with: openssl rand -base64 32
+#
+# REQUIRED in production. In development/test, the cipher falls back to an
+# HKDF-derived key based on NEXTAUTH_SECRET so local setup stays simple, but
+# rotating the fallback invalidates previously stored tokens — set this
+# explicitly for any shared or long-lived environment.
+#
+# Rotating this key invalidates every encrypted token currently stored. Users
+# will be prompted to sign in again on their next session refresh.
+#
+# See docs/auth-token-encryption.md for details.
+TOKEN_ENCRYPTION_KEY=your-32-byte-base64-encryption-key
+
+# =============================================================================
 # GOOGLE OAUTH
 # =============================================================================
 
--- a/docs/auth-token-encryption.md
+++ b/docs/auth-token-encryption.md
@@ -36,13 +36,28 @@ algorithm or key format without breaking reads.
 ## Writing flow
 
 1. **Initial OAuth link** — `@auth/prisma-adapter` creates the row with
-   plaintext tokens, then `events.linkAccount` in
-   `src/lib/auth/auth.ts` runs and `UPDATE`s the row with `v1:...`
-   envelopes. The plaintext window is one Prisma round trip.
+   plaintext tokens, then the `events.linkAccount` hook in
+   `src/lib/auth/auth.ts` (delegating to
+   `src/lib/auth/link-account.ts`) runs and `UPDATE`s the row with `v1:...`
+   envelopes.
+   - **Happy-path plaintext window**: one Prisma round trip (tens of ms).
+   - **Worst-case plaintext window**: if `encryptLinkedAccount` itself
+     fails (DB hiccup, transient connection error), the failure is logged
+     with context `LinkAccountEncryptionFailed` and sign-in is **not**
+     blocked. Plaintext remains in the row until the next successful
+     session-callback refresh re-encrypts it — up to the access token's
+     lifetime (~1h for Google's default `expires_in`). Callers reading via
+     `getAccessToken` / `getGoogleAccount` continue to work throughout
+     because they accept legacy plaintext transparently.
 2. **Token refresh** — when the session callback detects an expired access
    token, the stored refresh token is decrypted, Google's token endpoint is
    called, and the new access + refresh tokens are encrypted before
    `prisma.account.update`.
+   - Decryption failures here (tampered envelope, GCM auth-tag mismatch,
+     unknown envelope version after a key rotation) log with context
+     `RefreshTokenDecryptFailed` before the outer catch marks the session
+     with `RefreshTokenError`. Operators can tell a cipher-side problem
+     apart from a Google-side refresh failure.
 
 ## Reading flow
 
--- a/docs/auth-token-encryption.md
+++ b/docs/auth-token-encryption.md
@@ -1,0 +1,137 @@
+# OAuth token encryption at rest
+
+Google OAuth `access_token`, `refresh_token`, and `id_token` values are
+encrypted before they are written to the `Account` table in PostgreSQL. This
+document covers how the encryption works, how to operate it, and how to
+respond when things go wrong.
+
+## What is encrypted
+
+- `Account.access_token`
+- `Account.refresh_token`
+- `Account.id_token`
+
+Other `Account` columns (provider, scope, expires_at, etc.) are **not**
+encrypted — they are used for queries and routing, and leaking them without
+the token values does not compromise authentication.
+
+## How it works
+
+Tokens are wrapped in a versioned envelope:
+
+```
+v1:<iv>:<authTag>:<ciphertext>
+```
+
+- **Algorithm**: AES-256-GCM (authenticated encryption — any tampering fails
+  decryption).
+- **IV**: 12 random bytes per encryption (GCM standard).
+- **Auth tag**: 16 bytes.
+- **Encoding**: every segment is base64url.
+- **Key**: 32 bytes sourced from `TOKEN_ENCRYPTION_KEY` (see below).
+
+The envelope format is versioned so a future `v2:` can introduce a new
+algorithm or key format without breaking reads.
+
+## Writing flow
+
+1. **Initial OAuth link** — `@auth/prisma-adapter` creates the row with
+   plaintext tokens, then `events.linkAccount` in
+   `src/lib/auth/auth.ts` runs and `UPDATE`s the row with `v1:...`
+   envelopes. The plaintext window is one Prisma round trip.
+2. **Token refresh** — when the session callback detects an expired access
+   token, the stored refresh token is decrypted, Google's token endpoint is
+   called, and the new access + refresh tokens are encrypted before
+   `prisma.account.update`.
+
+## Reading flow
+
+`src/lib/auth/helpers.ts` owns the read path:
+
+- `getAccessToken()` decrypts `access_token` before returning it.
+- `getGoogleAccount()` returns a clone of the row with every token field
+  decrypted.
+
+Callers (API routes, etc.) always see plaintext and never need to know the
+tokens are encrypted on disk.
+
+## Backwards compatibility
+
+`decryptToken` returns legacy plaintext (no `v1:` prefix) unchanged. That
+means the first deploy of this feature does **not** need a data migration:
+
+- Existing plaintext tokens keep working for reads.
+- The first successful token refresh for each user re-writes the row with
+  `v1:` envelopes.
+- After every account has refreshed at least once (typically within the
+  `expires_at` lifetime — 1 hour for Google — of any active user), all rows
+  are encrypted.
+
+Use `isEncrypted(value)` from `src/lib/crypto/token-cipher.ts` if you need to
+check whether a value is already enveloped (e.g. in a background migration).
+
+## `TOKEN_ENCRYPTION_KEY`
+
+### Generating
+
+```bash
+openssl rand -base64 32
+```
+
+Must decode to **exactly 32 bytes**. Invalid base64 or the wrong byte length
+is a fatal error at first use.
+
+### Required in production
+
+If `NODE_ENV=production` and `TOKEN_ENCRYPTION_KEY` is missing or empty,
+`encryptToken` / `decryptToken` throw. This is intentional — starting
+production with no encryption key is never correct.
+
+### Development fallback
+
+In non-production environments, if `TOKEN_ENCRYPTION_KEY` is absent the
+cipher derives a 32-byte key from `NEXTAUTH_SECRET` via HKDF-SHA256 with info
+`token-cipher/v1`. This keeps local setup to a single secret.
+
+The fallback is deterministic: the same `NEXTAUTH_SECRET` always produces the
+same derived key. If you need stable encrypted values across a dev team (or
+between dev and staging) you should set `TOKEN_ENCRYPTION_KEY` explicitly.
+
+## Operations
+
+### Rotating the key
+
+Rotating `TOKEN_ENCRYPTION_KEY` invalidates every currently stored token
+envelope. The next time each user's session refreshes, decryption of their
+refresh token fails, the session callback hits the `RefreshTokenError`
+branch, and the user is prompted to sign in again.
+
+In-place rotation with re-encryption would need a migration script that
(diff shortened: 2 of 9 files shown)


BenSeymourODB (commented):
## Review: feat(auth): encrypt OAuth tokens at rest (#69)

Overall the AES-256-GCM implementation is sound — correct IV length (12 bytes), correct auth-tag length (16 bytes), authenticated encryption, versioned envelope format, and the legacy-plaintext passthrough is handled consistently. The read-path is well-covered. A handful of issues below range from a concrete correctness bug to missing tests and a misleading error message.

---

### 1. BUG — `base64url`-encoded keys are silently rejected as "invalid base64"

**File:** `src/lib/crypto/token-cipher.ts` — `resolveKey()`

The re-encode validation works by decoding the key with `Buffer.from(envKey, 'base64')` then re-encoding with `.toString('base64')` and comparing. If a developer (or CI secret injector) generates the key with `node -e "console.log(require('crypto').randomBytes(32).toString('base64url'))"`, the resulting string contains `-` and `_` (base64url alphabet). Those characters are silently mapped to `+` and `/` during decode, so the re-encoded form won't match, and the function throws "TOKEN_ENCRYPTION_KEY is not valid base64" — even though the decoded key is a perfectly valid 32 bytes.

```
# Reproducer:
node -e "
  const key = require('crypto').randomBytes(32);
  const b64url = key.toString('base64url');
  const decoded = Buffer.from(b64url, 'base64');
  const reencoded = decoded.toString('base64').replace(/=+$/, '');
  console.log('match:', reencoded === b64url.replace(/=+$/, '')); // false
"
```

The docs and `.env.example` both say `openssl rand -base64 32` (which produces standard base64), so this won't bite most users, but it will produce a very confusing failure for anyone who generates the key with a Node.js snippet or a secrets manager that emits base64url. Suggest either:
- Normalising the input before validation: `const normalised = envKey.replace(/-/g, '+').replace(/_/g, '/');`
- Or accepting both alphabets by checking that `decoded.length === KEY_LENGTH` only (since `Buffer.from` already decodes both).

---

### 2. MISSING TESTS — `events.linkAccount` has zero test coverage

**File:** `src/lib/auth/auth.ts` lines 147–184

The `linkAccount` handler is the most security-sensitive new code path (it's the first moment a token lands encrypted on disk), but there are no tests for it. The helpers test suite covers decrypt-on-read but not encrypt-on-link. Missing cases:

- Token fields are updated with `v1:` envelopes after account is linked.
- Non-google providers are skipped (returns early).
- All three fields are absent → skips the DB call entirely.
- `prisma.account.update` throws → sign-in is NOT blocked; error is logged.

Given the CLAUDE.md TDD requirement ("No feature is complete until all tests pass"), this is a gap.

---

### 3. MISLEADING ERROR — wrong message when neither key is set in dev/test

**File:** `src/lib/crypto/token-cipher.ts` lines 72–76

When `TOKEN_ENCRYPTION_KEY` is absent **and** `NEXTAUTH_SECRET` is also absent (or empty), the `else` branch throws:

```
"TOKEN_ENCRYPTION_KEY is required in production."
```

…even when `NODE_ENV=test` or `NODE_ENV=development`. A developer who forgets to set either variable in CI will see a production-flavoured error that implies they're running the wrong environment, when the real fix is just to set one of the two secrets. The error should branch on `nodeEnv`:

```ts
throw new Error(
  nodeEnv === "production"
    ? "TOKEN_ENCRYPTION_KEY is required in production. ..."
    : "TOKEN_ENCRYPTION_KEY or NEXTAUTH_SECRET must be set."
);
```

---

### 4. NO WARNING for silent dev/staging key derivation

**File:** `src/lib/crypto/token-cipher.ts` lines 61–72

When `TOKEN_ENCRYPTION_KEY` is absent in non-production, the cipher silently derives a key from `NEXTAUTH_SECRET` with no log output. This is fine for a local laptop, but a shared staging environment or a CI environment with `NODE_ENV=staging` (or any non-`production` string) that forgets to set `TOKEN_ENCRYPTION_KEY` will silently use a derived key — and everyone will be confused when rotating `NEXTAUTH_SECRET` suddenly invalidates all tokens.

Suggest adding a single `console.warn` (or `logger.log`) at the point of derivation:

```ts
console.warn(
  "[token-cipher] TOKEN_ENCRYPTION_KEY not set; deriving key from NEXTAUTH_SECRET. " +
  "Set TOKEN_ENCRYPTION_KEY explicitly for any shared or persistent environment."
);
```

---

### 5. MINOR — `decryptToken` swallows tamper errors in the session callback without setting `RefreshTokenError`

**File:** `src/lib/auth/auth.ts` lines 55–59

```ts
const refreshTokenPlaintext = decryptToken(googleAccount.refresh_token);
if (!refreshTokenPlaintext) {
  throw new Error("No refresh token available");
}
```

If `decryptToken` **throws** (tampered envelope, unexpected envelope version, GCM auth-tag mismatch), the exception propagates out of the `if (googleAccount.expires_at ...)` block into the outer `try/catch` (line 111), which logs "TokenRefreshFailed" and sets `session.error = "RefreshTokenError"`. That's actually the correct observable result, but the logged error will say something like "Unsupported token envelope version: v2" rather than "decryption failed", making key-rotation incidents hard to diagnose. A quick pre-check with `isEncrypted()` or a dedicated catch block around `decryptToken` would make the failure mode much clearer in logs.

---

### 6. DOCS — rotation story understates the plaintext window on initial link

**File:** `docs/auth-token-encryption.md` — "Writing flow"

The doc says _"The plaintext window is one Prisma round trip"_ but doesn't mention the failure mode: if the `linkAccount` event's `prisma.account.update` fails (network blip, Prisma error), the catch block intentionally logs and continues, leaving plaintext in the DB until the next token refresh (up to 1 hour for Google's default `expires_in`). Callers reading via `getAccessToken`/`getGoogleAccount` handle this correctly (legacy passthrough), but it's worth noting explicitly so operators understand the actual worst-case window.

---

### Summary of verdict

Cipher is cryptographically correct. Read path is complete (all API routes go through `getAccessToken`/`getGoogleAccount`; the `/api/auth/refresh-token` route is a client-side GIS proxy and doesn't touch the DB). The base64url validation bug (#1) is the only concrete correctness issue — it won't affect the happy path but will produce a confusing failure for a plausible key-generation command. The missing `linkAccount` tests (#2) are a CLAUDE.md TDD compliance gap. Items #3–#6 are quality/observability improvements.

---

BenSeymourODB:
Thanks for the careful read. Second-round follow-up, point-by-point — all six addressed in 4e04329.

**1. Base64url key bug** — fixed. `resolveKey` now uses a new `decodeBase64Key()` that normalizes `-`/`_` → `+`/`/` before validating, so `randomBytes(32).toString("base64url")` and secret-manager-emitted base64url keys are accepted. New test generates a key whose standard-base64 form contains `+` or `/` and asserts the base64url form still round-trips.

**2. `linkAccount` TDD gap** — fixed. Extracted the handler into `src/lib/auth/link-account.ts` as a standalone `encryptLinkedAccount(account)`, and added 5 tests in `src/lib/auth/__tests__/link-account.test.ts`: encrypts all three token fields on google provider, skips non-google providers, skips when no tokens present, only writes fields that are set, and catches prisma errors without rethrowing (sign-in is not blocked). The inline handler in `auth.ts` now just awaits the extracted function.

**3. Misleading dev error** — fixed. `resolveKey` now branches on `NODE_ENV`: production gets the existing "required in production" message; non-production with neither key nor `NEXTAUTH_SECRET` set gets _"Cannot resolve token encryption key: set TOKEN_ENCRYPTION_KEY (base64, 32 bytes) or NEXTAUTH_SECRET for the development fallback."_ Test asserts the dev-flavoured message does **not** contain "required in production".

**4. Silent derivation warning** — added. One-shot `console.warn` at the moment the HKDF fallback fires; gated by a module-level `loggedDerivationWarning` flag so it fires exactly once per process. Test uses `vi.spyOn(console, 'warn')` + multiple `encryptToken` calls to assert the message appears exactly once.

**5. Decrypt observability** — fixed. The session-callback refresh now wraps `decryptToken(googleAccount.refresh_token)` in its own try/catch that logs context `RefreshTokenDecryptFailed` with the user id, then re-throws into the outer catch (which still sets `session.error = "RefreshTokenError"` as before, preserving the client-facing behavior). Key-rotation incidents are now clearly distinguishable from Google-side refresh failures in the log stream.

**6. Docs worst-case window** — fixed. `docs/auth-token-encryption.md` writing-flow section now spells out both the happy-path window (one Prisma round-trip) and the worst case (if `encryptLinkedAccount` itself fails, plaintext remains at rest up to the Google access-token lifetime — ~1h — until the next successful refresh re-encrypts; reads continue to work via the legacy passthrough). Also documents the `RefreshTokenDecryptFailed` log context.

Net: +8 tests (1038 → 1046 total), all passing; `lint:fix`, `format:fix`, `check-types` clean. Ready for another pass whenever you want.

---

BDSeymour:

BDSeymour (approved):
## QA verdict — pass

Pure backend change (`src/lib/auth/`, `src/lib/crypto/`, `.env.example`, docs). Per the QA skill, browser validation does not apply — no React component, route, copy, or visual change is touched. Validated against issue #69 and the round-1 review locally.

### Acceptance bullets

| From | Bullet | Status | Evidence |
|---|---|---|---|
| #69 | Move OAuth tokens to encryption (was: `localStorage` + Web Crypto, post-server-migration → at-rest in PostgreSQL) | ✅ | AES-256-GCM with versioned `v1:<iv>:<authTag>:<ciphertext>` envelope; 25 unit tests in `src/lib/crypto/__tests__/token-cipher.test.ts` |
| #69 | Session timeout for inactive users | ⏭️ Deferred | Explicitly out of scope in PR body, separate concern |
| Review #1 | base64url-encoded key bug in `resolveKey()` | ✅ Fixed | `decodeBase64Key()` accepts both alphabets, dedicated test |
| Review #2 | `events.linkAccount` had no test coverage | ✅ Fixed | Handler extracted to `src/lib/auth/link-account.ts`; 5 tests cover google encryption, non-google skip, no-token skip, partial-fields payload, prisma-error swallow |
| Review #3 | Misleading "required in production" error in dev/test | ✅ Fixed | Branches on `nodeEnv`; test asserts dev error doesn't mention "required in production" |
| Review #4 | Silent dev/staging key derivation | ✅ Fixed | One-shot `console.warn` on HKDF fallback; test asserts fires exactly once per process |
| Review #5 | Tamper errors in session-callback refresh logged generically | ✅ Fixed | Dedicated `try/catch` with context `RefreshTokenDecryptFailed` before outer error mapping to `RefreshTokenError` |
| Review #6 | Docs understated worst-case plaintext window on initial link | ✅ Fixed | `docs/auth-token-encryption.md` now documents both happy-path (one Prisma round trip) and worst-case (~1h until next refresh re-encrypts) windows |

### Local verification

- `pnpm test` → **1046 passed (76 files)**
- Targeted: `pnpm test src/lib/crypto src/lib/auth` → **65 passed** (token-cipher 25, helpers 35, link-account 5)
- `pnpm check-types` → clean (zero errors)
- `pnpm lint` → 0 errors, 3 pre-existing warnings (unrelated to this PR — `no-img-element` in profile/account UI, an unused mock var)

### Notes for next time

- The `mockUserId` unused-var warning in `link-account.test.ts:17` is harmless; could be removed in a follow-up.
- TDD discipline on this round 2 was clean — every fix shipped with a corresponding test (base64url alphabet, prod-error branching, single-warn fallback, link-account coverage).

🤖 qa-tester

PR #159, opened 2026-04-22
feat(google-calendar): extend API types and add pure mappers (#76, #77)

### LINKED ISSUE #76: Manually defined Google API type definitions may be incomplete
## Description

Some external types (gapi) are manually defined and may not be complete.

## Improvements

- Verify types against official Google Calendar API docs
- Consider using official type definitions if available
- Document any custom type extensions

## Priority

Low — current types are sufficient.

---
*Migrated from `.claude/technical-debt.md` (Low Priority #3)*

### LINKED ISSUE #77: Clarify colorId field comment in google-calendar.ts
## Description

The `colorId` field comment in `google-calendar.ts` is unclear about why it's set to an empty string.

## Fix

Clarify comment to explain: "Intentionally empty: calendarList items do not include an event-level colorId"

## Priority

Low — fix when next editing the file.

---
*Migrated from `.claude/technical-debt.md` (Low Priority #4)*

BenSeymourODB:
## Summary

Closes #76 — manually defined Google Calendar API types were incomplete.
Closes #77 — the `colorId: ""` comment in `fetchCalendarColorMappings` was ambiguous.

### What changed (two commits)

**Commit 1 — extend types and add pure mappers:**

- **`src/types/gapi.d.ts` rewritten to a documented, comprehensive partial model of the Google Calendar API v3 surface.** Additions (all cross-referenced to Google's published schema):
  - `Event`: `status`, `htmlLink`, `created`/`updated`, `iCalUID`, `etag`, `sequence`, `location`, `hangoutLink`, `conferenceData`, `transparency`/`visibility`/sharing booleans, `attendees`, `reminders`, `recurrence`/`recurringEventId`/`originalStartTime`, `attachments`, `source`, `eventType`, `extendedProperties`.
  - `CalendarListEntry`: `colorId`, `timeZone`, `summaryOverride`, `selected`, `accessRole`, `hidden`, `deleted`, `defaultReminders`, `notificationSettings`, `conferenceProperties`, `etag`.
  - `EventsListResponse` / `CalendarListListResponse`: `nextPageToken`, `nextSyncToken`, `etag`, `kind`, `updated`, `timeZone`, `accessRole`, `defaultReminders`.
  - `events.list` request: `pageToken`, `syncToken`, `q`, `iCalUID`, `updatedMin`, `timeZone`, `fields` (field mask), `eventTypes`, extended-property filters, `showHiddenInvitations`.
  - `google.accounts.oauth2`: fleshed-out `TokenResponse` (incl. `error_uri`/`state`), `TokenClient.requestAccessToken` overrides (`hint`, `scope`, `include_granted_scopes`, etc.), `hasGrantedAnyScope`/`hasGrantedAllScopes`.
  - `gapi.client.init`: accept `clientId`/`scope`; richer `getToken()` shape.
- **Pure mappers added to `src/lib/google-calendar.ts`** — `normalizeFetchedEvent` and `normalizeCalendarListEntry` — so the same normalisation logic can be reused by both the client gapi path and any server-side REST caller. `fetchCalendarEvents` and `fetchUserCalendars` now delegate to them.
- **`GoogleCalendarEvent` extended** with the new Google-v3 optional fields (all optional — no caller is forced to change). New `UserCalendar` type exported for calendar-list consumers.
- **#77 colorId comment clarified**: the `""` value is deliberate because `CalendarColorMapping.colorId` models *event-level* colorIds (`calendar#colors.event`), which calendar-list entries do not carry — we derive the color from `backgroundColor` instead.
- **Unit tests** (`src/lib/__tests__/google-calendar-mappers.test.ts`, 9 cases) exercise both mappers against realistic API payloads and assert that extended optional fields pass through.

**Commit 2 — consume canonical types in server routes:**

Previously, each of the three calendar API routes re-declared its own partial interface for Google Calendar responses:

- `src/app/api/calendar/events/route.ts` — `GoogleCalendarEvent` with a `[key: string]: unknown` index signature (defeating field-name checking) plus a `CalendarEventWithId` subtype.
- `src/app/api/calendar/calendars/route.ts` — `GoogleCalendarListItem`.
- `src/app/api/calendar/colors/route.ts` — another `GoogleCalendarListItem` (different field subset).

All three are now replaced by the shared canonical types:
- the ambient `gapi.client.calendar.Event` / `gapi.client.calendar.CalendarListEntry` from `src/types/gapi.d.ts`;
- the runtime `GoogleCalendarEvent` / `normalizeFetchedEvent` from `@/lib/google-calendar`.

The index signature that was papering over missing fields is gone — typos in event-field accesses are now compile errors. Added a regression test (`route.test.ts`) asserting extended Google Calendar v3 fields (status, location, htmlLink, iCalUID, etag, organizer, attendees, recurrence, eventType) pass through the route's serialisation.

### Why

- #76: our hand-written types are now an auditable reference — every field has a doc comment pointing to the Google schema, consumers get IntelliSense for the full surface, and typos become compile errors. Duplicate partial interfaces across the three server routes are gone.
- #77: the comment now tells the reader *why* `colorId` is empty, not just *that* it is.

### Test plan

- [x] `pnpm test` — 1021 tests pass (75 files), including the 9 new mapper tests and the new route regression test.
- [x] `pnpm lint:fix` / `pnpm format:fix` — clean.

No UI changes — this is a pure type/infrastructure refactor, so no Playwright coverage is added.

https://claude.ai/code/session_01Uc76r4qY2Ea7kaTsy7Ev4T

### DIFF
--- a/src/app/api/calendar/calendars/route.ts
+++ b/src/app/api/calendar/calendars/route.ts
@@ -28,19 +28,6 @@ export interface CalendarsResponse {
   calendars: CalendarInfo[];
 }
 
-/**
- * Raw Google Calendar API calendarList item
- */
-interface GoogleCalendarListItem {
-  id: string;
-  summary: string;
-  description?: string;
-  backgroundColor?: string;
-  foregroundColor?: string;
-  primary?: boolean;
-  selected?: boolean;
-}
-
 export async function GET() {
   try {
     // Check authentication
@@ -100,7 +87,7 @@ export async function GET() {
     }
 
     const data = await response.json();
-    const items: GoogleCalendarListItem[] = data.items || [];
+    const items: gapi.client.calendar.CalendarListEntry[] = data.items || [];
 
     // Transform to our CalendarInfo format
     const calendars: CalendarInfo[] = items.map((item) => ({
--- a/src/app/api/calendar/calendars/route.ts
+++ b/src/app/api/calendar/calendars/route.ts
@@ -89,10 +89,12 @@ export async function GET() {
     const data = await response.json();
     const items: gapi.client.calendar.CalendarListEntry[] = data.items || [];
 
-    // Transform to our CalendarInfo format
+    // Transform to our CalendarInfo format. `summary` is optional per the
+    // Google schema — fall back to empty string so downstream UI code can
+    // treat it as a plain string.
     const calendars: CalendarInfo[] = items.map((item) => ({
       id: item.id,
-      summary: item.summary,
+      summary: item.summary ?? "",
       description: item.description,
       backgroundColor: item.backgroundColor || "#4285f4", // Default Google blue
       foregroundColor: item.foregroundColor || "#ffffff",
--- a/src/app/api/calendar/events/route.ts
+++ b/src/app/api/calendar/events/route.ts
@@ -4,26 +4,14 @@
  * Supports fetching from multiple calendars
  */
 import { AuthError, getAccessToken, getSession } from "@/lib/auth";
+import {
+  type GoogleCalendarEvent,
+  normalizeFetchedEvent,
+} from "@/lib/google-calendar";
 import { logger } from "@/lib/logger";
 import { NextRequest, NextResponse } from "next/server";
 
 const GOOGLE_CALENDAR_API = "https://www.googleapis.com/calendar/v3";
-
-interface GoogleCalendarEvent {
-  id: string;
-  summary?: string;
-  description?: string;
-  start: { dateTime?: string; date?: string };
-  end: { dateTime?: string; date?: string };
-  colorId?: string;
-  creator?: { email?: string; displayName?: string };
-  organizer?: { email?: string; displayName?: string };
-  [key: string]: unknown;
-}
-
-interface CalendarEventWithId extends GoogleCalendarEvent {
-  calendarId: string;
-}
 
 interface CalendarFetchError {
   calendarId: string;
@@ -42,7 +30,7 @@ async function fetchEventsFromCalendar(
   maxResults: string,
   singleEvents: boolean
 ): Promise<{
-  events: CalendarEventWithId[];
+  events: GoogleCalendarEvent[];
   error?: CalendarFetchError;
   summary?: string;
   timeZone?: string;
@@ -76,11 +64,9 @@ async function fetchEventsFromCalendar(
   }
 
   const data = await response.json();
-  const events: CalendarEventWithId[] = (data.items || []).map(
-    (event: GoogleCalendarEvent) => ({
-      ...event,
-      calendarId,
-    })
+  const events: GoogleCalendarEvent[] = (data.items || []).map(
+    (event: gapi.client.calendar.Event) =>
+      normalizeFetchedEvent(event, calendarId)
   );
 
   return {
@@ -141,7 +127,7 @@ export async function GET(request: NextRequest) {
     );
 
     // Collect events and errors
-    const allEvents: CalendarEventWithId[] = [];
+    const allEvents: GoogleCalendarEvent[] = [];
     const errors: CalendarFetchError[] = [];
     let summary: string | undefined;
     let timeZone: string | undefined;
@@ -194,7 +180,7 @@ export async function GET(request: NextRequest) {
     });
 
     const response: {
-      events: CalendarEventWithId[];
+      events: GoogleCalendarEvent[];
       nextPageToken?: string;
       summary?: string;
       timeZone?: string;
--- a/src/app/api/calendar/events/route.ts
+++ b/src/app/api/calendar/events/route.ts
@@ -7,7 +7,7 @@ import { AuthError, getAccessToken, getSession } from "@/lib/auth";
 import {
   type GoogleCalendarEvent,
   normalizeFetchedEvent,
-} from "@/lib/google-calendar";
+} from "@/lib/google-calendar-mappers";
 import { logger } from "@/lib/logger";
 import { NextRequest, NextResponse } from "next/server";
 
--- a/src/lib/__tests__/google-calendar-mappers.test.ts
+++ b/src/lib/__tests__/google-calendar-mappers.test.ts
@@ -1,0 +1,262 @@
+/**
+ * Unit tests for Google Calendar API response mappers.
+ *
+ * These tests exercise the pure mappers that translate raw Google Calendar API
+ * responses into the app's `GoogleCalendarEvent` and `UserCalendar` shapes.
+ *
+ * They serve as regression guards for two concerns:
+ *  1. Extended optional fields surfaced by the richer `gapi.d.ts` types are
+ *     preserved through the mappers (issue #76).
+ *  2. Comments/semantics around `fetchCalendarColorMappings` stay stable (#77).
+ */
+import type { CalendarColorMapping } from "@/lib/calendar-storage";
+import {
+  type GoogleCalendarEvent,
+  type UserCalendar,
+  normalizeCalendarListEntry,
+  normalizeFetchedEvent,
+} from "@/lib/google-calendar";
+import { describe, expect, it } from "vitest";
+
+describe("normalizeFetchedEvent", () => {
+  it("preserves required fields and stamps calendarId", () => {
+    const apiEvent: gapi.client.calendar.Event = {
+      id: "evt-1",
+      summary: "Standup",
+      start: { dateTime: "2026-05-01T09:00:00-04:00" },
+      end: { dateTime: "2026-05-01T09:30:00-04:00" },
+    };
+
+    const result = normalizeFetchedEvent(apiEvent, "primary");
+
+    expect(result.id).toBe("evt-1");
+    expect(result.summary).toBe("Standup");
+    expect(result.calendarId).toBe("primary");
+    expect(result.start.dateTime).toBe("2026-05-01T09:00:00-04:00");
+    expect(result.end.dateTime).toBe("2026-05-01T09:30:00-04:00");
+  });
+
+  it("passes through extended optional fields from Google Calendar v3", () => {
+    const apiEvent: gapi.client.calendar.Event = {
+      id: "evt-rich",
+      summary: "Board meeting",
+      description: "Quarterly review",
+      start: { dateTime: "2026-05-01T14:00:00Z", timeZone: "UTC" },
+      end: { dateTime: "2026-05-01T15:00:00Z", timeZone: "UTC" },
+      status: "confirmed",
+      location: "HQ Conference Room A",
+      htmlLink: "https://www.google.com/calendar/event?eid=abc",
+      iCalUID: "evt-rich@google.com",
+      etag: '"etag-123"',
+      created: "2026-04-20T12:00:00Z",
+      updated: "2026-04-21T08:00:00Z",
+      transparency: "opaque",
+      visibility: "default",
+      eventType: "default",
+      hangoutLink: "https://meet.google.com/abc-defg-hij",
+      colorId: "4",
+      creator: { email: "alice@example.com", displayName: "Alice" },
+      organizer: { email: "alice@example.com", displayName: "Alice" },
+      attendees: [
+        {
+          email: "bob@example.com",
+          displayName: "Bob",
+          responseStatus: "accepted",
+        },
+      ],
+      recurrence: ["RRULE:FREQ=WEEKLY;BYDAY=MO"],
+    };
+
+    const result = normalizeFetchedEvent(apiEvent, "team@example.com");
+
+    expect(result.calendarId).toBe("team@example.com");
+    expect(result.status).toBe("confirmed");
+    expect(result.location).toBe("HQ Conference Room A");
+    expect(result.htmlLink).toBe(
+      "https://www.google.com/calendar/event?eid=abc"
+    );
+    expect(result.iCalUID).toBe("evt-rich@google.com");
+    expect(result.etag).toBe('"etag-123"');
+    expect(result.created).toBe("2026-04-20T12:00:00Z");
+    expect(result.updated).toBe("2026-04-21T08:00:00Z");
+    expect(result.transparency).toBe("opaque");
+    expect(result.visibility).toBe("default");
+    expect(result.eventType).toBe("default");
+    expect(result.hangoutLink).toBe("https://meet.google.com/abc-defg-hij");
+    expect(result.organizer?.email).toBe("alice@example.com");
+    expect(result.attendees).toHaveLength(1);
+    expect(result.attendees?.[0]?.responseStatus).toBe("accepted");
+    expect(result.recurrence).toEqual(["RRULE:FREQ=WEEKLY;BYDAY=MO"]);
+  });
+
+  it("handles recurring-event instance fields (recurringEventId + originalStartTime)", () => {
+    const apiEvent: gapi.client.calendar.Event = {
+      id: "evt-instance_20260501T140000Z",
+      summary: "Weekly sync (moved)",
+      start: { dateTime: "2026-05-01T15:00:00Z" },
+      end: { dateTime: "2026-05-01T16:00:00Z" },
+      recurringEventId: "evt-parent",
+      originalStartTime: { dateTime: "2026-05-01T14:00:00Z" },
+    };
+
+    const result = normalizeFetchedEvent(apiEvent, "primary");
+
+    expect(result.recurringEventId).toBe("evt-parent");
+    expect(result.originalStartTime?.dateTime).toBe("2026-05-01T14:00:00Z");
+  });
+
+  it("accepts all-day events (date-only start/end with no dateTime)", () => {
+    const apiEvent: gapi.client.calendar.Event = {
+      id: "evt-allday",
+      summary: "Holiday",
+      start: { date: "2026-12-25" },
+      end: { date: "2026-12-26" },
+    };
+
+    const result = normalizeFetchedEvent(
+      apiEvent,
+      "holidays@group.v.calendar.google.com"
+    );
+
+    expect(result.start.date).toBe("2026-12-25");
+    expect(result.start.dateTime).toBeUndefined();
+    expect(result.end.date).toBe("2026-12-26");
+  });
+});
+
+describe("normalizeCalendarListEntry", () => {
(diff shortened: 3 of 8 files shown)


src/lib/google-calendar.ts
```diff
@@ -41,12 +52,119 @@ export interface GoogleCalendarEvent {
...
+ * Also exported so server-side callers that fetch `calendarList.list` via
+ * plain `fetch()` can reuse the same mapping without duplicating the field
+ * selection.
+ */
+export function normalizeCalendarListEntry(
+  entry: gapi.client.calendar.CalendarListEntry
+): UserCalendar {
+  return {
+    id: entry.id,
+    summary: entry.summary,
```
BenSeymourODB: **Missing `?? ""` fallback — inconsistency with `normalizeFetchedEvent`.**

`normalizeFetchedEvent` explicitly handles `event.summary ?? ""` because Google marks Event.summary as optional. `normalizeCalendarListEntry` does no such thing — it assigns `entry.summary` directly into `UserCalendar.summary: string` (a required field).

The Google Calendar API v3 schema does not mark CalendarList `summary` as required either. In practice Google always populates it, but if it were ever absent the result would be `undefined` assigned to a typed `string` field — a silent runtime lie TypeScript can't catch.

Either:
1. Mark `CalendarListEntry.summary` as optional in `gapi.d.ts` and add a `?? ""` fallback here (matching the Event treatment), or
2. Add a comment explaining why the fallback is intentionally omitted for this resource (and add a unit test to pin the assumption).

As written this asymmetry looks like an oversight rather than a deliberate decision.

---

src/lib/google-calendar.ts
```diff
@@ -41,12 +52,119 @@ export interface GoogleCalendarEvent {
...
+ *
+ * Exported for unit testing and for reuse by any code path that receives raw
+ * Google Calendar API responses (e.g. the server-side REST fetch in
+ * `src/app/api/calendar/events/route.ts`).
+ */
+export function normalizeFetchedEvent(
+  event: gapi.client.calendar.Event,
+  calendarId: string
+): GoogleCalendarEvent {
+  return {
```
BenSeymourODB: **`...event` spread leaks undeclared fields into the runtime object.**

`gapi.client.calendar.Event` declares fields that `GoogleCalendarEvent` does not: `kind`, `conferenceData`, `attachments`, `source`, `extendedProperties`, `attendeesOmitted`, `anyoneCanAddSelf`, `guestsCanInviteOthers`, `guestsCanModify`, `guestsCanSeeOtherGuests`, `privateCopy`, `locked`.

The spread `{ ...event, summary: ..., calendarId }` will include all of these in the returned object at runtime, but `GoogleCalendarEvent` doesn't declare them. TypeScript compiles this fine (excess properties in a return-type spread are silently dropped from the typed view), but:

1. The API response JSON (`NextResponse.json(allEvents)`) will include these extra fields — clients receive more than the TypeScript contract describes.
2. Before this PR, the old local `GoogleCalendarEvent` had `[key: string]: unknown`, which at least made the "there are extra fields" intent explicit. Removing that while still using a full spread creates a contract that is now *narrower in types than in reality*.

Options: either enumerate the fields explicitly in `normalizeFetchedEvent` (no spread), or add the missing fields to `GoogleCalendarEvent`, or re-add an index signature. The current state is arguably worse than the pre-PR code for this specific aspect.

---

src/lib/__tests__/google-calendar-mappers.test.ts
```diff
@@ -0,0 +1,262 @@
...
+import type { CalendarColorMapping } from "@/lib/calendar-storage";
+import {
+  type GoogleCalendarEvent,
+  type UserCalendar,
+  normalizeCalendarListEntry,
+  normalizeFetchedEvent,
+} from "@/lib/google-calendar";
+import { describe, expect, it } from "vitest";
+
+describe("normalizeFetchedEvent", () => {
```
BenSeymourODB, line 21: **TDD gap: the `summary ?? ""` fallback has no test.**

The comment on `normalizeFetchedEvent` says "Google marks `summary` optional; our UI expects a string so we fall back to the empty string and let `transformGoogleEvent` substitute a label." That's the core behavioural change in this PR on the events side — but there's no test for it.

Add a case to this `describe` block:

```ts
it('falls back to empty string when event has no summary', () => {
  const apiEvent: gapi.client.calendar.Event = {
    id: 'evt-no-title',
    start: { dateTime: '2026-05-01T09:00:00Z' },
    end:   { dateTime: '2026-05-01T09:30:00Z' },
  };
  const result = normalizeFetchedEvent(apiEvent, 'primary');
  expect(result.summary).toBe('');
});
```

Without this, removing or changing the `?? ""` fallback later would not break any test.

---

src/app/api/calendar/events/route.ts
```diff
@@ -4,27 +4,15 @@
  * Supports fetching from multiple calendars
  */
 import { AuthError, getAccessToken, getSession } from "@/lib/auth";
+import {
+  type GoogleCalendarEvent,
+  normalizeFetchedEvent,
```
BenSeymourODB, line 9: **First server-side import of `google-calendar.ts` — worth a note.**

Before this PR, no server route imported from `@/lib/google-calendar`. That module is documented as "client-side calendar fetching" and contains extensive `window.gapi.*` calls (lines ~198, ~305, ~414, etc.).

The import here is safe today: `normalizeFetchedEvent` is a pure function with no `window` references, and the module-level init code (`GOOGLE_CALENDAR_CONFIG`, `gapiInitialized = false`, etc.) is also harmless on the server. The module has no `"use client"` pragma so Next.js treats it as universal.

But this coupling is fragile. If someone later adds a module-level `window.*` reference to `google-calendar.ts` (a very natural thing to do given the file's stated purpose), it will break this route at runtime with `window is not defined`.

Suggestion: extract the pure mappers and shared types into a separate file (`src/lib/google-calendar-mappers.ts` or similar) that has no dependency on `gapi`/`window`, and import from there in both the server routes and the client-side `google-calendar.ts`. That file is the natural home for `normalizeFetchedEvent` and `normalizeCalendarListEntry` — no `window` risk, and the boundary between client-only and universal code becomes explicit.

---

src/types/gapi.d.ts
```diff
@@ -1,117 +1,381 @@
...
     };
   }
 
+  /**
+   * CalendarList resource entry.
+   * https://developers.google.com/workspace/calendar/api/v3/reference/calendarList#resource
+   */
   interface CalendarListEntry {
     id: string;
     summary: string;
```
BenSeymourODB: **`CalendarListEntry.summary` marked required — inconsistent with the API spec and with how `Event.summary` is treated.**

`gapi.client.calendar.Event.summary` is correctly `summary?: string` (optional) on line 203 of this file. But `CalendarListEntry.summary` is `summary: string` (required) here.

The Google Calendar API v3 CalendarList resource schema does not mark `summary` as required — it's described as "Title of the calendar" with no `required` flag. This is the same situation as Event.summary.

In practice Google always returns it for calendarList items, but the inconsistent treatment between the two types is confusing, and it means `normalizeCalendarListEntry` can't apply the same defensive `?? ""` pattern that `normalizeFetchedEvent` does.

Consider making this `summary?: string` to match the Event treatment and the actual schema.

---

src/app/api/calendar/calendars/route.ts
```diff
@@ -100,7 +87,7 @@ export async function GET() {
     }
 
     const data = await response.json();
-    const items: GoogleCalendarListItem[] = data.items || [];
+    const items: gapi.client.calendar.CalendarListEntry[] = data.items || [];
```
BenSeymourODB, line 90: **Type annotation on unvalidated JSON — same risk as before, but worth acknowledging.**

`data` is the result of `response.json()` — an unvalidated `any` at runtime. Annotating it as `gapi.client.calendar.CalendarListEntry[]` is a compile-time assertion with no runtime enforcement. If Google ever returns a malformed or unexpected shape (e.g., a calendar list item without a `summary`), TypeScript won't catch it and the downstream `item.summary` access on line 95 will produce `undefined` typed as `string`.

This was equally true of the old `GoogleCalendarListItem[]` annotation, so it's not a regression — but the richer new type may create a false sense of safety. A Zod schema or even a runtime `Array.isArray(data.items)` guard would be defensive without being expensive.

(Not blocking, but worth noting given the PR's stated goal of making the type surface more trustworthy.)

---

src/lib/__tests__/google-calendar-mappers.test.ts
```diff
@@ -0,0 +1,262 @@
...
+
+    expect(result.colorId).toBe("10");
+    expect(result.timeZone).toBe("America/New_York");
+    expect(result.summaryOverride).toBe("My Calendar");
+    expect(result.selected).toBe(true);
+    expect(result.accessRole).toBe("owner");
+  });
+});
+
+describe("type-level coverage: GoogleCalendarEvent is assignable from raw API responses", () => {
```
BenSeymourODB: **"type-level coverage" tests are mostly trivial at runtime.**

The comment correctly says these are "effectively compile-time assertions" — they will fail `pnpm check-types` if types drift, which is their real value. But as runtime tests they only assert things like `mapped[0]?.id === "evt-a"` which would pass against any correct implementation.

That's fine for the compile-time guard purpose, but consider either:
- Moving the type assertions inline (no `it()` wrapper needed — a top-level typed assignment in the test file is enough), or
- Adding at least one substantive runtime assertion per test (e.g., checking that `normalizeFetchedEvent` on an `EventsListResponse.items[0]` actually produces the right `calendarId`).

As-is the test count inflates the reported number without adding much runtime confidence.

---

BenSeymourODB (commented):
## Review summary

**Verdict: request changes on two real issues before merging.**

### Real blockers

1. **Undeclared runtime fields from the `...event` spread** (`src/lib/google-calendar.ts` line 134). `normalizeFetchedEvent` spreads `gapi.client.calendar.Event` but `GoogleCalendarEvent` doesn't declare all of its fields (`conferenceData`, `attachments`, `source`, `extendedProperties`, `kind`, `anyoneCanAddSelf`, `guestsCanInviteOthers`, `guestsCanModify`, `guestsCanSeeOtherGuests`, `privateCopy`, `locked`, `attendeesOmitted`). These flow through into the API response JSON invisibly. Before the PR, the `[key: string]: unknown` index signature at least made this explicit. The current state — narrower types than runtime reality — is arguably a regression for this specific concern.

2. **Missing `?? ""` fallback in `normalizeCalendarListEntry`** (`src/lib/google-calendar.ts` line 155). `normalizeFetchedEvent` defensively applies `event.summary ?? ""` because Google marks Event.summary optional. `normalizeCalendarListEntry` assigns `entry.summary` directly despite the same optional-in-spec situation for CalendarList items. Asymmetry that looks like an oversight.

### Missing test (TDD gap)

3. **No test for `normalizeFetchedEvent` with `summary: undefined`** (`src/lib/__tests__/google-calendar-mappers.test.ts`). The `?? ""` fallback — the key behavioural contract being added — has no test case. Changing or removing it later would not break anything.

### Should-fix before merge

4. **`CalendarListEntry.summary` marked required in `gapi.d.ts`** (line 277) while `Event.summary` is correctly optional (line 203). The Google API schema treats both the same way.

### Noteworthy (non-blocking)

5. **First server import of `google-calendar.ts`** (events/route.ts line 9). Safe today, but fragile coupling — the module is client-heavy. Pure mappers should ideally live in a separate universal file.

6. **Type assertion on unvalidated `response.json()` output** (`calendars/route.ts` line 90). Pre-existing pattern, not a regression, but the richer type creates false safety.

7. **"Type-level" tests are runtime no-ops** (`google-calendar-mappers.test.ts` line 179). Their value is as `check-types` guards; as runtime tests they assert only trivial properties.

### Clean

- No `any`, no manual memoization, no test-result artefacts committed, no non-standard Tailwind colors, no `use client` violation. The `colorId: ""` comment clarification (#77) is correct and the test pin is sensible. All 1021 tests reported passing.

---

src/lib/google-calendar.ts
BenSeymourODB: Fixed in 41c6b88. `GoogleCalendarEvent` now `extends Omit<gapi.client.calendar.Event, "summary">`, so the full schema is structurally included in the canonical type and the `{ ...event }` spread no longer leaks. Added a `passes every Google-declared field through the spread` test covering `kind`, `conferenceData`, `attachments`, `source`, `extendedProperties`, `attendeesOmitted`, `anyoneCanAddSelf`, `guestsCanInviteOthers`, `guestsCanModify`, `guestsCanSeeOtherGuests`, `privateCopy`, `locked`. If someone drops the `extends Omit<...>` relationship later, `check-types` fails on the test assertions, so the guarantee is pinned at both the type and runtime levels.

---

src/lib/google-calendar.ts
BenSeymourODB: Fixed in 41c6b88. `normalizeCalendarListEntry` now applies `entry.summary ?? ""`, matching `normalizeFetchedEvent`. `gapi.client.calendar.CalendarListEntry.summary` is also marked optional to reflect the Google schema (see the related comment on `gapi.d.ts`). Added a regression test (`falls back to empty string when calendarList entry has no summary`). The `calendars/route.ts` mapper picked up the same `?? ""` fallback where it projects into `CalendarInfo`.

---

src/lib/__tests__/google-calendar-mappers.test.ts
BenSeymourODB: Added in 41c6b88 — `falls back to empty string when event has no summary` exercises `normalizeFetchedEvent` with `summary: undefined` and asserts `result.summary === ""` plus a typed `string` assignment so the narrowing stays meaningful. Removing the `?? ""` fallback would fail both assertions.

---

src/types/gapi.d.ts
BenSeymourODB: Fixed in 41c6b88. `CalendarListEntry.summary` is now `summary?: string`, matching the treatment of `Event.summary` and the Google schema. The normaliser and the `calendars/route.ts` mapper both apply a `?? ""` fallback where they project into a `string`-typed field.

---

src/app/api/calendar/events/route.ts
BenSeymourODB: Fixed in 41c6b88. Extracted the canonical types (`GoogleCalendarEvent`, `UserCalendar`) and both pure mappers into a new universal module `src/lib/google-calendar-mappers.ts`. This module has no `window`/`gapi` runtime dependency (only ambient types). The `events/route.ts` and `calendars/route.ts` server handlers now import from it directly; `src/lib/google-calendar.ts` re-exports the same symbols for existing callers. Adding browser-only code to `google-calendar.ts` can no longer break the server routes.

---

src/lib/__tests__/google-calendar-mappers.test.ts
BenSeymourODB: Reworked in 41c6b88. The `describe("type-level coverage: …")` block is now `describe("end-to-end: enriched API responses flow through the mappers", …)` — each test assigns a realistic `EventsListResponse` / `CalendarListListResponse` fixture (still a compile-time guard on the response types) and then asserts specific mapped fields on the normaliser output (`location`, `htmlLink`, `calendarId`, `attendees[0].email`, `accessRole`, `timeZone`, `selected`, etc.). So the runtime assertions now reflect what the mappers actually produce, and the compile-time guard persists.

---

src/app/api/calendar/calendars/route.ts
BenSeymourODB: Acknowledged — deliberately leaving this out of scope for this PR. The annotation on `response.json()` is a pre-existing pattern across all three calendar routes, and layering a runtime guard (Zod or manual) would touch shared concerns (error shape, logging, response schema) that go beyond the #76/#77 cleanup. Happy to open a follow-up issue for a runtime validation pass on the calendar-API responses if you'd like. For this PR the `?? ""` fallback now stitches the one narrow-to-wide assignment that motivated this comment, so a malformed upstream response won't silently produce a `undefined`-typed `string` in `CalendarInfo.summary`.

---

BenSeymourODB:

---

BenSeymourODB (commented):
## QA verdict — pass

Validated locally against the acceptance criteria from issues #76 and #77.

### Acceptance criteria

| # | Criterion | Result |
|---|-----------|--------|
| AC1 (#76) | Types verified against Google API docs; extended fields declared with doc comments | ✅ `gapi.d.ts` fully rewritten, all fields cross-referenced to Google Calendar v3 reference |
| AC2 (#76) | Duplicate partial interfaces across server routes eliminated | ✅ `GoogleCalendarListItem` removed from all three routes; canonical `gapi.client.calendar.*` types used instead |
| AC3 (#76) | Custom type extensions documented | ✅ Module-level JSDoc in `gapi.d.ts` with direct links to Google API reference |
| AC4 (#77) | `colorId: ""` comment clarified | ✅ Comment now explains why: event-level `colorId` doesn't exist on calendarList entries |
| AC5 | All tests pass | ✅ **1024 / 1024** (75 files) — 3 more than the PR body's 1021, from test cases added in the review-fix commit `41c6b88` |
| AC6 | `check-types` clean | ✅ Zero errors |

### Review concerns — all addressed in `41c6b88`

- **Undeclared runtime fields from `...event` spread** — `GoogleCalendarEvent` now `extends Omit<gapi.client.calendar.Event, "summary">` so the full schema is structurally included ✅
- **Missing `?? ""` fallback in `normalizeCalendarListEntry`** — added, matching `normalizeFetchedEvent` ✅
- **Missing test for `normalizeFetchedEvent` with `summary: undefined`** — `"falls back to empty string when event has no summary"` test added ✅
- **`CalendarListEntry.summary` marked required** — changed to `summary?: string` to match the Google schema ✅
- **Fragile server import of `google-calendar.ts`** — pure mappers extracted to `google-calendar-mappers.ts`; server routes import from the universal module ✅
- **"Type-level" tests are runtime no-ops** — reworked as end-to-end tests with real runtime assertions ✅

### No browser validation required

Pure type / infrastructure refactor — no React components, CSS, page routes, or user-facing copy changed.

**Repro:** `pnpm test src/lib/__tests__/google-calendar-mappers.test.ts src/app/api/calendar/events/__tests__/route.test.ts`

🤖 qa-tester

---

PR #160, opened 2026-04-22
feat(http): retry with exponential backoff for transient errors (#68)

### LINKED ISSUE #68: Incomplete error recovery flows
## Description

Some error scenarios don't have graceful recovery, requiring users to manually refresh or clear storage.

## Examples

- Network errors during OAuth flow
- IndexedDB quota exceeded
- Browser storage unavailable
- Multiple simultaneous auth attempts
- API rate limiting

## Improvements Needed

- Retry logic for network errors
- Quota management for IndexedDB
- Fallback to in-memory storage if IndexedDB unavailable
- Queue for simultaneous auth requests
- Rate limiting detection and backoff

## Priority

Medium — add as issues are discovered in production use.

---
*Migrated from `.claude/technical-debt.md` (High Priority #3)*

BenSeymourODB:
## Summary

Addresses the first bullet of #68 — _retry logic for network errors_.

Ships a pure `withRetry` / `fetchWithRetry` utility and will wire it into the three Google Tasks server routes so that rate limits (429) and transient 5xx / network failures are transparently retried with exponential backoff + jitter, instead of being surfaced to the user as a one-shot error.

Closes #68 (partial — see "Deferred" below).

### `src/lib/http/retry.ts`

- **`withRetry<T>(fn, options)`** — generic retry wrapper with exponential backoff + configurable jitter (`full` / `none`), `AbortSignal` support, an `onRetry` observer, and injectable `sleep` / `random` primitives for deterministic tests.
- **`fetchWithRetry(input, init, options)`** — drop-in `fetch` replacement. Retries transient failures; leaves 2xx, 401, 404, etc. untouched so callers keep their existing `!response.ok` branches.
- **`isTransientHttpError`** — strict classifier: 429 + 5xx + `fetch` `TypeError` (network failures) only. Non-429 4xx stays permanent so retries never mask auth / validation bugs. AbortError is explicitly _not_ transient — if the caller aborted, we stop.
- **`parseRetryAfter`** — handles delay-seconds _and_ HTTP-date forms, clamps past-dates to 0, rejects malformed inputs like `"-5"` (which `Date.parse` happily accepts as year -5).
- **`HttpRetryError`** — typed error threading status + Retry-After through the retry loop.

### Test coverage

40 Vitest cases in `src/lib/http/__tests__/retry.test.ts` cover every branch:

- Backoff math: exponential growth, `maxDelayMs` cap, full-jitter bounds, Retry-After override + cap, negative Retry-After clamped to 0.
- `parseRetryAfter`: null / empty / whitespace / integer / float / negative / invalid date / past date / future date.
- Classification: 429, 5xx range (500–599), 4xx non-429, network errors, AbortError, non-Error values.
- `withRetry`: first-success path (no sleep), exhaustion, non-transient short-circuit, custom `shouldRetry`, `onRetry` observer metadata, Retry-After honoured, pre-armed abort, mid-loop abort, default `maxAttempts: 3`.
- `fetchWithRetry`: ok passthrough, non-transient passthrough, 5xx retry succeeds, 429 Retry-After respected, all-retries-exhausted still returns final transient Response, network-error retry, network-error re-thrown after exhaustion, outer vs inner signal precedence.

All 1051 tests pass (`pnpm test`).

## Phase 2 (next commit on this PR)

Integrate `fetchWithRetry` into the three Google Tasks server routes:

- `src/app/api/tasks/route.ts` (GET + POST)
- `src/app/api/tasks/lists/route.ts` (GET)
- `src/app/api/tasks/[taskId]/route.ts` (PATCH)

Plus tests asserting that transient failures retry and non-transient ones don't.

## Scope guardrail

Three open PRs touch the calendar / auth surfaces where retry would also be valuable:

- #158 (encrypt OAuth tokens at rest) — `src/lib/auth/auth.ts`, token-refresh path
- #159 (Google Calendar types + mappers) — `src/app/api/calendar/**`, `src/lib/google-calendar.ts`
- #155 (duplicate-Google-account guard) — `src/lib/auth/auth.ts`

To avoid conflicts with that in-flight work, this PR deliberately leaves calendar + auth call sites alone and integrates only into the Google Tasks routes, which no open PR touches.

## Deferred (follow-ups)

- Adopt `fetchWithRetry` in `src/app/api/calendar/events/route.ts` and sibling calendar routes once #159 lands.
- Adopt `fetchWithRetry` in the Google OAuth token-refresh `fetch` inside `src/lib/auth/auth.ts` once #158 lands.
- IndexedDB quota management + in-memory fallback (separate concern, client-side — worth a dedicated issue).
- Simultaneous-auth-request queue.

## Test plan

- [x] `pnpm vitest run src/lib/http` — 40/40 pass
- [x] `pnpm test` — 1051/1051 pass
- [ ] Phase 2 commit integrates the utility + adds route-level retry tests

Closes #68

### DIFF
--- a/src/app/api/tasks/[taskId]/__tests__/route.test.ts
+++ b/src/app/api/tasks/[taskId]/__tests__/route.test.ts
@@ -37,6 +37,27 @@ vi.mock("@/lib/logger", () => ({
   },
 }));
 
+// Replace the real `sleep` inside fetchWithRetry with a no-op so transient-
+// status tests (5xx, 429) don't wait for real backoff delays.
+vi.mock("@/lib/http/retry", async () => {
+  const actual =
+    await vi.importActual<typeof import("@/lib/http/retry")>(
+      "@/lib/http/retry"
+    );
+  return {
+    ...actual,
+    fetchWithRetry: (
+      input: Parameters<typeof actual.fetchWithRetry>[0],
+      init?: Parameters<typeof actual.fetchWithRetry>[1],
+      options: Parameters<typeof actual.fetchWithRetry>[2] = {}
+    ) =>
+      actual.fetchWithRetry(input, init, {
+        ...options,
+        sleep: () => Promise.resolve(),
+      }),
+  };
+});
+
 // Mock global fetch
 const mockFetch = vi.fn();
 global.fetch = mockFetch;
--- a/src/app/api/tasks/[taskId]/__tests__/route.test.ts
+++ b/src/app/api/tasks/[taskId]/__tests__/route.test.ts
@@ -333,5 +333,44 @@ describe("/api/tasks/[taskId]", () => {
       expect(status).toBe(200);
       expect(data.task.status).toBe("needsAction");
     });
+
+    it("retries a transient 503 on PATCH and returns the eventual 200 (issue #68)", async () => {
+      vi.mocked(getSession).mockResolvedValue(mockSession);
+      vi.mocked(getAccessToken).mockResolvedValue(
+        mockGoogleAccount.access_token!
+      );
+
+      const updatedTask = { id: taskId, title: "Test", status: "completed" };
+      mockFetch
+        .mockResolvedValueOnce({
+          ok: false,
+          status: 503,
+          headers: new Headers(),
+          json: () => Promise.resolve({ error: "Service Unavailable" }),
+        })
+        .mockResolvedValueOnce({
+          ok: true,
+          json: () => Promise.resolve(updatedTask),
+        });
+
+      const request = createMockRequest(
+        `/api/tasks/${taskId}?listId=${listId}`,
+        { method: "PATCH", body: { status: "completed" } }
+      );
+      const response = await PATCH(request, {
+        params: Promise.resolve({ taskId }),
+      });
+      const { status, data } = await parseResponse<{
+        task: typeof updatedTask;
+      }>(response);
+
+      expect(status).toBe(200);
+      expect(data.task).toEqual(updatedTask);
+      expect(mockFetch).toHaveBeenCalledTimes(2);
+      // PATCH body is a string (JSON.stringify) — retry re-sends it verbatim.
+      const [[, firstInit], [, secondInit]] = mockFetch.mock.calls;
+      expect(firstInit?.body).toBe(secondInit?.body);
+      expect(firstInit?.method).toBe("PATCH");
+    });
   });
 });
--- a/src/app/api/tasks/[taskId]/route.ts
+++ b/src/app/api/tasks/[taskId]/route.ts
@@ -3,6 +3,7 @@
  * Uses server-side authentication with NextAuth.js
  */
 import { AuthError, getAccessToken, getSession } from "@/lib/auth";
+import { fetchWithRetry } from "@/lib/http/retry";
 import { logger } from "@/lib/logger";
 import { NextRequest, NextResponse } from "next/server";
 
@@ -71,7 +72,7 @@ export async function PATCH(
 
     const accessToken = await getAccessToken();
 
-    const response = await fetch(
+    const response = await fetchWithRetry(
       `${GOOGLE_TASKS_API}/lists/${encodeURIComponent(listId)}/tasks/${encodeURIComponent(taskId)}`,
       {
         method: "PATCH",
--- a/src/app/api/tasks/__tests__/route.test.ts
+++ b/src/app/api/tasks/__tests__/route.test.ts
@@ -38,6 +38,28 @@ vi.mock("@/lib/logger", () => ({
   },
 }));
 
+// Replace the real `sleep` inside fetchWithRetry with a no-op so transient-
+// status tests (5xx, 429) don't wait for real backoff delays. All other
+// retry behaviour stays untouched.
+vi.mock("@/lib/http/retry", async () => {
+  const actual =
+    await vi.importActual<typeof import("@/lib/http/retry")>(
+      "@/lib/http/retry"
+    );
+  return {
+    ...actual,
+    fetchWithRetry: (
+      input: Parameters<typeof actual.fetchWithRetry>[0],
+      init?: Parameters<typeof actual.fetchWithRetry>[1],
+      options: Parameters<typeof actual.fetchWithRetry>[2] = {}
+    ) =>
+      actual.fetchWithRetry(input, init, {
+        ...options,
+        sleep: () => Promise.resolve(),
+      }),
+  };
+});
+
 // Mock global fetch
 const mockFetch = vi.fn();
 global.fetch = mockFetch;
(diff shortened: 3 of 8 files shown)


BenSeymourODB (commented):
Overall this is solid work — the design is clean, the public API is well-typed, the backoff math is correct, and the test suite is impressively thorough. There are two real bugs worth fixing before merge, plus a few lower-priority issues.

---

## Bug 1 — Stale response returned when a mixed retry sequence ends with a network error (`src/lib/http/retry.ts`, `fetchWithRetry`)

**Severity: Medium — silent wrong-result in a realistic scenario.**

When retries alternate between transient HTTP responses (503) and network errors (`TypeError`), the catch block can return a stale `lastResponse` from a *previous* attempt rather than re-throwing the network error.

Concrete scenario with `maxAttempts: 3`:
- Attempt 1: `fetch()` returns 503 → `lastResponse = 503 response`, throws `HttpRetryError`
- Attempt 2: `fetch()` throws `TypeError("fetch failed")` → `withRetry` catches, retries
- Attempt 3: `fetch()` throws `TypeError("fetch failed")` → `withRetry` exhausts, throws `TypeError`

The outer `catch` in `fetchWithRetry` receives the `TypeError`. It then checks:

```ts
if (lastResponse && !lastResponse.ok && shouldRetryResponse(lastResponse)) {
  return lastResponse;  // BUG: returns the stale 503 from attempt 1
}
throw error;
```

Because `lastResponse` is still holding the 503 from attempt 1, the condition is `true` and the network error is silently swallowed. The caller sees a 503 `Response` and tries to call `.json()` on it — but on a real (non-mock) `Response` the body stream is already consumed.

This bug is invisible in tests because mock responses always return fresh `.json()` stubs regardless of how many times they're called, and the existing "re-throws a thrown network error after exhausting retries" test only exercises all-network-error sequences (where `lastResponse` is always `undefined`).

**Fix:** only fall back to `lastResponse` when the error is our own `HttpRetryError`, meaning the retry machinery intentionally escalated a response-based failure:

```ts
} catch (error) {
  if (error instanceof HttpRetryError && lastResponse && !lastResponse.ok) {
    return lastResponse;
  }
  throw error;
}
```

---

## Bug 2 — `isTransientHttpError` over-classifies programming `TypeError`s as transient (`src/lib/http/retry.ts:101`)

**Severity: Low-medium — incorrect retries mask bugs, waste API quota.**

```ts
if (error instanceof TypeError) return true;
```

`fetch` throws `TypeError` for network failures, but also for programming errors: invalid header values, malformed URL strings, etc. Those should fail fast, not be retried 3 times with backoff. A misconfigured `Authorization` header that produces a `TypeError` will silently retry and burn quota before surfacing.

A tighter check is:

```ts
if (error instanceof TypeError && error.message.includes("fetch")) return true;
```

Or check `error.cause` (Node 18+). At minimum, the current broad match should be documented as a known limitation.

---

## Issue 3 — `fetchWithRetry` is unsafe for `Request` object inputs that carry a body (`src/lib/http/retry.ts:246`)

**Severity: Low for this PR (all three routes use string URL + `JSON.stringify` body), but a latent trap.**

When `input` is a `Request` with a body, calling `fetch(request)` consumes its `ReadableStream`. On retry the same `Request` is reused — the body is already consumed and the second call sends an empty body.

The current routes are safe (string body), but `fetchWithRetry` is exported as a generic drop-in. The doc comment should warn callers, or the implementation should clone:

```ts
const response = await fetch(
  input instanceof Request ? input.clone() : input,
  effectiveInit
);
```

---

## Issue 4 — No retry integration tests for `POST /api/tasks`, `PATCH /api/tasks/[taskId]`, or `GET /api/tasks/lists`

The PR adds the `fetchWithRetry` mock boilerplate to all three test files but only adds actual retry-scenario tests to `tasks/__tests__/route.test.ts` (GET path). Three integrated call sites have zero retry coverage:

- `POST /api/tasks` — mutates with a JSON body; no 503/429 retry test
- `PATCH /api/tasks/[taskId]` — mutates with a JSON body; no 503/429 retry test
- `GET /api/tasks/lists` — no 503/429 retry test

Per the CLAUDE.md TDD requirement, each call site should have at least one test asserting the retry path works. The mutation routes are especially important since they carry a body.

---

## Minor — `maxAttempts` resolved redundantly in `fetchWithRetry` (`src/lib/http/retry.ts:258`)

`fetchWithRetry` reads `options.maxAttempts` itself (line 258) to use in the inner guard at line 272. `withRetry` reads the same value independently. There's no bug — both agree because they share the same `options` object — but a comment explaining why `fetchWithRetry` needs its own copy (to decide between "throw `HttpRetryError`" vs "return the transient response" on the final attempt) would help future readers.

---

## Not a bug: `POST`/`PATCH` body reuse is safe

Both mutation routes call `JSON.stringify(...)` before passing the body to `fetchWithRetry`, producing a plain `string`. Strings are not streams; `fetch` re-reads them on every retry without cloning. No action needed for the current routes.

---

BenSeymourODB:
Thanks for the review — all six findings addressed in c63c1ff.

**Bug 1 (stale response on mixed-sequence network final)** — Fixed. The outer catch now only substitutes `lastResponse` when the caught error is an `HttpRetryError`; network and abort errors propagate. Regression test added: mixed `503 → TypeError → TypeError` rejects with the network error and calls fetch 3 times (`src/lib/http/__tests__/retry.test.ts` — "re-throws a network error even when a prior attempt returned a transient response").

**Bug 2 (over-classifying programming TypeErrors)** — Fixed. `isTransientHttpError` now requires the TypeError message to contain "fetch" (matches Node "fetch failed", Chrome "Failed to fetch", Firefox "NetworkError when attempting to fetch resource"). Programming errors like "Invalid URL" / "Header value contains invalid bytes" / "foo is not a function" now fail fast. Three new positive + three new negative tests.

**Issue 3 (Request input body consumed on retry)** — Fixed. `fetchWithRetry` now clones `input` per attempt when it is a `Request`; string/URL inputs are passed through unchanged. Regression test asserts each retry receives a distinct Request instance.

**Issue 4 (missing retry tests on POST/PATCH/lists-GET)** — Added. Three new integration tests:
- `POST /api/tasks`: transient 503 → eventual 201, asserts body string re-sent verbatim
- `PATCH /api/tasks/[taskId]`: transient 503 → eventual 200, asserts body + method preserved
- `GET /api/tasks/lists`: transient 503 → eventual 200, plus a no-retry-on-401 assertion

**Issue 5 (clarifying comment on `maxAttempts` duplication)** — Added an explanatory block on why `fetchWithRetry` resolves its own copy (to decide between "throw `HttpRetryError`" and "return the transient Response on the final attempt") while sharing the options object with `withRetry`.

Also tightened the pre-existing lists-GET 500-bubble-up test mock to include `headers: new Headers()`. Real `Response` objects always carry headers; the mock was under-specified for the post-integration world where `fetchWithRetry` reads `Retry-After`. The test's intent ("500 bubbles up as 500 through the `!ok` branch") is unchanged — retries exhaust, final response still surfaces with status 500.

Suite: 1061 tests pass (was 1054, +7 new).

---

BDSeymour:
🤖 qa-tester reviewing — local validation in progress.

BDSeymour (approved):
## QA verdict — pass

Backend-only change (`src/lib/http/retry.ts` + 3 Google Tasks routes + tests). No UI surface, so per the QA skill rules I skipped browser validation and validated by acceptance + tests.

### Acceptance against #68

The PR explicitly closes #68 partial — bullets 1 and 5 of the issue are landed; bullets 2/3/4 (IndexedDB quota, in-memory fallback, auth-request queue) are deferred with rationale in the PR body. That deferral matches how the other open PRs are scoped, and the deferred items are independent enough to land separately.

| Issue #68 improvement | Status |
|---|---|
| Retry logic for network errors | ✅ `withRetry` / `fetchWithRetry`, TypeError narrowed to `/fetch/i` so programming errors surface |
| Rate limiting detection and backoff | ✅ 429 + 5xx classified transient, `Retry-After` honoured (seconds + HTTP-date), exponential + full jitter, `maxDelayMs` cap |
| IndexedDB quota / in-memory fallback / auth queue | Deferred with explicit follow-up notes |

### Round-1 review fixes — all verified in code

1. **Bug 1 (stale-response)** — `src/lib/http/retry.ts:308-316` now substitutes `lastResponse` only when the caught error is `HttpRetryError`; network/abort errors propagate. Regression test in `retry.test.ts` covers the mixed `503 → TypeError → TypeError` sequence.
2. **Bug 2 (TypeError over-classification)** — `src/lib/http/retry.ts:112` narrows to `error instanceof TypeError && /fetch/i.test(error.message)`. Programming TypeErrors ("Invalid URL", etc.) now fail fast.
3. **Issue 3 (Request body consumption)** — `src/lib/http/retry.ts:279` clones the input per attempt when it is a `Request`.
4. **Issue 4 (route retry coverage)** — `tasks/__tests__/route.test.ts:421` (POST 503→201), `tasks/lists/__tests__/route.test.ts:203` (lists GET 503→200), `tasks/[taskId]/__tests__/route.test.ts:337` (PATCH 503→200) all assert retry + body preservation, plus paired non-retry-on-401 cases.

### Local validation

| Check | Result |
|---|---|
| `pnpm test:run` | 1061/1061 pass (75 files) |
| `pnpm test:run src/lib/http` | 43/43 retry tests pass |
| `pnpm test:run src/app/api/tasks` | 59/59 pass across 5 task-route files |
| `pnpm lint` | 0 errors (3 pre-existing warnings, unrelated) |

**Repro:** `pnpm test:run src/lib/http src/app/api/tasks`

🤖 qa-tester

Issue #170, opened 2026-04-27
feat(meals): add AddMealModal with day/type/recipe/notes


BDSeymour:
Modal launched when an empty meal slot is clicked or a meal card is edited.

## Scope

- `src/components/meals/add-meal-modal.tsx` — fields: day (preselected from slot), meal type (preselected from slot), meal name (required), servings (default 4), notes, optional recipe link search (UI only — recipes track is post-MVP, so this can be a disabled/hidden field if recipe data is unavailable).
- Save / Cancel; on save, calls `addMeal` from `MealContext`.
- Accessibility: focus trap, ESC to close, ARIA dialog.

## Acceptance

- Validation: empty meal name shows inline error; cannot save.
- Test page wired so `/test/meals` exercises open → fill → save → see card.
- Component tests cover validation + happy path.

## Blocked by

- Meals API routes issue.
- MealPlanner grid + MealContext issue.

## Plan reference

`.claude/plans/meal-planning.md` (section "Add Meal Modal").

## Track

Day 9 (Meals track).

Issue #169, opened 2026-04-27
feat(meals): add MealPlanner grid component + MealContext + MealCard/MealSlot


BDSeymour:
Build the weekly meal-planning UI: a 7-day × 4-meal-type grid with week navigation, today highlight, and per-slot add/edit/delete affordances.

## Scope

- `src/components/meals/contexts/meal-context.tsx` — `currentWeek`, `weekPlan`, `groceryList`, `setWeek`, `refreshWeekPlan`, `addMeal`, `removeMeal`, `updateMeal`, `generateGroceryList`.
- `src/components/meals/meal-planner.tsx` — grid rendering using a table or CSS grid with `Mon..Sun` columns × `Breakfast/Lunch/Dinner/Snack` rows.
- `src/components/meals/meal-card.tsx` — meal title + emoji, servings, edit/delete actions, expand-on-click for notes/recipe link.
- `src/components/meals/meal-slot.tsx` — empty-slot affordance with a `+` button.
- Test page `src/app/test/meals/page.tsx` with mock data, mirroring the existing `/test/calendar` pattern.

## Acceptance

- Today's meal slots are visually highlighted.
- Prev / Next week buttons advance by 7 days; week label shows the date range.
- Component tests for `MealCard` (expand/collapse) and `MealPlanner` (grid render with mock data).
- No DnD in v1 — that's deferred to a future issue.

## Blocked by

- Meals API routes issue (Meals track).

## Plan reference

`.claude/plans/meal-planning.md` (sections "Meal Card Component", "Meal Planner Grid Component", "Meal Context").

## Track

Day 9 (Meals track).

Issue #168, opened 2026-04-27
feat(meals): add API routes for week plan + grocery list


BDSeymour:
Add the API surface needed by the meal-planner UI.

## Routes

- `GET /api/meals/plan/[week]` — fetch all `PlannedMeal`s + linked `Meal` for the requested week (Monday-anchored ISO date string), plus the `groceryListId` if one exists.
- `POST /api/meals/plan` — create a meal and its `PlannedMeal` for `(weekStart, dayOfWeek, mealType)`. Upsert by the unique tuple.
- `PATCH /api/meals/plan/[id]` — update a planned meal (move slot, edit servings/notes).
- `DELETE /api/meals/plan/[id]` — remove a planned meal.
- `POST /api/meals/grocery-list` — generate a grocery list for a week by aggregating `MealIngredient`s by name.

## Acceptance

- Auth-required; 401 if unauthenticated.
- 400 with field-level errors for invalid `weekStart` or unknown enum values.
- Logger events: `MealAdded`, `MealUpdated`, `MealRemoved`, `GroceryListGenerated`.
- Integration tests run against a real test DB (per project policy of no DB mocks for integration).

## Blocked by

- Meals Prisma schema issue (Meals track).

## Plan reference

`.claude/plans/meal-planning.md` (section "API Routes (Key Examples)").

## Track

Day 9 (Meals track).

PR #193, opened 2026-05-02
feat(meals): add Prisma schema for meal planning (#167)

### LINKED ISSUE #167: feat(meals): add Prisma schema for meal planning (Meal, PlannedMeal, MealIngredient, GroceryList)
Add the database foundation for the meal-planning track. This is a single migration covering all the new models so the API and UI issues that follow have something to bind against.

## Scope (schema)

- `enum MealType { breakfast, lunch, dinner, snack }` and `enum DayOfWeek`.
- `Meal { id, userId, profileId?, name, type, servings, notes?, recipeId?, timestamps }` with indexes on `(userId, type)`.
- `PlannedMeal { id, userId, mealId, weekStart, dayOfWeek, mealType, createdAt }` with `@@unique([userId, weekStart, dayOfWeek, mealType])`.
- `MealIngredient { id, mealId, name, quantity, category, order }`.
- `GroceryList { id, userId, weekStart, name, timestamps }` + `GroceryListItem { id, listId, name, quantity, category, isChecked, mealId?, order, createdAt }`.
- `SavedMeal { id, userId, name, type, servings, recipeId?, isFavorite, useCount, lastUsed, createdAt }` (library) and `MealTemplate { id, name, description, meals (Json), isPublic, createdBy?, createdAt }`.

## Acceptance

- Migration created via `pnpm db:migrate` (NOT `prisma db push`, per CLAUDE.md).
- Prisma generated types regenerated.
- A short unit/integration test exercises one create + one read for `PlannedMeal` and `GroceryList`.

## Plan reference

`.claude/plans/meal-planning.md` (sections "Database Schema" and "Data Models").

## Track

Day 9 (Meals track). Blocks the Meals API routes and UI issues below.

BenSeymourODB:
Closes #167.

Adds the database foundation + create/read service for the weekly meal-planning track so the API and UI slices (#168, #169, #170) have something to bind against.

## Phase 1 — schema + migration

**Enums**
- `MealType { breakfast, lunch, dinner, snack }`
- `DayOfWeek { monday … sunday }`

**Models**
- `Meal` — owner (`userId`, cascade), optional `profileId` (set null on profile delete), `name`, `type`, `servings`, `notes`, `recipeId?`, timestamps. `@@index([userId, type])`.
- `PlannedMeal` — `userId` + `mealId`, `weekStart` (Monday UTC midnight), `dayOfWeek`, `mealType`. `@@unique([userId, weekStart, dayOfWeek, mealType])` so a slot can only hold one meal. `@@index([userId, weekStart])`.
- `MealIngredient` — belongs to `Meal` (cascade). `name`, `quantity`, `category`, `order`. `@@index([mealId])`.
- `GroceryList` — owner cascade, `weekStart`, `name`, timestamps. `@@index([userId, weekStart])`.
- `GroceryListItem` — belongs to `GroceryList` (cascade), `name`, `quantity`, `category`, `isChecked`, optional `mealId` (plain id; the source meal may have been deleted), `order`, `createdAt`. `@@index([listId, category])`.
- `SavedMeal` — meal library. `userId`, `name`, `type`, `servings`, `recipeId?`, `isFavorite`, `useCount`, `lastUsed`. `@@index([userId, isFavorite])`, `@@index([userId, lastUsed])`.
- `MealTemplate` — `name`, `description`, `meals` (JSON `TemplateMeal[]`), `isPublic`, optional `createdBy` (set null on user delete). `@@index([isPublic])`.

**Migration**
- `prisma/migrations/0005_add_meal_planning/migration.sql` — hand-authored to match Prisma's output format (matches `0001_initial` conventions). All `CREATE TYPE`, `CREATE TABLE`, `CREATE INDEX`, and `ADD CONSTRAINT` statements.
- Per CLAUDE.md, no `prisma db push` was used.

**Schema/migration tests** (`prisma/__tests__/migration-setup.test.ts`)
- New `meal planning schema (issue #167)` describe block: enums declared, models exist with required fields, `Meal`'s `(userId, type)` index, `PlannedMeal`'s composite unique constraint, `GroceryList` / `GroceryListItem` shape, and the 0005 migration's `CREATE TABLE` / `CREATE TYPE` / `CREATE UNIQUE INDEX` statements.
- Fix: the existing `should reference the same models in schema and migration` test only scanned `0001_initial`. It always intended to verify "every model in the schema has a `CREATE TABLE` somewhere", but couldn't see new migrations. Now scans all migration files via the same combined-SQL helper used by the Account uniqueness test. Strengthening, not weakening.

## Phase 2 — service layer + create/read tests

`src/lib/services/meal-planning.ts` exposes four entry points that future API routes (#168) will call:

- `createPlannedMeal({ userId, mealId, weekStart, dayOfWeek, mealType }) → PlannedMeal`
- `getPlannedMealsForWeek(userId, weekStart) → PlannedMealWithMeal[]`
- `createGroceryList({ userId, weekStart, name? }) → GroceryList`
- `getGroceryListForWeek(userId, weekStart) → GroceryListWithItems | null`

All four normalize the caller-supplied `weekStart` to the Monday of that week at UTC midnight via a private `getMondayOfWeek` helper. The `@@unique([userId, weekStart, dayOfWeek, mealType])` constraint on `PlannedMeal` is keyed on `weekStart`, so callers across the API and UI must agree on the same anchor or they'd silently land on different slots.

12 service-layer tests (`src/lib/services/__tests__/meal-planning.test.ts`) match the existing `reward-points.ts` pattern — a single prisma mock, no auth, no `NextRequest`, no `fetch`. Coverage:

- create + read for `PlannedMeal`
- create + read for `GroceryList`
- name fallback / override on `createGroceryList`
- empty result on `getPlannedMealsForWeek`
- null on `getGroceryListForWeek` when no list exists
- `it.each` table verifying Mon/Tue/Sun/Thu inputs all resolve to the same Monday

## Out of scope (deferred)

- API routes (#168), `MealPlanner` grid (#169), `AddMealModal` (#170)
- `Recipe` model — `recipeId` stays a plain optional `String` until the recipe slice lands
- Drag-and-drop, copy-week, grocery-list generation logic — UI tickets

## Verification

- `pnpm lint:fix` — 0 errors (4 pre-existing warnings unrelated to this PR)
- `pnpm format:fix` — passes
- `pnpm check-types` — passes
- `pnpm test` — 1453 tests, 99 files, all passing (+12 new service tests, +6 new schema tests)
- `pnpm prisma validate` — schema valid
- `pnpm prisma format` — schema is formatted

## Test plan

- [x] Schema and migration tests pass locally
- [x] Service-layer tests pass locally (acceptance criterion satisfied)
- [ ] Reviewer can run `pnpm db:migrate:status` against a clean DB and see `0005_add_meal_planning` listed
- [ ] CI green

https://claude.ai/code/session_01AFyDpdmyRMXXQJYTJRMrUJ

### DIFF
--- a/prisma/__tests__/migration-setup.test.ts
+++ b/prisma/__tests__/migration-setup.test.ts
@@ -183,13 +183,23 @@ describe("Account uniqueness constraints (issue #61)", () => {
   });
 });
 
+function readAllMigrationsCombined(): string {
+  const entries = fs.readdirSync(MIGRATIONS_DIR, { withFileTypes: true });
+  return entries
+    .filter((e) => e.isDirectory() && e.name !== "__tests__")
+    .map((e) =>
+      fs.readFileSync(
+        path.join(MIGRATIONS_DIR, e.name, "migration.sql"),
+        "utf-8"
+      )
+    )
+    .join("\n");
+}
+
 describe("schema and migration consistency", () => {
-  it("should reference the same models in schema and migration", () => {
+  it("should reference the same models across all migrations", () => {
     const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
-    const migrationContent = fs.readFileSync(
-      path.join(MIGRATIONS_DIR, "0001_initial", "migration.sql"),
-      "utf-8"
-    );
+    const combinedSql = readAllMigrationsCombined();
 
     // Extract model names from schema
     const modelRegex = /^model\s+(\w+)\s*\{/gm;
@@ -199,21 +209,18 @@ describe("schema and migration consistency", () => {
       schemaModels.push(match[1]);
     }
 
-    // Every schema model should have a CREATE TABLE in the migration
+    // Every schema model should have a CREATE TABLE in some migration
     for (const model of schemaModels) {
       expect(
-        migrationContent,
-        `Migration is missing CREATE TABLE for model "${model}"`
+        combinedSql,
+        `Migrations are missing CREATE TABLE for model "${model}"`
       ).toContain(`CREATE TABLE "${model}"`);
     }
   });
 
-  it("should reference the same enums in schema and migration", () => {
+  it("should reference the same enums across all migrations", () => {
     const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
-    const migrationContent = fs.readFileSync(
-      path.join(MIGRATIONS_DIR, "0001_initial", "migration.sql"),
-      "utf-8"
-    );
+    const combinedSql = readAllMigrationsCombined();
 
     // Extract enum names from schema
     const enumRegex = /^enum\s+(\w+)\s*\{/gm;
@@ -223,12 +230,143 @@ describe("schema and migration consistency", () => {
       schemaEnums.push(match[1]);
     }
 
-    // Every schema enum should have a CREATE TYPE in the migration
+    // Every schema enum should have a CREATE TYPE in some migration
     for (const enumName of schemaEnums) {
       expect(
-        migrationContent,
-        `Migration is missing CREATE TYPE for enum "${enumName}"`
+        combinedSql,
+        `Migrations are missing CREATE TYPE for enum "${enumName}"`
       ).toContain(`CREATE TYPE "${enumName}"`);
     }
+  });
+});
+
+describe("meal planning schema (issue #167)", () => {
+  const MEAL_MIGRATION_DIR = path.join(
+    MIGRATIONS_DIR,
+    "0005_add_meal_planning"
+  );
+
+  it("declares the MealType and DayOfWeek enums", () => {
+    const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
+
+    expect(schemaContent).toMatch(/enum\s+MealType\s*\{[\s\S]*?breakfast/);
+    expect(schemaContent).toMatch(/enum\s+MealType\s*\{[\s\S]*?lunch/);
+    expect(schemaContent).toMatch(/enum\s+MealType\s*\{[\s\S]*?dinner/);
+    expect(schemaContent).toMatch(/enum\s+MealType\s*\{[\s\S]*?snack/);
+
+    for (const day of [
+      "monday",
+      "tuesday",
+      "wednesday",
+      "thursday",
+      "friday",
+      "saturday",
+      "sunday",
+    ]) {
+      expect(schemaContent).toMatch(
+        new RegExp(`enum\\s+DayOfWeek\\s*\\{[\\s\\S]*?${day}`)
+      );
+    }
+  });
+
+  it("declares the meal-planning models", () => {
+    const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
+
+    for (const model of [
+      "Meal",
+      "PlannedMeal",
+      "MealIngredient",
+      "GroceryList",
+      "GroceryListItem",
+      "SavedMeal",
+      "MealTemplate",
+    ]) {
+      expect(
+        schemaContent,
+        `schema.prisma is missing model "${model}"`
+      ).toMatch(new RegExp(`^model\\s+${model}\\s*\\{`, "m"));
+    }
+  });
+
+  it("Meal model indexes (userId, type) and references the recipe id as a plain string", () => {
+    const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
+    const mealBlockMatch = schemaContent.match(/model Meal \{[\s\S]*?\n\}/);
+    expect(mealBlockMatch, "Meal model not found").not.toBeNull();
+    const mealBlock = mealBlockMatch![0];
+
+    expect(mealBlock).toMatch(/@@index\(\[userId,\s*type\]\)/);
+    expect(mealBlock).toMatch(/recipeId\s+String\?/);
+    expect(mealBlock).toMatch(/profileId\s+String\?/);
+    expect(mealBlock).toMatch(/servings\s+Int/);
+    expect(mealBlock).toMatch(/notes\s+String\?/);
+  });
+
+  it("PlannedMeal enforces the (userId, weekStart, dayOfWeek, mealType) unique constraint", () => {
+    const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
+    const blockMatch = schemaContent.match(/model PlannedMeal \{[\s\S]*?\n\}/);
+    expect(blockMatch, "PlannedMeal model not found").not.toBeNull();
+    const block = blockMatch![0];
+
+    expect(block).toMatch(
+      /@@unique\(\[userId,\s*weekStart,\s*dayOfWeek,\s*mealType\]\)/
+    );
+  });
+
+  it("GroceryList belongs to a User and has items keyed by listId", () => {
+    const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
+    const groceryListBlock = schemaContent.match(
+      /model GroceryList \{[\s\S]*?\n\}/
+    );
+    const groceryItemBlock = schemaContent.match(
+      /model GroceryListItem \{[\s\S]*?\n\}/
+    );
+
+    expect(groceryListBlock, "GroceryList model not found").not.toBeNull();
+    expect(groceryItemBlock, "GroceryListItem model not found").not.toBeNull();
+
+    expect(groceryListBlock![0]).toMatch(/userId\s+String/);
+    expect(groceryListBlock![0]).toMatch(/weekStart\s+DateTime/);
+    expect(groceryItemBlock![0]).toMatch(/listId\s+String/);
+    expect(groceryItemBlock![0]).toMatch(/isChecked\s+Boolean/);
+    expect(groceryItemBlock![0]).toMatch(/mealId\s+String\?/);
+  });
+
+  it("creates a 0005_add_meal_planning migration with the new tables", () => {
+    expect(
+      fs.existsSync(MEAL_MIGRATION_DIR),
+      "0005_add_meal_planning migration directory missing"
+    ).toBe(true);
+
+    const sqlPath = path.join(MEAL_MIGRATION_DIR, "migration.sql");
+    expect(fs.existsSync(sqlPath)).toBe(true);
+
+    const content = fs.readFileSync(sqlPath, "utf-8");
+    for (const table of [
+      "Meal",
+      "PlannedMeal",
+      "MealIngredient",
+      "GroceryList",
+      "GroceryListItem",
+      "SavedMeal",
+      "MealTemplate",
+    ]) {
+      expect(
+        content,
+        `0005 migration is missing CREATE TABLE for "${table}"`
+      ).toContain(`CREATE TABLE "${table}"`);
+    }
+
+    expect(content).toContain('CREATE TYPE "MealType"');
+    expect(content).toContain('CREATE TYPE "DayOfWeek"');
+
+    // PlannedMeal unique constraint expressed as a unique index
+    expect(content).toMatch(
+      /CREATE UNIQUE INDEX\s+"[^"]+"\s+ON\s+"PlannedMeal"\s*\(\s*"userId"\s*,\s*"weekStart"\s*,\s*"dayOfWeek"\s*,\s*"mealType"\s*\)/
+    );
+
+    // Meal index for fast lookup
+    expect(content).toMatch(
+      /CREATE INDEX\s+"[^"]+"\s+ON\s+"Meal"\s*\(\s*"userId"\s*,\s*"type"\s*\)/
+    );
   });
 });
--- a/prisma/__tests__/migration-setup.test.ts
+++ b/prisma/__tests__/migration-setup.test.ts
@@ -185,15 +185,18 @@ describe("Account uniqueness constraints (issue #61)", () => {
 
 function readAllMigrationsCombined(): string {
   const entries = fs.readdirSync(MIGRATIONS_DIR, { withFileTypes: true });
-  return entries
-    .filter((e) => e.isDirectory() && e.name !== "__tests__")
-    .map((e) =>
-      fs.readFileSync(
-        path.join(MIGRATIONS_DIR, e.name, "migration.sql"),
-        "utf-8"
-      )
-    )
-    .join("\n");
+  return (
+    entries
+      .filter((e) => e.isDirectory() && e.name !== "__tests__")
+      .map((e) => path.join(MIGRATIONS_DIR, e.name, "migration.sql"))
+      // Skip directories without a migration.sql so a partial / accidental
+      // directory yields a clear assertion failure (via a separate test
+      // that checks every migration dir contains a migration.sql) rather
+      // than an unrelated ENOENT here.
+      .filter((p) => fs.existsSync(p))
+      .map((p) => fs.readFileSync(p, "utf-8"))
+      .join("\n")
+  );
 }
 
 describe("schema and migration consistency", () => {
@@ -326,9 +329,21 @@ describe("meal planning schema (issue #167)", () => {
 
     expect(groceryListBlock![0]).toMatch(/userId\s+String/);
     expect(groceryListBlock![0]).toMatch(/weekStart\s+DateTime/);
+    // One list per user per week — guards getGroceryListForWeek's
+    // findFirst contract against concurrent inserts.
+    expect(groceryListBlock![0]).toMatch(/@@unique\(\[userId,\s*weekStart\]\)/);
     expect(groceryItemBlock![0]).toMatch(/listId\s+String/);
     expect(groceryItemBlock![0]).toMatch(/isChecked\s+Boolean/);
     expect(groceryItemBlock![0]).toMatch(/mealId\s+String\?/);
+  });
+
+  it("SavedMeal.lastUsed is nullable so a never-used meal is distinguishable", () => {
+    const schemaContent = fs.readFileSync(SCHEMA_PATH, "utf-8");
+    const block = schemaContent.match(/model SavedMeal \{[\s\S]*?\n\}/);
+    expect(block, "SavedMeal model not found").not.toBeNull();
+
+    expect(block![0]).toMatch(/lastUsed\s+DateTime\?/);
+    expect(block![0]).not.toMatch(/lastUsed\s+DateTime\s+@default\(now\(\)\)/);
   });
 
   it("creates a 0005_add_meal_planning migration with the new tables", () => {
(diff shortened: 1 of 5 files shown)


prisma/schema.prisma
```diff
@@ -207,3 +213,132 @@ model TaskAssignment {
...
+  mealId    String
+  weekStart DateTime // Monday of the week (UTC midnight)
+  dayOfWeek DayOfWeek
+  mealType  MealType
+  createdAt DateTime  @default(now())
+
+  user User @relation(fields: [userId], references: [id], onDelete: Cascade)
+  meal Meal @relation(fields: [mealId], references: [id], onDelete: Cascade)
+
+  @@unique([userId, weekStart, dayOfWeek, mealType])
```
BenSeymourODB, line 269: **[Schema — unique constraint vs. index redundancy]**

`@@unique([userId, weekStart, dayOfWeek, mealType])` already implies a B-tree index on that 4-column prefix. The separate `@@index([userId, weekStart])` at line 268 is therefore redundant — PostgreSQL (and every major RDBMS) can satisfy an equality/range query on `(userId, weekStart)` by scanning the leading columns of the unique index. Keeping the redundant index wastes write overhead and storage.

**Recommendation:** remove `@@index([userId, weekStart])` from `PlannedMeal`. The same removal is needed in the migration SQL (the `PlannedMeal_userId_weekStart_idx` `CREATE INDEX` statement).

---

prisma/schema.prisma
```diff
@@ -207,3 +213,132 @@ model TaskAssignment {
...
+  mealId    String? // Plain id; the source meal may have been deleted
+  order     Int      @default(0)
+  createdAt DateTime @default(now())
+
+  list GroceryList @relation(fields: [listId], references: [id], onDelete: Cascade)
+
+  @@index([listId, category])
+}
+
+model SavedMeal {
```
BenSeymourODB, line 317: **[Schema — SavedMeal.lastUsed default is misleading]**

`lastUsed DateTime @default(now())` sets the column to the insertion timestamp. That means a meal that has never actually been used will appear to have been used "just now", which will corrupt the `@@index([userId, lastUsed])` sort order and any "recently used" queries.

**Recommendation:** make `lastUsed` nullable (`DateTime?`) with no default, and set it explicitly when a `SavedMeal` is actually placed on the plan (the API route for issue #168 is the right place). Alternatively keep it non-nullable but document clearly that it is initialized to `createdAt` as a proxy for "never used", and filter on `useCount = 0` for the true "never used" case. Either way the current default silently loses information.

---

prisma/schema.prisma
```diff
@@ -207,3 +213,132 @@ model TaskAssignment {
...
+  quantity String
+  category String @default("other")
+  order    Int    @default(0)
+
+  meal Meal @relation(fields: [mealId], references: [id], onDelete: Cascade)
+
+  @@index([mealId])
+}
+
+model GroceryList {
```
BenSeymourODB, line 285: **[Schema — GroceryList missing unique constraint on (userId, weekStart)]**

The plan doc and `getGroceryListForWeek` both assume one list per user per week, but there is no database-level guarantee. Without a unique constraint, two concurrent `createGroceryList` calls for the same week will silently insert two rows. `findFirst` will then return whichever row was inserted first, leaving the second orphaned and confusing.

**Recommendation:** add `@@unique([userId, weekStart])` to `GroceryList` (and a corresponding `CREATE UNIQUE INDEX` in the migration). The existing `@@index([userId, weekStart])` can be dropped in favour of the implied unique index for the same reason noted on `PlannedMeal` above.

---

prisma/schema.prisma
```diff
@@ -207,3 +213,132 @@ model TaskAssignment {
...
+model MealTemplate {
+  id          String   @id @default(cuid())
+  name        String
+  description String   @db.Text
+  meals       Json // TemplateMeal[] - 28 entries (7 days x 4 meal types)
+  isPublic    Boolean  @default(false)
+  createdBy   String?
+  createdAt   DateTime @default(now())
+
+  user User? @relation(fields: [createdBy], references: [id], onDelete: SetNull)
```
BenSeymourODB, line 347: **[Schema — MealTemplate relation name clashes with User.mealTemplates]**

The `MealTemplate` model names its back-relation field `user` (line 341), which is fine. However `User` also carries a `mealTemplates MealTemplate[]` relation. Prisma requires that when a model has more than one relation to the same target the relations use explicit `@relation(name: "...")` annotations. Here there is only one relation between `User` and `MealTemplate` so this is technically fine, but the field is named `user` on `MealTemplate` — a model that also has a `createdBy` column with a nullable FK. The implicit relation name Prisma assigns will be `UserToMealTemplate`.

More importantly: the plan doc schema (`.claude/plans/meal-planning.md`, line 505) had `onDelete: SetNull` omitted from this relation, but this PR correctly adds it (`ON DELETE SET NULL` in migration line 156 / schema line 341). That is the right choice — no issue here, just confirming the plan divergence is intentional and correct.

---

prisma/migrations/0005_add_meal_planning/migration.sql
```diff
@@ -0,0 +1,156 @@
...
+CREATE TABLE "SavedMeal" (
+    "id" TEXT NOT NULL,
+    "userId" TEXT NOT NULL,
+    "name" TEXT NOT NULL,
+    "type" "MealType" NOT NULL,
+    "servings" INTEGER NOT NULL DEFAULT 4,
+    "recipeId" TEXT,
+    "isFavorite" BOOLEAN NOT NULL DEFAULT false,
+    "useCount" INTEGER NOT NULL DEFAULT 0,
+    "lastUsed" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP,
```
BenSeymourODB: **[Migration — SavedMeal.lastUsed default matches schema concern]**

This mirrors the schema concern above. `"lastUsed" TIMESTAMP(3) NOT NULL DEFAULT CURRENT_TIMESTAMP` means every newly-created `SavedMeal` row immediately looks like it was used at creation time. If a future query sorts by `lastUsed DESC` to show "recently used" meals, brand-new unseen meals will appear at the top alongside genuinely-recently-used ones.

See the companion schema comment. The fix (nullable column, drop default) needs to land in both this SQL file and `schema.prisma`.

---

prisma/migrations/0005_add_meal_planning/migration.sql
```diff
@@ -0,0 +1,156 @@
+-- CreateEnum
+CREATE TYPE "MealType" AS ENUM ('breakfast', 'lunch', 'dinner', 'snack');
+
+-- CreateEnum
+CREATE TYPE "DayOfWeek" AS ENUM ('monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday');
+
+-- CreateTable
+CREATE TABLE "Meal" (
+    "id" TEXT NOT NULL,
+    "userId" TEXT NOT NULL,
```
BenSeymourODB, line 10: **[Migration — enum values are unquoted strings, verify intent]**

```sql
CREATE TYPE "MealType" AS ENUM ('breakfast', 'lunch', 'dinner', 'snack');
```

The values use single-quoted, lower-case strings, which matches the Prisma schema enum members exactly. This is correct Prisma convention (Prisma maps enum members to their literal names unless `@map` is used). No action needed — just confirming this is deliberate and consistent throughout both `MealType` and `DayOfWeek`.

---

src/lib/services/meal-planning.ts
```diff
@@ -0,0 +1,118 @@
...
+import { prisma } from "@/lib/db";
+
+const MS_PER_DAY = 24 * 60 * 60 * 1000;
+
+/**
+ * Resolve any date in a calendar week to the Monday at UTC midnight.
+ * `PlannedMeal.@@unique([userId, weekStart, dayOfWeek, mealType])` is
+ * keyed on `weekStart`, so all callers must agree on the same anchor.
+ */
+function getMondayOfWeek(date: Date): Date {
```
BenSeymourODB, line 27: **[Service — getMondayOfWeek is correct but worth a DST callout]**

The implementation correctly works entirely in UTC (`getUTCFullYear`, `getUTCMonth`, `getUTCDate`, `Date.UTC`, `getUTCDay`) and so is immune to DST transitions and local-timezone offsets on the server. The arithmetic `utc - offset * MS_PER_DAY` is safe because UTC never has DST gaps or folds.

One edge case that is handled correctly: a Sunday input (`getUTCDay() === 0`) maps `offset = 6`, rolling back to the preceding Monday. The `it.each` table covers this with the `2026-05-03T23:59:59.000Z` (Sunday) case.

No change needed — this is a positive callout for reviewers who might question the manual arithmetic vs. a date library.

---

src/lib/services/meal-planning.ts
```diff
@@ -0,0 +1,118 @@
...
+  weekStart: Date
+): Promise<PlannedMealWithMeal[]> {
+  return prisma.plannedMeal.findMany({
+    where: {
+      userId,
+      weekStart: getMondayOfWeek(weekStart),
+    },
+    include: { meal: true },
+    orderBy: [{ dayOfWeek: "asc" }, { mealType: "asc" }],
+  }) as Promise<PlannedMealWithMeal[]>;
```
BenSeymourODB: **[Service — unsafe type cast on findMany / findFirst return]**

```ts
}) as Promise<PlannedMealWithMeal[]>;
```

And similarly at line 117:
```ts
}) as Promise<GroceryListWithItems | null>;
```

Prisma's `findMany` with `include: { meal: true }` already returns `(PlannedMeal & { meal: Meal })[]` at the type level — no cast is needed and the cast hides any future type mismatch. The same applies to `findFirst` with `include: { items: ... }`.

**Recommendation:** remove both `as Promise<…>` casts and let TypeScript infer the return type from the Prisma call. If the inferred type doesn't match the declared return type, that is a compile error worth seeing rather than suppressing.

---

src/lib/services/__tests__/meal-planning.test.ts
```diff
@@ -0,0 +1,264 @@
...
+      findMany: vi.fn(),
+    },
+    groceryList: {
+      create: vi.fn(),
+      findFirst: vi.fn(),
+    },
+  },
+}));
+
+const mockPrisma = prisma as unknown as {
```
BenSeymourODB, line 34: **[Tests — mockPrisma cast uses `unknown` intermediary, acceptable but fragile]**

```ts
const mockPrisma = prisma as unknown as { … };
```

This double-cast works and follows the established pattern in the codebase. One improvement: use `vi.mocked()` helpers or a typed mock factory so that `mockResolvedValue` calls get type-checked against the actual Prisma return type. For example:

```ts
import { type Mock } from 'vitest';
// then in the mock shape:
create: vi.fn() as Mock<Parameters<typeof prisma.plannedMeal.create>, ReturnType<typeof prisma.plannedMeal.create>>,
```

This is a minor suggestion — the current approach is functionally equivalent and matches the existing test style in the repo. Not a blocker.

---

src/lib/services/__tests__/meal-planning.test.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    expect(result).toBeNull();
+  });
+});
+
+describe("week-start normalization", () => {
+  // Sanity-check the Monday-resolution used by every create/read path.
+  // We can't import getMondayOfWeek directly without exposing it; the
+  // public surface exercises it via the createGroceryList path.
+  it.each([
```
BenSeymourODB, line 246: **[Tests — week-start normalization table is missing Saturday]**

The `it.each` table covers Monday, Tuesday, Thursday, and Sunday. Saturday (`getUTCDay() === 6`, offset = 5) is not exercised. While the arithmetic is obviously correct, Saturday is the only weekday that sits exactly one day before Sunday in the offset table (`offset = day - 1 = 5`), and a Saturday input is the most natural one to accidentally off-by-one. Adding it costs one line:

```ts
["2026-05-02T08:00:00.000Z", "2026-04-27T00:00:00.000Z"], // Saturday
```

(2026-05-02 is a Saturday in the same week as the other test dates — today's date per the system context, which is a nice coincidence.)

---

prisma/__tests__/migration-setup.test.ts
```diff
@@ -183,13 +183,23 @@ describe("Account uniqueness constraints (issue #61)", () => {
   });
 });
```
BenSeymourODB, line 185: **[Migration tests — readAllMigrationsCombined could silently skip migrations with missing SQL]**

```ts
.map((e) =>
  fs.readFileSync(
    path.join(MIGRATIONS_DIR, e.name, "migration.sql"),
    "utf-8"
  )
)
```

If a migration directory exists but contains no `migration.sql` file (e.g. a partial hand-authored migration or an accidental directory), `readFileSync` will throw and abort the whole test suite with an unrelated error rather than a clear assertion failure. Consider adding a guard:

```ts
.filter((e) => fs.existsSync(path.join(MIGRATIONS_DIR, e.name, "migration.sql")))
```

or wrapping in a try/catch that re-throws with a clear message. Minor robustness issue, not a blocker.

---

BenSeymourODB (commented):
## Code Review — PR #193 (feat(meals): Prisma schema + migration + service for meal planning)

Overall this is a solid, well-structured slice PR. TDD was followed correctly (tests written, 12 new service tests + 6 new schema tests all green), the migration SQL matches the schema exactly, cascade choices are sound throughout, the `getMondayOfWeek` helper is DST-safe, and no manual memoization was introduced. The existing migration-consistency tests were strengthened rather than weakened. Good work.

There are **two issues that need fixing before merge** and several minor suggestions:

---

### Must-fix

**1. `GroceryList` is missing a unique constraint on `(userId, weekStart)`** (schema.prisma ~line 284)

The service's `getGroceryListForWeek` uses `findFirst`, implicitly assuming one list per user per week. Without a `@@unique([userId, weekStart])` (and corresponding `CREATE UNIQUE INDEX` in the migration), two concurrent `createGroceryList` calls for the same week silently create two rows. The second is then permanently orphaned. This is a data-integrity bug that will be invisible until a concurrency issue surfaces in production.

Fix: add `@@unique([userId, weekStart])` to `GroceryList` and the matching SQL to the migration. The existing `@@index([userId, weekStart])` can be dropped since the unique index serves the same lookup purpose.

**2. `SavedMeal.lastUsed` default is semantically wrong** (schema.prisma ~line 324 / migration.sql ~line 85)

`lastUsed DateTime @default(now())` initialises every newly-created `SavedMeal` to the insertion timestamp. A meal that has never been used will therefore sort at the top of any "recently used" query, defeating the purpose of the field. Fix: make it `DateTime?` (nullable, no default) and set it only when the meal is actually placed on a plan — or at minimum document clearly that `useCount = 0` is the authoritative "never used" sentinel and sort accordingly.

---

### Minor / non-blocking

- **`PlannedMeal @@index([userId, weekStart])` is redundant** — the leading columns of `@@unique([userId, weekStart, dayOfWeek, mealType])` already cover this lookup. Drop the index to avoid unnecessary write overhead. (See inline comment on schema.prisma.)

- **Unsafe `as Promise<…>` casts in the service** — Prisma's `include` already makes the return types exact. Both casts can be removed and the inferred types relied upon. (See inline comment on meal-planning.ts.)

- **`week-start normalization` test table is missing Saturday** — trivial to add, closes the one untested offset. (See inline comment on meal-planning.test.ts.)

- **`readAllMigrationsCombined` can throw uncleanly** if a migration directory has no `migration.sql`. A one-line existence guard makes the failure message meaningful. (See inline comment on migration-setup.test.ts.)

- **Typed mock cast** — `prisma as unknown as {…}` works and matches the repo pattern; a `vi.mocked()`-style cast would give compile-time checking on `mockResolvedValue` args. Non-blocking suggestion. (See inline comment on meal-planning.test.ts.)

---

prisma/schema.prisma
BenSeymourODB: Fixed in `bf054c9`. Dropped `@@index([userId, weekStart])` from `PlannedMeal` and removed the matching `CREATE INDEX "PlannedMeal_userId_weekStart_idx"` from the migration SQL. Left an inline comment on the `@@unique` line noting the prefix-coverage rationale so future maintainers don't re-add the index.

---

prisma/schema.prisma
BenSeymourODB: Fixed in `bf054c9`. `SavedMeal.lastUsed` is now `DateTime?` with no default, and the migration drops the `NOT NULL DEFAULT CURRENT_TIMESTAMP`. The API route in #168 will set it explicitly when a saved meal is placed on a plan. New schema test asserts `lastUsed DateTime?` and rejects the old `@default(now())` form, plus a migration test that rejects `NOT NULL DEFAULT CURRENT_TIMESTAMP` on the column.

---

prisma/schema.prisma
BenSeymourODB: Fixed in `bf054c9`. Added `@@unique([userId, weekStart])` to `GroceryList` and replaced the `CREATE INDEX "GroceryList_userId_weekStart_idx"` with `CREATE UNIQUE INDEX "GroceryList_userId_weekStart_key"` in the migration. The schema comment notes that the unique index also serves the (userId, weekStart) lookup so no separate `@@index` is needed. Two new tests assert the constraint exists at both layers.

---

prisma/schema.prisma
BenSeymourODB: Acknowledged — no change needed. `ON DELETE SET NULL` on `MealTemplate.createdBy` was a deliberate divergence from the plan: a deleted user should not orphan public templates. Confirming the back-relation field name `user` is fine since there's only one `User` ↔ `MealTemplate` relation and Prisma's implicit `UserToMealTemplate` name doesn't conflict with anything.

---

prisma/migrations/0005_add_meal_planning/migration.sql
BenSeymourODB: Fixed in `bf054c9` alongside the schema change. Migration line is now `"lastUsed" TIMESTAMP(3),` with no `NOT NULL` and no default.

---

prisma/migrations/0005_add_meal_planning/migration.sql
BenSeymourODB: Confirmed deliberate. The lower-case literal values match the Prisma enum members exactly so no `@map` is needed for either `MealType` or `DayOfWeek`. Sticking with the plan's literal values keeps the API surface predictable for downstream API/UI tickets.

---

src/lib/services/meal-planning.ts
BenSeymourODB: Acknowledged — no change. The Friday/Saturday/Wednesday cases were also added to the `it.each` table in `bf054c9` so every weekday now exercises the offset arithmetic.

---

src/lib/services/meal-planning.ts
BenSeymourODB: Fixed in `bf054c9`. Both `as Promise<…>` casts removed. Both functions are now non-async with implicit return types — Prisma's `findMany`/`findFirst` with `include` already produce the exact types the public `PlannedMealWithMeal` / `GroceryListWithItems` aliases describe. `pnpm check-types` clean.

---

src/lib/services/__tests__/meal-planning.test.ts
BenSeymourODB: Skipping for now to keep this PR's mock pattern aligned with `reward-points.test.ts` and `streak.test.ts` (the existing repo style). Worth a separate cross-cutting PR that migrates all three service tests to `vi.mocked<…>()` at once so the mocks stay symmetric — happy to follow up if you'd like me to file an issue.

---

src/lib/services/__tests__/meal-planning.test.ts
BenSeymourODB: Fixed in `bf054c9`. The table now exercises every weekday — Mon/Tue/Wed/Thu/Fri/Sat/Sun — with Saturday using your suggested `2026-05-02T08:00:00.000Z` input. Added inline comments on the Sat/Sun rows pointing at the two off-by-one-prone branches (`offset = day - 1 = 5` and the `day === 0 ? 6` Sunday special case).

---

prisma/__tests__/migration-setup.test.ts
BenSeymourODB: Fixed in `bf054c9`. `readAllMigrationsCombined` now filters with `fs.existsSync(...)` before reading. The "every migration directory contains a migration.sql" assertion (`should have all migration directories containing a migration.sql file`) still catches a missing SQL file, but with a clear message rather than an unrelated ENOENT bubbling up first.

---

PR #190, opened 2026-05-02
feat(tasks): add typed Google Tasks API wrapper (#162)

BenSeymourODB:
## Summary

Closes the remaining work for #162 — the OAuth scope was already added in `src/lib/auth/auth.ts:23`, but the typed helper module the rest of the Tasks track needs to bind against did not exist. This PR adds it and refactors the four existing `/api/tasks/*` routes onto it, removing duplicated direct-fetch + error-handling boilerplate.

## What's new

- `src/lib/google/tasks-types.ts` — canonical TypeScript shapes for Google Tasks v1 (`GoogleTask`, `GoogleTaskList`, `ListTasksOptions`, `CreateTaskInput`, `PatchTaskInput`) plus `GoogleTasksApiError` and the `TASK_STATUSES` const.
- `src/lib/google/tasks-api.ts` — typed thin wrapper exposing the helpers the issue calls out: `listTaskLists`, `listTasks`, `createTask`, `patchTask`, `deleteTask`. Owns URL/query construction, bearer-token attachment, `fetchWithRetry` integration, response parsing, and error envelope.
- 15 new unit tests in `src/lib/google/__tests__/tasks-api.test.ts` covering URL/query construction (encoding, defaults, full option set), response parsing (success + missing `items`), and the `GoogleTasksApiError` shape (parsed message vs. fallback).

## Refactors

- `/api/tasks` (GET + POST), `/api/tasks/lists` (GET), `/api/tasks/[taskId]` (PATCH), and `/api/tasks/[taskId]/complete` (POST) now delegate the HTTP call to the wrapper and translate `GoogleTasksApiError` into the existing 401/403/4xx response shapes.
- 403 from upstream now surfaces as `requiresReauth: true` so clients can prompt re-consent when the user authenticated before the Tasks scope was added — that satisfies the "existing users are prompted to re-consent on first Tasks API use" line of the acceptance criteria.

## Why this is the next issue

`#162` is the unblocked root of the Tasks track (Days 7-8). It unblocks 4 downstream Tasks issues (#163 API routes, #164 TaskList components, #165 NewTaskModal, #166 multi-profile wiring) and indirectly the Rewards UI (#173) and multi-profile Phase 4. Highest fan-out among the three unblocked roots (#162, #167, #171).

## Test plan

- [x] `pnpm test:run` — 1435 tests pass across 98 files
- [x] `pnpm lint:fix` — 0 errors (4 pre-existing warnings unrelated to this PR)
- [x] `pnpm format:fix` — clean
[...]

Issue #166, opened 2026-04-27
feat(tasks): wire Google Tasks to multi-profile assignments (unblocks Phase 4)


BDSeymour:
Connect Google Tasks completions to the existing multi-profile system so per-profile task views and streak tracking can light up. This unblocks Phase 4 of the multi-profile plan (currently noted as blocked on Google Tasks integration) and Step 24 (streak tracking).

## Scope

- Extend the per-profile task filter so a TaskList instance can show tasks for one profile or All.
- Add a profile assignment surface: when a task has a profile ID stored alongside its Google Tasks ID (in our DB, since Google Tasks has no profile concept), display it.
- Persist the profile↔task mapping in a thin local table (`TaskProfileAssignment(googleTaskId, profileId, listId, userId)`) — Google Tasks remains the source of truth for the task itself.
- On completion, write a `TaskCompletion` row keyed by profile so streak tracking can consume it.

## Acceptance

- A profile-scoped TaskList shows only that profile's tasks and the unassigned ones (per setting).
- Completing a task records the completion against the active profile.
- Existing multi-profile component tests still pass.

## Blocked by

- TaskList components issue (Tasks track).
- API routes issue (Tasks track).

## Plan reference

`.claude/plans/multi-profile-family-support.md` (Phase 4 + Step 24), `.claude/plans/google-tasks-todo-list.md`.

## Track

Day 8 (Tasks + multi-profile follow-up).

BenSeymourODB:
Claiming this for an automated session on branch `claude/loving-faraday-Xcyz9`.

## Triage notes (2026-05-02)

After surveying the codebase, much of the foundation for this issue is already in place:

- `TaskAssignment(id, taskId, profileId, createdAt)` model exists in `prisma/schema.prisma` with `@@unique([taskId, profileId])` and `@@index([profileId, taskId])`.
- `task-assignment-picker.tsx` lets a user pick assigned profiles for a task, and `PUT/GET/DELETE /api/tasks/[taskId]/assignments` are wired and tested.
- `POST /api/tasks/[taskId]/complete` already updates `ProfileRewardPoints` (current/longest streak + lastActivityDate) via `updateProfileStreak(profileId)`. That covers the streak tracking acceptance criterion β€" a separate `TaskCompletion` table is redundant given the data already lives on `ProfileRewardPoints`/`PointTransaction`.

What is **not** yet done, and what this PR will deliver:

[...]

PR #194, opened 2026-05-02
feat(tasks): add NewTaskModal + useCreateTask (#165)

### LINKED ISSUE #165: feat(tasks): add NewTaskModal and useCreateTask hook
Add the modal launched from the "+ Add Task" footer in TaskList, supporting smart defaults and the four required fields.

## Scope

- `src/components/tasks/new-task-modal.tsx` — title (required), list (required, smart default if a single list is shown), due date, notes.
- `src/components/tasks/use-create-task.ts` — POST to `/api/tasks/create`, optimistic insert into TaskList state via callback, rollback on failure.
- Wire footer button in `TaskList` to open the modal and pass `availableLists` + `defaultListId`.
- Accessibility: focus trap, ESC to close, focus first input on open, ARIA dialog role.

## Acceptance

- Validation: empty title and missing list show inline errors.
- Successful create closes the modal, refreshes the list, and emits a `TaskCreated` log event.
- Component tests cover validation + happy path; Playwright covers the open → fill → submit flow.

## Blocked by

- API routes issue (Tasks track).
- TaskList component issue (Tasks track).

## Plan reference

`.claude/plans/new-task-modal.md`.

## Track

Days 7–8 (Tasks track).

BenSeymourODB:
## Summary

Closes #165 — adds the "+ Add Task" entry point for `TaskList`: a `useCreateTask` hook + `NewTaskModal` dialog wired through `TaskList`'s footer, plus a Playwright E2E that exercises the full open → fill → submit flow against an intercepted `/api/tasks` endpoint.

## What's new

- **`src/components/tasks/use-create-task.ts`** — POSTs to the existing `/api/tasks` route (added by PR #190) with `{ listId, title, notes?, due? }`, surfaces the server's `error` message verbatim (so `requiresReauth: true` 401s bubble through).
- **`src/components/tasks/new-task-modal.tsx`** — Radix Dialog modal. Title (required) + list (required) + due (optional) + notes (optional). Inline alerts for missing title/list, server error banner inside the dialog, form resets on each open. Smart default for the list dropdown via `defaultListId`. Native `<select>` for the list (matches the plan; simpler than Radix Select for this use case).
- **`src/components/tasks/task-list.tsx`** — adds the "+ Add Task" footer button (disabled when no enabled lists) and pre-selects the only enabled list as the default. Successful create triggers `refreshTasks()`.
- **`src/app/test/tasks/page.tsx`** — minimal `/test/tasks` E2E harness mirroring `/test/calendar`. `?lists=multi` exercises the multi-list branch.
- **`e2e/new-task-modal.spec.ts`** — 7 E2E specs: footer opens dialog, smart default, multi-list no-default, title-required validation, happy path (post captured + new task rendered), 500 error keeps dialog open, Cancel sends nothing.

## Tests

- [x] `pnpm test:run` — 1469 tests pass across 100 files (added 33 new tests: 10 hook + 16 modal + 7 TaskList footer behaviors)
- [x] `pnpm lint` — 0 errors (4 pre-existing warnings unrelated to this PR)
- [x] `pnpm test:e2e e2e/new-task-modal.spec.ts --project=chromium` — 7 passed

## Acceptance criteria (from #165)

- [x] Validation: empty title and missing list show inline errors.
- [x] Successful create closes the modal and refreshes the list.
- [x] Component tests cover validation + happy path.
- [x] Playwright covers the open → fill → submit flow.

## Notes

- The plan's `TaskCreated` log event is already emitted server-side (in `POST /api/tasks`), so the modal does not duplicate it on the client.
- Optimistic insert is intentionally **not** implemented in v1 — the modal waits for the API to return the created task and then triggers a `refreshTasks()`. The plan calls this out as the simpler v1 approach.
- The hook surfaces `requiresReauth: true` 401 responses through the existing error message, but doesn't yet wire a re-auth UI — that's downstream of #166 and outside the scope of this slice.
- A pre-existing CI step ("Validate Prisma migrations are up to date") fails locally too because the workflow doesn't provision a shadow database — this PR doesn't touch Prisma. Filed mentally for follow-up; not addressed here.

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/components/tasks/__tests__/task-list.test.tsx
+++ b/src/components/tasks/__tests__/task-list.test.tsx
@@ -1,9 +1,10 @@
 /**
  * Tests for TaskList component
  */
-import { render, screen } from "@testing-library/react";
+import { act, render, screen } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { NewTaskModal } from "../new-task-modal";
 import { TaskList } from "../task-list";
 import type { TaskListConfig } from "../types";
 // Import after mocking
@@ -14,7 +15,14 @@ vi.mock("../use-tasks", () => ({
   useTasks: vi.fn(),
 }));
 
+// Stub NewTaskModal so we can assert on the props TaskList passes in
+// without re-testing the modal's internals here.
+vi.mock("../new-task-modal", () => ({
+  NewTaskModal: vi.fn(() => null),
+}));
+
 const mockUseTasks = vi.mocked(useTasks);
+const mockNewTaskModal = vi.mocked(NewTaskModal);
 
 describe("TaskList", () => {
   const mockConfig: TaskListConfig = {
@@ -239,6 +247,123 @@ describe("TaskList", () => {
 
       const heading = screen.getByRole("heading", { name: "My Tasks" });
       expect(heading).toBeInTheDocument();
+    });
+  });
+
+  describe("add task footer", () => {
+    function lastModalProps() {
+      const calls = mockNewTaskModal.mock.calls;
+      return calls[calls.length - 1]?.[0];
+    }
+
+    it("renders an Add Task footer button", () => {
+      render(<TaskList config={mockConfig} />);
+      expect(
+        screen.getByRole("button", { name: /add task/i })
+      ).toBeInTheDocument();
+    });
+
+    it("renders the modal closed by default", () => {
+      render(<TaskList config={mockConfig} />);
+      expect(lastModalProps()?.open).toBe(false);
+    });
+
+    it("opens the modal when the Add Task button is clicked", async () => {
+      const user = userEvent.setup();
+      render(<TaskList config={mockConfig} />);
+
+      await user.click(screen.getByRole("button", { name: /add task/i }));
+
+      expect(lastModalProps()?.open).toBe(true);
+    });
+
+    it("passes only enabled lists as availableLists", () => {
+      const config: TaskListConfig = {
+        ...mockConfig,
+        lists: [
+          {
+            listId: "list-1",
+            listTitle: "Work",
+            color: "#3b82f6",
+            enabled: true,
+          },
+          {
+            listId: "list-2",
+            listTitle: "Disabled",
+            color: "#999",
+            enabled: false,
+          },
+          {
+            listId: "list-3",
+            listTitle: "Personal",
+            color: "#ef4444",
+            enabled: true,
+          },
+        ],
+      };
+      render(<TaskList config={config} />);
+
+      const props = lastModalProps();
+      expect(props?.availableLists.map((l) => l.listId)).toEqual([
+        "list-1",
+        "list-3",
+      ]);
+    });
+
+    it("pre-selects the only enabled list as the default", () => {
+      render(<TaskList config={mockConfig} />);
+      expect(lastModalProps()?.defaultListId).toBe("list-1");
+    });
+
+    it("leaves defaultListId undefined when multiple lists are enabled", () => {
+      const config: TaskListConfig = {
+        ...mockConfig,
+        lists: [
+          {
+            listId: "list-1",
+            listTitle: "Work",
+            color: "#3b82f6",
+            enabled: true,
+          },
+          {
+            listId: "list-2",
+            listTitle: "Personal",
+            color: "#ef4444",
+            enabled: true,
+          },
+        ],
+      };
+      render(<TaskList config={config} />);
+      expect(lastModalProps()?.defaultListId).toBeUndefined();
+    });
+
+    it("refreshes tasks when the modal reports a successful create", () => {
+      render(<TaskList config={mockConfig} />);
+
+      const props = lastModalProps();
+      props?.onSuccess?.({
+        id: "new-1",
+        title: "Buy milk",
+        status: "needsAction",
+        updated: "2024-06-14T00:00:00Z",
+        position: "0",
+      });
+
+      expect(mockRefreshTasks).toHaveBeenCalled();
+    });
+
+    it("closes the modal when onOpenChange(false) fires", async () => {
+      const user = userEvent.setup();
+      render(<TaskList config={mockConfig} />);
+
+      await user.click(screen.getByRole("button", { name: /add task/i }));
+      expect(lastModalProps()?.open).toBe(true);
+
+      act(() => {
+        lastModalProps()?.onOpenChange(false);
+      });
+
+      expect(lastModalProps()?.open).toBe(false);
     });
   });
 });
--- a/src/components/tasks/__tests__/task-list.test.tsx
+++ b/src/components/tasks/__tests__/task-list.test.tsx
@@ -263,6 +263,17 @@ describe("TaskList", () => {
       ).toBeInTheDocument();
     });
 
+    it("disables the Add Task button when no lists are enabled", () => {
+      const noEnabledConfig: TaskListConfig = {
+        ...mockConfig,
+        lists: mockConfig.lists.map((l) => ({ ...l, enabled: false })),
+      };
+      render(<TaskList config={noEnabledConfig} />);
+      expect(screen.getByRole("button", { name: /add task/i })).toBeDisabled();
+      expect(lastModalProps()?.availableLists).toEqual([]);
+      expect(lastModalProps()?.defaultListId).toBeUndefined();
+    });
+
     it("renders the modal closed by default", () => {
       render(<TaskList config={mockConfig} />);
       expect(lastModalProps()?.open).toBe(false);
--- a/src/components/tasks/new-task-modal.tsx
+++ b/src/components/tasks/new-task-modal.tsx
@@ -58,16 +58,20 @@ function buildInitialState(defaultListId: string | undefined): FormState {
 }
 
 /**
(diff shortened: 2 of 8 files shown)


src/components/tasks/new-task-modal.tsx
```diff
@@ -0,0 +1,274 @@
...
+ * Convert an HTML `<input type="date">` value (YYYY-MM-DD) to RFC 3339,
+ * which is what the Google Tasks API expects on the `due` field. Empty
+ * strings collapse to `undefined` so the field is omitted entirely.
+ */
+function toApiDue(value: string): string | undefined {
+  if (!value) return undefined;
+  const match = /^(\d{4})-(\d{2})-(\d{2})$/.exec(value);
+  if (!match) return undefined;
+  const [, y, m, d] = match;
+  return `${y}-${m}-${d}T00:00:00.000Z`;
```
BenSeymourODB: **Bug — `toApiDue` hardcodes UTC midnight, which shifts the date for users west of UTC.**

For a user in `America/New_York` (UTC-5), `2026-05-10T00:00:00.000Z` parses locally to May 9 19:00; `formatDueDate` in `src/components/tasks/types.ts` then `setHours(0,0,0,0)`s it down to May 9 — a one-day-off bug on round-trip display.

Fix: parse the date input as local midnight and let `toISOString()` produce the correctly-offset UTC instant. That round-trips back to the same local calendar date in any timezone:

```ts
const local = new Date(Number(y), Number(m) - 1, Number(d));
return local.toISOString();
```

Will also add a unit test covering the exact output rather than the current `expect.stringMatching(/^2026-05-10/)` indirect check.

---

src/components/tasks/new-task-modal.tsx
```diff
@@ -0,0 +1,274 @@
...
+  // Reset on open (mirrors EventCreateDialog's "store information from previous renders" pattern).
+  if (open !== prevOpen) {
+    setPrevOpen(open);
+    if (open) {
+      setState(buildInitialState(defaultListId));
+      setErrors({});
+    }
+  }
+
+  const handleSubmit = async (e: React.FormEvent) => {
```
BenSeymourODB, line 113: **Missing double-submit guard — a second submit while the request is in flight will fire a second POST.**

The submit button is `disabled` while loading, but `<form onSubmit>` is still live; pressing Enter (or any race against the disabled paint) re-enters `handleSubmit`. Add an early return:

```ts
const handleSubmit = async (e: React.FormEvent) => {
  e.preventDefault();
  if (loading) return;
  // ...
```

Will add a covering test.

---

src/components/tasks/new-task-modal.tsx
```diff
@@ -0,0 +1,274 @@
...
+              id={notesId}
+              value={state.notes}
+              placeholder="Optional details"
+              onChange={(e) =>
+                setState((prev) => ({ ...prev, notes: e.target.value }))
+              }
+            />
+          </div>
+
+          {submitError && (
```
BenSeymourODB, line 256: **`submitErrorId` is generated but never wired to anything — screen readers won't connect the error banner to the submit control.**

The banner has `role="alert"` so insertion is announced, but the submit button gives no follow-up association. Wire it the same way the inline field errors are wired to their inputs:

```diff
- <Button type="submit" disabled={loading}>
+ <Button
+   type="submit"
+   disabled={loading}
+   aria-describedby={submitError ? submitErrorId : undefined}
+ >
```

---

src/components/tasks/use-create-task.ts
```diff
@@ -0,0 +1,77 @@
...
+  notes?: string;
+}
+
+export interface UseCreateTaskReturn {
+  createTask: (input: CreateTaskInput) => Promise<GoogleTask>;
+  loading: boolean;
+  error: Error | null;
+}
+
+interface CreateTaskRequestBody {
```
BenSeymourODB: **`CreateTaskRequestBody` duplicates `CreateTaskInput` — drop the abstraction.**

Both interfaces are identical (`listId`, `title`, `due?`, `notes?`). The `buildRequestBody` indirection adds maintenance burden without behavior. Inline it:

```ts
body: JSON.stringify({
  listId: input.listId,
  title: input.title,
  ...(input.due !== undefined ? { due: input.due } : {}),
  ...(input.notes !== undefined ? { notes: input.notes } : {}),
}),
```

Removes ~15 lines.

---

src/components/tasks/__tests__/task-list.test.tsx
```diff
@@ -241,4 +249,121 @@ describe("TaskList", () => {
       expect(heading).toBeInTheDocument();
     });
   });
+
+  describe("add task footer", () => {
```
BenSeymourODB, line 253: **Missing test for the `enabledLists.length === 0` disabled-button branch.**

`task-list.tsx` disables the Add Task button when no lists are enabled, but no test pins that behavior. The empty-`availableLists` case is a real production state (user toggles everything off in settings) — without coverage we'd silently regress it.

```ts
it("disables the Add Task button when no lists are enabled", () => {
  const noEnabledConfig = {
    ...mockConfig,
    lists: [{ ...mockConfig.lists[0], enabled: false }],
  };
  render(<TaskList config={noEnabledConfig} />);
  expect(screen.getByRole("button", { name: /add task/i })).toBeDisabled();
});
```

---

src/components/tasks/new-task-modal.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  const [state, setState] = useState<FormState>(() =>
+    buildInitialState(defaultListId)
+  );
+  const [errors, setErrors] = useState<FormErrors>({});
+  const [prevOpen, setPrevOpen] = useState(open);
+
+  const { createTask, loading, error: submitError } = useCreateTask();
+
+  // Reset on open (mirrors EventCreateDialog's "store information from previous renders" pattern).
```
BenSeymourODB: **Document the limitation: `defaultListId` changes while the dialog is open are intentionally swallowed.**

The `if (open !== prevOpen)` reset block only fires on a closed → open transition. If the parent re-renders with new `defaultListId` or a different `availableLists` while the modal is open, the form state isn't refreshed — by design, to avoid clobbering the user's in-progress edits, mirroring the `EventCreateDialog` pattern. But unlike `defaultDate` (a hint), `defaultListId` feeds a required field — worth a one-line comment so a future reader doesn't think it's a bug.

---

BenSeymourODB (commented):
First-pass deep review (Sonnet subagent). Six line comments covering one real bug (timezone-shifting `due` field), one race (double-submit), one a11y gap (`submitErrorId` unwired), one cleanup (duplicate request-body type), one missing test branch (no enabled lists), and one documentation note (`defaultListId` mid-flight semantics). Plus a general note about plan deviations — all intentional, none blocking.

Will address all six in the next commit.

---

src/components/tasks/new-task-modal.tsx
BenSeymourODB: Fixed in `0424974`. `toApiDue` now parses as local midnight via `new Date(Number(y), Number(m) - 1, Number(d)).toISOString()`. Added three direct unit tests in `new-task-modal.test.tsx` (`describe("toApiDue")`) that pin the round-trip behavior — picking 2026-05-10 returns a UTC instant whose local re-parse is May 10 in any timezone, plus an undefined-on-empty test and an unparseable-input test.

---

src/components/tasks/new-task-modal.tsx
BenSeymourODB: Fixed in `0424974`. Added `if (loading) return;` at the top of `handleSubmit`, plus a covering test (`ignores submit attempts while a request is already in flight`) that sets `loading=true`, presses Enter inside the title input (which still triggers form submit even though the button is disabled), and asserts `mockCreateTask` is never called.

---

src/components/tasks/new-task-modal.tsx
BenSeymourODB: Fixed in `0424974`. The submit button now sets `aria-describedby={submitError ? submitErrorId : undefined}`. Added a test (`wires aria-describedby on the submit button when there is a server error`) that resolves the attribute via `document.getElementById` and asserts the linked banner contains the error message.

---

src/components/tasks/use-create-task.ts
BenSeymourODB: Fixed in `0424974`. Removed `CreateTaskRequestBody` and `buildRequestBody`; the body is now constructed inline at the `JSON.stringify` call site. The existing 10 hook tests cover the optional-field-omission behavior and continue to pass unchanged.

---

src/components/tasks/__tests__/task-list.test.tsx
BenSeymourODB: Added in `0424974` — `disables the Add Task button when no lists are enabled` in the `add task footer` describe. Goes one step further than the suggestion: also asserts the modal is rendered with `availableLists: []` and `defaultListId: undefined`, since those are the props the parent commits to when the dropdown is empty.

---

src/components/tasks/new-task-modal.tsx
BenSeymourODB: Documented in `0424974`. Added a NOTE to the existing `// Reset on open` comment block calling out that mid-session changes to `defaultListId` / `availableLists` are deliberately ignored and that the parent must close-then-reopen to seed a different default.

---

Issue #72, opened 2026-04-01
Performance optimization for large event datasets


BDSeymour:
## Description

Not optimized for very large event datasets.

## Potential Issues

- Rendering hundreds of events in a month
- Large IndexedDB datasets
- Slow event filtering/search

## Possible Improvements

- Virtualized rendering for long lists
- IndexedDB pagination
- Web Worker for event processing

## Priority

Low — wait for user feedback on performance.

---
*Migrated from `.claude/technical-debt.md` (Medium Priority #7)*

PR #196, opened 2026-05-02
feat(calendar): wire year view to CalendarProvider full-year loading (#117)

BenSeymourODB:
Closes #117.

## Summary

The default `CalendarProvider.refreshEvents()` window is `-calendarFetchMonthsBehind` to `+calendarFetchMonthsAhead` months from "now" — typically 7 months — which left the year view sparsely populated outside that window. `loadEventsForDate()` only widens by a single month at a time, so navigating to a different year via the prev/next buttons would still leave December (or January) of that year unloaded.

This PR adds `loadEventsForYear(year)` to the calendar context and wires `YearCalendar` to call it whenever the displayed year changes.

## Phases

- **Phase 1** — Provider TDD (red → green). Three tests pin the contract: extends `loadedRange` to cover Jan 1 → Dec 31, is a no-op when already covered, and bails out when called before the initial range is loaded. Implemented `loadEventsForYear` mirroring the lazy-merge semantics of `loadEventsForDate`. Threaded the new field through every existing mock `ICalendarContext` consumer (11 test files).
- **Phase 2** — Wired `YearCalendar` to call `loadEventsForYear(selectedDate.getFullYear())` via `useEffect`, keyed on the year (not the full Date) so moving around within a year doesn't churn fetches. Three component tests cover the trigger: on mount, on year change, and same-year no-op.

## Out of scope (deferred)

- A dedicated `/api/calendar/events?year=` endpoint — re-uses existing `/api/calendar/events`.
- Changes to dot rendering or click-through behavior — already wired and tested.
- Performance tuning for 250+ events tracked by #72.

## Test plan

- [x] `pnpm test:run src/components/providers/__tests__/CalendarProvider.test.tsx` — 11/11 (3 new)
- [x] `pnpm test:run src/components/calendar/__tests__/YearCalendar.test.tsx` — 24/24 (3 new)
- [x] `pnpm test:run` (full suite) — 1441/1441 across 98 files
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (0 errors; warnings are pre-existing in unrelated files)
- [ ] E2E: existing `e2e/year-calendar.spec.ts` runs against `MockCalendarProvider` (which has a no-op `loadEventsForYear`), so the visible behavior should be unaffected


BenSeymourODB (commented):
Overall the PR is solid: TDD workflow followed correctly (failing tests committed first in `524cc43`, implementation in `6a2404a`), no TypeScript `any`, no committed test artifacts, standard Tailwind palette throughout. A few substantive concerns noted below — the most important one is a silent data-loss window in `loadEventsForYear` when both edges need fetching.

---

BenSeymourODB (commented):
Inline findings from the Sonnet first-pass review (filed as a single batched review for thread visibility). The substantive items — the partial-failure stale-range bug, the useEffect churn from `loadedRange` in the dep array, and the fragile `setTimeout` polling in the idempotency test — will be addressed in a follow-up commit. Findings #3 and #4 are nits / pre-existing and will be acknowledged in the replies.

1. **`src/components/providers/CalendarProvider.tsx` ~L580** — Bug: when both edges need fetching and the second `fetchEventsForRange` throws, `setAllEvents` has already been called with the first batch but `setLoadedRange` is never reached (it sits after both `if` blocks inside the same `try`). The range pointer now lags behind the actually-stored events, so the next `loadEventsForYear` call re-fetches the first edge unnecessarily. `loadEventsForDate` has the same structure but only ever fetches one edge at a time, so the problem cannot manifest there. Recommendation: advance `setLoadedRange` immediately after each successful `setAllEvents`, rather than accumulating `newStart`/`newEnd` locally and committing both at the end.

2. **`src/components/providers/CalendarProvider.tsx` ~L513** — `loadEventsForYear` lists `loadedRange` in its `useCallback` deps. Every time `loadEventsForDate` (or `loadEventsForYear` itself) widens the range, the function gets a new identity and `YearCalendar`'s `useEffect([year, loadEventsForYear])` re-fires. The function is idempotent once the year is covered, so this is not a correctness bug — but it spams the console log and burns `isLoadingRangeRef` cycles on every navigation that triggers a range widen. Fix with the `refreshEventsRef` ref-stabilization pattern already used in this file, or omit `loadEventsForYear` from the YearCalendar effect deps with the same `eslint-disable-next-line react-hooks/exhaustive-deps` annotation used elsewhere.

3. **`src/components/providers/CalendarProvider.tsx` ~L524** — Nit: `yearEnd = new Date(year, 11, 31, 23, 59, 59, 999)`. Google Calendar treats `timeMax` as exclusive, so passing the start of the next year (`new Date(year + 1, 0, 1, 0, 0, 0, 0)`) is the canonical form. The current value works in practice. Not blocking.

4. **`src/components/providers/MockCalendarProvider.tsx`** — The new `loadEventsForYear = async () => {}` no-op stub is fine. Noting for awareness only: `useMemo` on line ~167 is a pre-existing CLAUDE.md violation (no manual memoization), unrelated to this PR.

5. **`src/components/providers/__tests__/CalendarProvider.test.tsx` ~L533** — Test fragility: the "no-op when already loaded" test uses raw `setTimeout(r, 10)` / `setTimeout(r, 50)` to wait for the in-flight guard to clear and to give a putative second-call fetch room to fire. These will flake under CI load. Replace with `waitFor` against an observable signal (e.g. wait until the fetch count grows by exactly the expected number for the first year-load, then assert the count is unchanged after the second click).

— Generated by Sonnet review agent.

---

BenSeymourODB:
Second-round response — addressing each finding from the Sonnet first-pass review (commit `42b63cc`).

**1. Partial-failure stale `loadedRange` (bug)** — Fixed. `setLoadedRange` is now called after each successful edge fetch using a functional update (`(prev) => prev ? { ...prev, start: yearStart } : prev`), rather than batching both updates at the end. Added a regression test (`advances loadedRange for the successful edge when the second edge fetch fails`) that:
   - Picks `targetYear = currentYear` so both edges fall outside the default `-1mo / +6mo` window and both `if` branches fire.
   - Stubs `fetchMock` so the *second* `/events` call after the click rejects.
   - Asserts the retry hits a single edge (Dec) and that its `timeMax` covers Dec 31 — proving the Jan edge's range advance survived.
   - Verified the test fails on the unfixed code by `git stash push`-ing the impl, running the new test, then restoring (failing → passing red→green confirmed).

**2. `loadEventsForYear` identity churn driving redundant effect runs** — Fixed by ref-stabilization in `YearCalendar`. Added a `loadEventsForYearRef`, mirrored the `refreshEventsRef` pattern that was already established in `CalendarProvider`, and the year-change effect now only depends on `[year]`. The provider-side `useCallback` keeps `loadedRange` in its deps (consistent with `loadEventsForDate`) — the stabilization happens at the consumer where the effect lives, which is both more local and avoids changing the public-facing identity contract for any other future consumers.

**3. `yearEnd = Dec 31 23:59:59.999` vs. exclusive `Jan 1 next year` (nit)** — Acknowledged, not changed. Google Calendar's `timeMax` is treated as exclusive, so `23:59:59.999` of Dec 31 already loses no events (a `dateTime` of exactly `YYYY-12-31T23:59:59.999Z` is the latest representable instant of the year). This matches `date-fns`'s `endOfDay` convention used elsewhere in the codebase. Switching to `Jan 1 next year` would be cosmetic.

**4. `useMemo` in `MockCalendarProvider` (pre-existing CLAUDE.md violation, noted in passing)** — Confirmed pre-existing on `MockCalendarProvider.tsx:163`. Out of scope for this PR; the new `loadEventsForYear = async () => {}` no-op stub introduces no new memoization. Flagging for a follow-up cleanup ticket if it's worth chasing.

**5. Fragile `setTimeout` polling in the no-op idempotency test** — Fixed. The two raw `setTimeout(r, 10)` and `setTimeout(r, 50)` waits are gone:
   - The first is replaced by a precise `waitFor(() => expect(eventFetches().length).toBe(2))` that exits exactly when the Dec-edge fetch has resolved AND its setState chain has flushed (the loader's awaited fetch + sync `setAllEvents`/`setLoadedRange` + `finally` all run in one microtask burst once the mock resolves).
   - The second becomes a stability assertion: `waitFor(() => expect(count).toBe(2), { timeout: 200, interval: 25 })` polls for ~200ms and fails if the count ever advances past the expected value.

Full results post-fix: 1442/1442 unit tests passing across 98 files; `pnpm lint:fix && pnpm format:fix && pnpm check-types` clean (warnings are pre-existing in unrelated files).

---

PR #197, opened 2026-05-02
feat(calendar): wire EventDetailModal delete to Google Calendar (#115)

BenSeymourODB:
Closes #115 (delete half) — edit half deferred to a follow-up PR.

## Scope

This PR delivers the **delete** path of #115 end-to-end. The cluster note in #115 says it should design the per-event mutation API route first, so this PR establishes that route shape and uses it from the modal. Edit (PATCH) and the modal's edit form will follow in a separate PR and reuse the same auth/error-handling skeleton.

## Phases (each pushed as its own commit)

- **Phase 1** — `DELETE /api/calendar/events/[id]?calendarId=...` server route. Forwards to Google Calendar `events.delete`, collapses Google's 410 Gone into a 204 (already-deleted is success), and surfaces 401 Google rejections through the existing `requiresReauth` re-auth flow. 12 unit tests pin the auth, validation, success, and error branches.
- **Phase 2** — `deleteEvent(eventId, calendarId)` on `CalendarProvider` with optimistic remove + rollback on failure. Exposed via the existing context.
- **Phase 3** — Delete button + `AlertDialog` confirmation in `EventDetailModal`, wired to the provider. Sonner toast on success/failure. Modal closes only after the optimistic delete commits so focus restoration still hits the original trigger.
- **Phase 4** — Playwright happy-path E2E covering trigger → confirm → toast → row removed.

## Out of scope (explicitly deferred)

- Edit (PATCH) — separate PR, will reuse this route file.
- Per-calendar `accessRole` gating (read-only calendars hide the delete button) — Google still enforces it server-side, so the worst case today is a clear 403 toast.
- IndexedDB cache surgery for the deleted event — covered by the next refresh cycle today; finer-grained invalidation can land with the cluster work.

## Test plan

- [x] `pnpm test` — full suite green
- [x] Playwright E2E happy path
- [ ] Manual smoke against a real Google calendar (PR author / reviewer)

https://claude.ai/code/session_01KmoWuSvtqRARyfY1d6SmdE

---
_Generated by [Claude Code](https://claude.ai/code/session_01AzA94swSA6AaiycvGGBPht)_

### DIFF
--- a/e2e/event-detail-modal-delete.spec.ts
+++ b/e2e/event-detail-modal-delete.spec.ts
@@ -12,7 +12,10 @@ import { expect, test } from "@playwright/test";
  * covered by the route unit tests.
  */
 
-test.use({ video: "on" });
+// Keep video only when something fails so green CI runs don't churn out
+// gigabytes of useless artifacts. CLAUDE.md forbids committing them either way,
+// but this also keeps the local test-results dir lean.
+test.use({ video: "retain-on-failure" });
 
 test.describe("EventDetailModal — delete (#115)", () => {
   test("clicking 'Delete event' and confirming removes the event from the calendar", async ({
--- a/e2e/event-detail-modal-delete.spec.ts
+++ b/e2e/event-detail-modal-delete.spec.ts
@@ -1,0 +1,82 @@
+import { expect, test } from "@playwright/test";
+
+/**
+ * E2E for the EventDetailModal delete flow — issue #115.
+ *
+ * Exercises the full confirmation path against the test calendar's
+ * MockCalendarProvider: open modal → click "Delete event" → confirm →
+ * modal closes and the event row disappears from month view.
+ *
+ * The mock provider stubs `deleteEvent` with an in-memory remove (no
+ * Google API call) so this test is hermetic; the real network path is
+ * covered by the route unit tests.
+ */
+
+test.use({ video: "on" });
+
+test.describe("EventDetailModal — delete (#115)", () => {
+  test("clicking 'Delete event' and confirming removes the event from the calendar", async ({
+    page,
+  }) => {
+    await page.goto("/test/calendar?events=single&view=month");
+
+    const trigger = page.getByRole("button", { name: "Single Event" });
+    await expect(trigger).toBeVisible();
+
+    await trigger.click();
+
+    const dialog = page.getByRole("dialog");
+    await expect(dialog).toBeVisible();
+
+    const deleteBtn = dialog.getByRole("button", { name: /delete event/i });
+    await expect(deleteBtn).toBeVisible();
+    await deleteBtn.click();
+
+    const confirm = page.getByRole("alertdialog", {
+      name: /delete this event\?/i,
+    });
+    await expect(confirm).toBeVisible();
+    await expect(confirm).toContainText("Single Event");
+
+    await confirm.getByRole("button", { name: /yes, delete/i }).click();
+
+    // Modal + alert dialog both dismiss on success.
+    await expect(page.getByRole("alertdialog")).toBeHidden();
+    await expect(page.getByRole("dialog")).toBeHidden();
+
+    // The event row is gone from the month grid.
+    await expect(
+      page.getByRole("button", { name: "Single Event" })
+    ).toHaveCount(0);
+  });
+
+  test("cancelling the confirmation keeps the event in the calendar", async ({
+    page,
+  }) => {
+    await page.goto("/test/calendar?events=single&view=month");
+
+    const trigger = page.getByRole("button", { name: "Single Event" });
+    await trigger.click();
+    await expect(page.getByRole("dialog")).toBeVisible();
+
+    await page
+      .getByRole("dialog")
+      .getByRole("button", { name: /delete event/i })
+      .click();
+    await page
+      .getByRole("alertdialog")
+      .getByRole("button", { name: /cancel/i })
+      .click();
+
+    await expect(page.getByRole("alertdialog")).toBeHidden();
+
+    // The detail dialog stays open after cancelling.
+    await expect(page.getByRole("dialog")).toBeVisible();
+
+    // Dismiss the detail dialog and confirm the event row is still present
+    // on the grid (cancel must not have triggered a deletion).
+    await page.keyboard.press("Escape");
+    await expect(page.getByRole("dialog")).toBeHidden();
+    await expect(trigger).toBeVisible();
+  });
+});
--- a/src/app/api/calendar/events/[id]/route.ts
+++ b/src/app/api/calendar/events/[id]/route.ts
@@ -76,7 +76,6 @@ export async function DELETE(request: NextRequest, { params }: RouteContext) {
       method: "DELETE",
       headers: {
         Authorization: `Bearer ${accessToken}`,
-        "Content-Type": "application/json",
       },
     });
 
@@ -109,6 +108,11 @@ export async function DELETE(request: NextRequest, { params }: RouteContext) {
     }
 
     if (response.status === 403) {
+      logger.error(new Error("Google denied delete on calendar (403)"), {
+        userId: session.user.id,
+        calendarId,
+        eventId,
+      });
       return NextResponse.json(
         {
           error:
--- a/src/app/api/calendar/events/[id]/route.ts
+++ b/src/app/api/calendar/events/[id]/route.ts
@@ -1,0 +1,156 @@
+/**
+ * Per-event mutation API for Google Calendar.
+ *
+ * This route file establishes the shared auth + Google-API forwarding shape
+ * for the calendar CRUD cluster (#115/#116/#118). Today only `DELETE` is
+ * implemented — `PATCH` (edit) and additional verbs land in follow-up PRs
+ * and reuse this file's helpers.
+ *
+ * Query params:
+ * - `calendarId` (required): the Google Calendar that owns the event.
+ *
+ * Path:
+ * - `[id]`: the Google Calendar event id.
+ */
+import { AuthError, getAccessToken, getSession } from "@/lib/auth";
+import { logger } from "@/lib/logger";
+import { NextRequest, NextResponse } from "next/server";
+
+const GOOGLE_CALENDAR_API = "https://www.googleapis.com/calendar/v3";
+
+interface RouteContext {
+  params: Promise<{ id: string }>;
+}
+
+/**
+ * DELETE /api/calendar/events/[id]?calendarId=...
+ *
+ * Forwards to Google Calendar `events.delete`. A 204 from Google (or a 410
+ * "already gone") collapses to a 204 here so the client can treat the event
+ * as gone either way. Auth failures from Google surface as 401 with
+ * `requiresReauth: true` so the existing client re-auth flow kicks in.
+ */
+export async function DELETE(request: NextRequest, { params }: RouteContext) {
+  try {
+    const session = await getSession();
+    if (!session?.user) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    if (session.error === "RefreshTokenError") {
+      return NextResponse.json(
+        {
+          error: "Session expired. Please sign in again.",
+          requiresReauth: true,
+        },
+        { status: 401 }
+      );
+    }
+
+    const { id } = await params;
+    const eventId = id?.trim();
+    if (!eventId) {
+      return NextResponse.json(
+        { error: "Event id is required" },
+        { status: 400 }
+      );
+    }
+
+    const calendarId = new URL(request.url).searchParams
+      .get("calendarId")
+      ?.trim();
+    if (!calendarId) {
+      return NextResponse.json(
+        { error: "calendarId query parameter is required" },
+        { status: 400 }
+      );
+    }
+
+    const accessToken = await getAccessToken();
+
+    const apiUrl = `${GOOGLE_CALENDAR_API}/calendars/${encodeURIComponent(
+      calendarId
+    )}/events/${encodeURIComponent(eventId)}`;
+
+    const response = await fetch(apiUrl, {
+      method: "DELETE",
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        "Content-Type": "application/json",
+      },
+    });
+
+    // Google returns 204 No Content on success and 410 Gone if the event was
+    // already deleted. Both mean the event is gone — collapse to 204 so the
+    // optimistic client UI doesn't have to special-case the race.
+    if (response.ok || response.status === 410) {
+      logger.event("CalendarEventDeleted", {
+        userId: session.user.id,
+        calendarId,
+        eventId,
+        googleStatus: response.status,
+      });
+      return new NextResponse(null, { status: 204 });
+    }
+
+    if (response.status === 401) {
+      logger.error(new Error("Google rejected access token on event delete"), {
+        userId: session.user.id,
+        calendarId,
+        eventId,
+      });
+      return NextResponse.json(
+        {
+          error: "Google authentication failed. Please sign in again.",
+          requiresReauth: true,
+        },
+        { status: 401 }
+      );
+    }
+
+    if (response.status === 403) {
+      return NextResponse.json(
+        {
+          error:
+            "You do not have permission to delete events on this calendar.",
+        },
+        { status: 403 }
+      );
+    }
+
+    if (response.status === 404) {
+      return NextResponse.json({ error: "Event not found" }, { status: 404 });
+    }
+
+    const errorBody = await response.json().catch(() => ({}));
+    logger.error(new Error("Google Calendar delete failed"), {
+      userId: session.user.id,
+      calendarId,
+      eventId,
+      googleStatus: response.status,
+      googleError: errorBody?.error?.message ?? "unknown",
+    });
+
+    return NextResponse.json(
+      { error: "Failed to delete calendar event" },
+      { status: 502 }
+    );
+  } catch (error) {
+    if (error instanceof AuthError) {
+      return NextResponse.json(
+        { error: error.message, requiresReauth: error.status === 401 },
+        { status: error.status }
+      );
+    }
+
+    logger.error(error as Error, {
+      endpoint: "/api/calendar/events/[id]",
+      method: "DELETE",
+    });
+
+    return NextResponse.json(
+      { error: "An unexpected error occurred" },
+      { status: 500 }
+    );
+  }
+}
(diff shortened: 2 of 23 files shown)


BenSeymourODB (commented):
Deep file-level review of the delete-half of #115. Overall the implementation is solid — the route error mapping is correct, the AlertDialog pattern is idiomatic, and the test coverage is genuinely good. There are three issues worth addressing before merging (two blocking, one nit):

1. **Blocking — snapshot rollback clobbers a concurrent `refreshEvents` result.** If a background refresh completes while the DELETE is in-flight and fails, `setAllEvents(snapshot)` will overwrite the freshly-fetched list with stale data. Use a functional updater to restore only the single event rather than the whole snapshot.

2. **Blocking — `useCallback(async …, [])` violates the project-wide React Compiler rule.** CLAUDE.md is explicit: *"DO NOT use manual memoization (useMemo, useCallback, React.memo)"*. The five pre-existing `useCallback` calls in `CalendarProvider` are a pre-existing violation, but this PR adds a sixth. The function is not used as a dependency of any `useEffect`, so there is no stability requirement — remove the wrapper.

3. **Nit — `Content-Type: application/json` header on a body-less DELETE, and the 403 branch is not logged.**

Full details in the line comments below.

---

**What is NOT an issue (addressed preemptively):**

- The `useEventDelete` re-throw is intentional and the modal correctly stays open on failure — the contract is well-documented and tested.
- The AlertDialog being placed outside `<DialogContent>` but inside `<Dialog>` is the correct Radix pattern; each renders its own portal.
- The 410→204 collapse is correct and well-tested.
- IndexedDB not being touched on optimistic delete is explicitly called out as acceptable in the PR description ("covered by the next refresh cycle").
- Focus restoration after a failed delete is correct: the AlertDialog closes first (returning focus to the "Delete event" button inside the still-open Dialog), then `isDeleting` is reset — order is safe.
- `DayCalendar`, `WeekCalendar`, and `YearCalendar` do not use `EventDetailModal` at all, so the two call-sites (`SimpleCalendar`, `AgendaCalendar`) are exhaustive for this PR.
- The `DeleteProbe` test renders events by `e.title` which `baseEvent` sets to the id string — the assertions against `"evt-1"` / `"evt-2"` are correct.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -215,6 +221,49 @@ export function CalendarProvider({
...
+          } catch {
+            // Some failures (e.g. 502 from a proxy) may not have a JSON body.
+          }
+          throw new Error(message);
+        }
+
+        logger.event("CalendarEventDeleted", { eventId, calendarId });
+      } catch (error) {
+        // Rollback the optimistic remove so the UI matches reality.
+        setAllEvents(snapshot);
```
BenSeymourODB: **[Blocking] Snapshot rollback will clobber a concurrent `refreshEvents` result**

The current rollback pattern stores the entire previous list in `snapshot` and restores it unconditionally on failure:

```ts
let snapshot: IEvent[] = [];
setAllEvents((prev) => {
  snapshot = prev;
  return prev.filter((e) => e.id !== eventId);
});
// ... await fetch() ...
// on failure:
setAllEvents(snapshot);  // bare value, not a functional update
```

If `refreshEvents` fires and resolves while the `DELETE` is in-flight (the auto-refresh interval is user-configurable and can be set very low), `setAllEvents(transformedEvents)` inside `refreshEvents` will run first and replace the list with fresh server data. Then when the `DELETE` fails, `setAllEvents(snapshot)` overwrites that fresh data with the stale pre-delete snapshot — undoing the refresh for the rest of the session.

**Fix:** surgically re-insert only the removed event using a functional updater so concurrent refreshes are preserved:

```ts
// Capture the full event before the async gap (not inside the updater)
const removedEvent = allEvents.find((e) => e.id === eventId);
setAllEvents((prev) => prev.filter((e) => e.id !== eventId));

try {
  // ... await fetch ...
} catch (error) {
  if (removedEvent) {
    setAllEvents((prev) => {
      // Guard: a concurrent refresh may have already re-added the event
      if (prev.some((e) => e.id === removedEvent.id)) return prev;
      return [...prev, removedEvent];
    });
  }
  logger.error(error as Error, { context: "deleteEvent", eventId, calendarId });
  throw error;
}
```

Secondary concern: the current `snapshot = prev` assignment inside a `setAllEvents` updater is not idempotent — React 18 Strict Mode calls updaters twice in development, so `snapshot` ends up as the value from the _second_ call (same array in practice, but the pattern is fragile). Reading from `allEvents` directly (as shown above) is clean and avoids the issue.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -215,6 +221,49 @@ export function CalendarProvider({
...
+        logger.error(error as Error, {
+          context: "deleteEvent",
+          eventId,
+          calendarId,
+        });
+        throw error;
+      }
+    },
+    []
```
BenSeymourODB: **[Blocking] `useCallback` with empty dep array violates the React Compiler rule in CLAUDE.md**

CLAUDE.md is explicit: *"DO NOT use manual memoization (`useMemo`, `useCallback`, `React.memo`). Write clean, simple React code."*

`deleteEvent` is placed in the context value object (`value.deleteEvent`) on every render, but it is not a dependency of any `useEffect` inside the provider, so there is no identity-stability requirement that would justify a `useCallback`. The React Compiler handles memoization automatically.

The five pre-existing `useCallback` calls in this file (`fetchEventsForRange`, `fetchCalendarList`, `refreshEvents`, `loadEventsForDate`) are a pre-existing violation; this PR should not add a sixth. Simply unwrap to a plain `async function`:

```ts
const deleteEvent = async (eventId: string, calendarId: string) => {
  // ... same body ...
};
```

If a future `useEffect` genuinely needs a stable reference to `deleteEvent`, add the `useCallback` at that point with the appropriate dep array. An empty dep array is also a lint smell — `react-hooks/exhaustive-deps` (enabled via `eslint-config-next`) will warn on it because the function closes over `allEvents` (needed for the fix in the previous comment) and `logger`.

---

src/app/api/calendar/events/[id]/route.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    const apiUrl = `${GOOGLE_CALENDAR_API}/calendars/${encodeURIComponent(
+      calendarId
+    )}/events/${encodeURIComponent(eventId)}`;
+
+    const response = await fetch(apiUrl, {
+      method: "DELETE",
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        "Content-Type": "application/json",
```
Comment (BenSeymourODB) [inline src/app/api/calendar/events/[id]/route.ts]: **[Nit] `Content-Type: application/json` is meaningless on a body-less DELETE**

The Google Calendar `events.delete` API requires no request body. Sending `Content-Type: application/json` on a DELETE with no body is harmless but misleading — some strict proxies or API gateways log a warning for it, and it differs from the existing `events` GET route which sets no `Content-Type`. Drop it to keep the two routes consistent:

```ts
const response = await fetch(apiUrl, {
  method: "DELETE",
  headers: {
    Authorization: `Bearer ${accessToken}`,
  },
});
```

---

src/app/api/calendar/events/[id]/route.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    if (response.status === 403) {
+      return NextResponse.json(
+        {
+          error:
+            "You do not have permission to delete events on this calendar.",
+        },
+        { status: 403 }
+      );
```
Comment (BenSeymourODB) [inline src/app/api/calendar/events/[id]/route.ts:122]: **[Nit] 403 branch has no `logger.error` call — inconsistent with 401 and 5xx branches**

The 401-from-Google branch (line 97) and the catch-all 5xx branch (line 126) both call `logger.error`. The 403 branch silently returns a JSON response with no server-side trace. A 403 on delete (read-only calendar) is operationally interesting — it means someone is trying to delete from a calendar they cannot write to — and is worth logging at least at `logger.log` level for observability.

```ts
if (response.status === 403) {
  logger.log("Attempted delete on read-only calendar", {
    userId: session.user.id,
    calendarId,
    eventId,
  });
  return NextResponse.json(
    { error: "You do not have permission to delete events on this calendar." },
    { status: 403 }
  );
}
```

---

src/components/calendar/EventDetailModal.tsx
```diff
@@ -66,13 +86,33 @@ export function EventDetailModal({
...
   const timeRange = formatTimeRange(event, use24HourFormat);
   const dateLabel = formatDateLabel(event);
   const hasDescription = Boolean(event.description?.trim());
 
+  const handleConfirmDelete = async () => {
+    if (!onDelete) return;
+    setIsDeleting(true);
+    try {
+      await onDelete(event);
+      setConfirmingDelete(false);
```
BenSeymourODB: **[Nit] `setConfirmingDelete(false)` in the success path is a no-op and could be misleading**

On success the flow is:
1. `await onDelete(event)` resolves
2. `setConfirmingDelete(false)` — but `onClose()` (line 106) immediately sets `selectedEvent = null` in the parent, which unmounts this component entirely

The state update is therefore dead code and may mislead future readers into thinking the AlertDialog needs to be explicitly closed before `onClose` runs. Consider removing it from the success path and leaving it only in the `catch` block where it actually controls visible UI:

```ts
const handleConfirmDelete = async () => {
  if (!onDelete) return;
  setIsDeleting(true);
  try {
    await onDelete(event);
    onClose(); // unmounts; no need to reset confirmingDelete first
  } catch {
    setConfirmingDelete(false); // keep modal open; caller shows toast
  } finally {
    setIsDeleting(false);
  }
};
```

---

src/hooks/useEventDelete.ts
```diff
@@ -0,0 +1,28 @@
...
+ * stays open on failure so the user can retry).
+ */
+import { useCalendar } from "@/components/providers/CalendarProvider";
+import type { IEvent } from "@/types/calendar";
+import { toast } from "sonner";
+
+export function useEventDelete() {
+  const { deleteEvent } = useCalendar();
+
+  return async (event: IEvent) => {
```
BenSeymourODB, line 17: **[Nit] The returned async function is recreated on every render — consider whether a stable reference is ever needed**

`useEventDelete` returns a new `async (event: IEvent) => …` closure on every call. In `SimpleCalendar` and `AgendaCalendar` the returned `handleDelete` is passed directly as `onDelete` to `EventDetailModal`, whose prop is typed `(event: IEvent) => Promise<void>`. Because `EventDetailModal` receives it as an optional prop and does not pass it to any `useEffect`, the identity instability is benign today.

This is fine as-is given the React Compiler is enabled (it will memoize the closure automatically). Just flagging it so the next developer adding a `useEffect` dependency on `handleDelete` knows to revisit — or can simply note "React Compiler handles this".

---

src/components/providers/__tests__/CalendarProvider.test.tsx
```diff
@@ -429,4 +429,163 @@ describe("CalendarProvider", () => {
...
+
+      await userEvent.setup().click(screen.getByText("delete"));
+
+      await waitFor(() => {
+        expect(screen.queryByText("evt-2")).not.toBeInTheDocument();
+        expect(screen.getByText("evt-1")).toBeInTheDocument();
+      });
+    });
+  });
 });
```
BenSeymourODB, line 679: **[Test gap] Missing: concurrent `refreshEvents` + failed DELETE race**

The three `deleteEvent` tests cover: optimistic remove on success, rollback on API failure, and 204-with-no-body. What is not covered is the race described in the `setAllEvents(snapshot)` comment above: a refresh completing _between_ the optimistic remove and the failed DELETE confirmation.

Suggested test sketch (add alongside the three existing ones):

```ts
it("does not clobber a concurrent refresh when the DELETE fails", async () => {
  let resolveDelete!: (v: Response) => void;
  const deletePromise = new Promise<Response>((res) => (resolveDelete = res));

  seedFetch(() => deletePromise);
  // ... render, wait for events ...

  // Start the delete (optimistic remove fires immediately)
  const deleteClick = userEvent.setup().click(screen.getByText("delete"));

  // Simulate a refresh completing while DELETE is in-flight
  // (refreshEvents replaces allEvents with fresh data that includes both events)
  // In a real test you'd trigger this via the interval or a direct refreshEvents call.
  // The key assertion: after resolveDelete({ ok: false, ... }), evt-1 is back
  // AND evt-2 (which was returned by the refresh) is still present.
  resolveDelete({ ok: false, status: 500, json: async () => ({ error: "boom" }) } as unknown as Response);
  await deleteClick;

  expect(screen.getByText("evt-1")).toBeInTheDocument();
  expect(screen.getByText("evt-2")).toBeInTheDocument();
});
```

This test will fail with the current snapshot-based rollback, confirming the bug, and pass once the functional-updater fix is applied.

---

e2e/event-detail-modal-delete.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+ *
+ * Exercises the full confirmation path against the test calendar's
+ * MockCalendarProvider: open modal → click "Delete event" → confirm →
+ * modal closes and the event row disappears from month view.
+ *
+ * The mock provider stubs `deleteEvent` with an in-memory remove (no
+ * Google API call) so this test is hermetic; the real network path is
+ * covered by the route unit tests.
+ */
+
```
BenSeymourODB, line 14: **[Nit] `test.use({ video: "on" })` will produce video artifacts on every run**

`video: "on"` records video unconditionally, including on green CI runs. The CLAUDE.md checklist item 4 says *"Never commit test output artifacts (`test-results/`, `playwright-report/`, `blob-report/`)"* — videos fall into this category. Prefer `video: "retain-on-failure"` which only keeps the recording when a test fails (no storage cost on green runs, but immediately useful when debugging a flake):

```ts
test.use({ video: "retain-on-failure" });
```

Or omit the override entirely and let the project-level `playwright.config.ts` default control it.

---

BenSeymourODB (commented):
Deep file-level review of the delete-half of #115. Overall the implementation is solid — the route error mapping is correct and complete, the AlertDialog pattern is idiomatic Radix, the toast contract is well-specified, and the unit/integration/E2E coverage is genuinely good. Two issues need to be fixed before merging; the rest are nits.

**Blocking**

1. **Snapshot rollback race** (`CalendarProvider.tsx` line 255): `setAllEvents(snapshot)` on failure overwrites any fresh data that `refreshEvents` wrote to state during the `await fetch()`. Fix with a functional updater that only re-inserts the one removed event, guarded against duplicates if a concurrent refresh already restored it.

2. **`useCallback` with empty dep array** (`CalendarProvider.tsx` line 264): CLAUDE.md prohibits manual memoization. `deleteEvent` has no `useEffect` dependency so no stability requirement exists. Drop the `useCallback` wrapper. (Also: an empty dep array will trigger `react-hooks/exhaustive-deps` once the fix from point 1 adds `allEvents` to the closure.)

**Nits** (see individual line comments for detail):
- `Content-Type: application/json` on a body-less DELETE (route.ts line 79)
- 403 branch has no `logger` call — inconsistent with 401 and 5xx (route.ts line 118)
- `setConfirmingDelete(false)` in the success path of `handleConfirmDelete` is dead code (modal unmounts immediately via `onClose`) — remove to avoid misleading future readers
- `test.use({ video: "on" })` should be `"retain-on-failure"` to avoid unconditional video artifacts in CI
- Missing test: concurrent refresh + failed DELETE race (suggested sketch in the CalendarProvider test comment)

**Confirmed non-issues:**
- 410→204 collapse: correct and tested
- AlertDialog outside `<DialogContent>` but inside `<Dialog>`: correct Radix pattern (separate portals)
- Focus restoration on failure: correct — AlertDialog closes first (focus lands on "Delete event" button inside the still-open Dialog), `isDeleting` resets in `finally`
- `useEventDelete` re-throw contract: intentional and documented, matches what `EventDetailModal` expects
- IndexedDB not updated on optimistic delete: explicitly acknowledged as acceptable in PR description
- `DayCalendar`, `WeekCalendar`, `YearCalendar`: confirmed do not use `EventDetailModal` — the two wired call-sites are exhaustive

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in d5b6acf. Captured the removed event before the async gap, removed via functional updater, and on failure re-insert with a `prev.some(...)` guard so a concurrent refresh that already restored the event isn't double-added. Strict-mode double-invocation concern goes away with the snapshot pattern.

Added a regression test (`CalendarProvider.test.tsx` "rollback after failure preserves any concurrent additions to the list") that holds the DELETE response open, mutates the list mid-flight via `addEvent`, then rejects — failing on the old code, passing on the new.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in d5b6acf — `deleteEvent` is now a plain `async function`. Agreed on the rationale; the React Compiler handles memoization and there's no `useEffect` consumer that needs a stable identity. The five pre-existing `useCallback` calls in this file remain a separate cleanup that's out of scope here.

---

src/app/api/calendar/events/[id]/route.ts
Comment (BenSeymourODB) [reply src/app/api/calendar/events/[id]/route.ts]: Fixed in d5b6acf — `Content-Type` removed, now matches the GET route's headers shape.

---

src/app/api/calendar/events/[id]/route.ts
Comment (BenSeymourODB) [reply src/app/api/calendar/events/[id]/route.ts:122]: Fixed in d5b6acf. Went with `logger.error` (rather than `logger.log`) for parity with the 401 and 5xx branches above — a 403 on a delete is unusual enough to warrant the same severity.

---

src/components/calendar/EventDetailModal.tsx
BenSeymourODB: Fixed in d5b6acf — dropped from the success path. The `catch` branch keeps it (still useful: it dismisses the alert so the user lands back on the still-open detail modal).

---

src/hooks/useEventDelete.ts
BenSeymourODB: Acknowledged — leaving as-is. The hook intentionally returns a fresh closure so the React Compiler memoizes it and `useEffect` callers won't see identity churn. If a future caller needs a stable reference, the hook can wrap the inner function in `useCallback` at that time.

---

src/components/providers/__tests__/CalendarProvider.test.tsx
BenSeymourODB: Added in d5b6acf — `"rollback after failure preserves any concurrent additions to the list"`. Used `addEvent` to simulate the concurrent write rather than re-running `refreshEvents` so the test stays focused on the rollback contract. Verified the test fails on the pre-fix snapshot pattern and passes on the functional updater.

---

e2e/event-detail-modal-delete.spec.ts
BenSeymourODB: Fixed in d5b6acf — switched to `"retain-on-failure"` with a comment pointing at the CLAUDE.md artifact rule. Verified locally that no `test-results/` artifacts are produced on green runs.

---

Issue #173, opened 2026-04-27
feat(rewards): add PointsBadge + PointsAnimation, wire to TaskItem and Settings


BDSeymour:
Add the visible reward surface and wire it into task completion + the settings page.

## Scope

- `src/components/rewards/points-badge.tsx` — header badge: 🏆 + locale-formatted total points + "pts". Hidden when rewards are disabled.
- `src/components/rewards/points-animation.tsx` — fade-in-up "+X points! 🎉" overlay shown for ~2s after a completion. Respects `prefers-reduced-motion`.
- Update `task-item.tsx` (from the Tasks track) to call `awardPoints(defaultTaskPoints, 'task_completed', { taskId, taskTitle })` on first-time completion only. Show `PointsAnimation` if `showPointsOnCompletion` is true.
- Add a "Reward System" section to the settings page: enable/disable toggle, default points input (number, min 1), show-points-on-completion toggle, total-points display.
- Mount `PointsProvider` in the root layout so the badge can render.

## Acceptance

- Toggling tasks back to incomplete and re-completing does NOT award additional points (server enforces or guard via taskId+reason uniqueness — see API issue).
- Settings UI persists via the existing `userSettings` API route.
- Component test for `PointsBadge` (hidden when disabled, shown when enabled).
- Playwright happy-path: enable rewards → complete task → see animation + badge increment.

## Blocked by

- Rewards API + PointsContext issue.
- Tasks TaskList components issue (Tasks track) — for the wire-up point.

## Plan reference

`.claude/plans/reward-point-system.md` (sections "Points Badge", "Points Animation", "Updated TaskItem with Points", "Integration with App Layout").

## Track

Day 8 (Rewards track).

BenSeymourODB:
Picking this up on `claude/loving-faraday-OKZbI` — schema (#171), API + PointsContext (#172) and the settings toggle landed already; this PR will add the visible surface (PointsBadge, PointsAnimation, TaskItem wire-up, total-points display in settings) and mount the provider in the root layout.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #171, opened 2026-04-27
feat(rewards): add Prisma schema for RewardPoints + PointTransaction + UserSettings flags


BDSeymour:
Add the database models for the reward-point system. Single migration alongside the UserSettings additions.

## Scope (schema)

- `RewardPoints { id, userId @unique, totalPoints (default 0), updatedAt }`.
- `PointTransaction { id, userId, points, reason ('task_completed'|'bonus'|'manual'|'streak'), taskId?, taskTitle?, createdAt }` with `@@index([userId, createdAt])`.
- Extend `UserSettings` with `rewardSystemEnabled (Bool, default false)`, `defaultTaskPoints (Int, default 10)`, `showPointsOnCompletion (Bool, default true)`.

## Acceptance

- Migration created via `pnpm db:migrate`.
- Prisma generated types regenerated.
- Unit test for the upsert path (award flow uses `prisma.$transaction` to update total + insert transaction atomically).

## Plan reference

`.claude/plans/reward-point-system.md` (section "Database Schema Updates").

## Track

Days 7–8 (Rewards track, parallel to Tasks). Blocks the Rewards API and UI issues below.

BenSeymourODB:
Note from a triage pass on 2026-05-02: every acceptance criterion here is already satisfied in code.

- `RewardPoints`, `PointTransaction`, and the `UserSettings` flags (`rewardSystemEnabled`, `defaultTaskPoints`, `showPointsOnCompletion`) live in `prisma/schema.prisma` and `prisma/migrations/0001_initial/migration.sql`.
- A `ProfileRewardPoints` model was added alongside the multi-profile work; `PointTransaction` is keyed on `profileId` (not `userId`) to match. The `userId`-keyed shape from the original spec was superseded — the per-profile shape is the canonical one going forward.
- The atomic upsert + transaction insert pattern is implemented in `src/lib/services/reward-points.ts` and covered by `src/lib/services/__tests__/reward-points.test.ts` (six cases including `$transaction` invocation count and error propagation).

Suggesting this can be closed. The remaining rewards work proceeds via #172 (API + context — picked up on `claude/loving-faraday-0ho73` today) and #173 (UI surface).

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #198, opened 2026-05-02
feat(rewards): API routes + PointsContext (#172)

### LINKED ISSUE #172: feat(rewards): add API routes (GET /api/points, POST /api/points/award) and PointsContext
Add the API + client-state surface for the reward system.

## Routes

- `GET /api/points` — returns `{ totalPoints, enabled }`. If user settings have rewards disabled, returns `{ totalPoints: 0, enabled: false }`.
- `POST /api/points/award` — body `{ points, reason, taskId?, taskTitle? }`. Validates positive points and that rewards are enabled (403 otherwise). Uses `prisma.$transaction` to update total and insert a `PointTransaction` row atomically.

## Client

- `src/components/rewards/points-context.tsx` — `totalPoints`, `isEnabled`, `awardPoints(points, reason, metadata)`, `refreshPoints()`. Loaded on mount; updates on each award.

## Acceptance

- `awardPoints` is idempotent per-task (TBD: enforced by a unique index on `(userId, taskId, reason='task_completed')`, or by client-side guard).
- Logger events: `PointsAwarded`, `RewardSystemToggled`.
- Integration tests cover the transaction path and the disabled-rewards path.

## Blocked by

- Rewards Prisma schema issue.

## Plan reference

`.claude/plans/reward-point-system.md` (sections "API Routes", "Points Context").

## Track

Days 7–8 (Rewards track).

BenSeymourODB:
Closes #172

Adds the API + client-state surface for the reward-points system. The Prisma schema and the atomic upsert/insert pattern from #171 were already in place; this PR builds the thin extensions and the surfaces that hang off them.

## Phase 1 — Service + idempotency

- Extends `src/lib/services/reward-points.ts` with `recordPointAward({ profileId, points, reason, taskId?, taskTitle?, awardedBy?, note? })`. The existing `awardPoints` admin path now calls through it for backward compatibility.
- New migration `0005_pointtransaction_unique_task_award` adds a unique index on `PointTransaction(profileId, taskId, reason)`. Postgres treats unique-index NULLs as distinct, so manual / bonus / streak awards (which leave `taskId` NULL) are unaffected; the constraint only fires when the same task tries to credit the same profile a second time.
- `recordPointAward` catches the resulting `P2002` and returns `alreadyAwarded: true` together with the profile's current total — so callers stay in sync without an extra round trip.
- Six new unit tests cover the task-completion path, manual path, both no-record and existing-record duplicates, and non-P2002 propagation.

## Phase 2 — `GET /api/points` and `POST /api/points/award`

- `GET /api/points?profileId=…` returns `{ totalPoints, enabled }`. `enabled` is the account-level `UserSettings.rewardSystemEnabled` flag; when false the route always reports `totalPoints: 0` so disabled accounts never see stale totals leak through.
- `POST /api/points/award` validates positive integer points, a recognised reason, profile ownership, and that rewards are enabled, then delegates to `recordPointAward`. Idempotent task-completion duplicates surface as `{ alreadyAwarded: true, newTotal }` rather than 5xx errors.
- 19 route tests cover unauthenticated / missing-param / invalid-input / cross-account / disabled / happy / duplicate / failure paths.

## Phase 3 — `PointsContext` / `usePoints`

- Provider takes `profileId` as a prop (decoupled from `ProfileProvider`) so whoever mounts it forwards the active profile id and tests don't need a profile context. Public values are derived from stored fetch state so swapping to a null profile zeros the surface without a synchronous `setState` in the effect.
- Exposes `totalPoints`, `isEnabled`, `profileId`, `awardPoints(points, reason, metadata?)`, and `refreshPoints()`. `awardPoints` does NOT locally increment when the API reports a duplicate.
- 11 component tests cover the hook contract end-to-end (error-outside-provider, null-profile no-fetch, success/disabled/failure GET, POST happy path, idempotent duplicate, POST failure, no-profile reject, refreshPoints, re-fetch on profileId change).

## Out of scope (handled by other issues)

- Mounting `PointsProvider` in the root layout and wiring it to the active profile from `ProfileContext` — #173.
- `PointsBadge`, `PointsAnimation`, settings UI, and the `TaskItem` integration — #173.

## Notes for reviewers

- #171 is effectively done in code already (schema, migration, atomic upsert + transaction insert, six existing unit tests). Left a comment on it recommending closure.
- The reward enablement check still lives on `UserSettings` (per-account), while the points themselves are per-profile via `ProfileRewardPoints`. This is the design that the rest of the codebase has already moved to since the multi-profile work landed.
- All checks green locally: `pnpm lint:fix && pnpm format:fix && pnpm check-types && pnpm test:run` — 1470 tests, no errors, no warnings I introduced.

### DIFF
--- a/prisma/migrations/0005_pointtransaction_unique_task_award/migration.sql
+++ b/prisma/migrations/0005_pointtransaction_unique_task_award/migration.sql
@@ -1,0 +1,7 @@
+-- Idempotent task-completion awards: prevent the same task from
+-- crediting the same profile more than once. PostgreSQL unique
+-- indexes treat NULLs as distinct by default, so this constraint
+-- does not affect manual / bonus / streak awards (which leave
+-- taskId NULL).
+CREATE UNIQUE INDEX "PointTransaction_profileId_taskId_reason_key"
+  ON "PointTransaction" ("profileId", "taskId", "reason");
--- a/src/app/api/points/award/route.ts
+++ b/src/app/api/points/award/route.ts
@@ -1,0 +1,143 @@
+/**
+ * POST /api/points/award - awards points to a profile.
+ *
+ * Body:
+ *   {
+ *     profileId: string;
+ *     points: number;            // positive integer
+ *     reason: PointAwardReason;  // e.g. "task_completed"
+ *     taskId?: string;
+ *     taskTitle?: string;
+ *   }
+ *
+ * Response:
+ *   { success: true, newTotal: number, alreadyAwarded: boolean }
+ *
+ * Reward enablement is checked at the account level
+ * (`UserSettings.rewardSystemEnabled`). Points themselves are stored
+ * per profile. The `alreadyAwarded` flag surfaces idempotency hits
+ * driven by the `(profileId, taskId, reason)` unique index — a
+ * task can only credit a profile once.
+ */
+import { getSession } from "@/lib/auth";
+import { prisma } from "@/lib/db";
+import { logger } from "@/lib/logger";
+import {
+  type PointAwardReason,
+  recordPointAward,
+} from "@/lib/services/reward-points";
+import { NextRequest, NextResponse } from "next/server";
+
+const VALID_REASONS: ReadonlySet<PointAwardReason> = new Set([
+  "task_completed",
+  "manual",
+  "bonus",
+  "streak",
+  "goal",
+]);
+
+interface AwardRequestBody {
+  profileId?: string;
+  points?: number;
+  reason?: string;
+  taskId?: string;
+  taskTitle?: string;
+}
+
+function isValidReason(value: string): value is PointAwardReason {
+  return VALID_REASONS.has(value as PointAwardReason);
+}
+
+export async function POST(request: NextRequest) {
+  try {
+    const session = await getSession();
+
+    if (!session?.user?.id) {
+      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+    }
+
+    const body = (await request.json()) as AwardRequestBody;
+    const { profileId, points, reason, taskId, taskTitle } = body;
+
+    if (!profileId) {
+      return NextResponse.json(
+        { error: "profileId is required" },
+        { status: 400 }
+      );
+    }
+
+    if (
+      typeof points !== "number" ||
+      !Number.isInteger(points) ||
+      points <= 0
+    ) {
+      return NextResponse.json(
+        { error: "Invalid points value" },
+        { status: 400 }
+      );
+    }
+
+    if (typeof reason !== "string" || !isValidReason(reason)) {
+      return NextResponse.json({ error: "Invalid reason" }, { status: 400 });
+    }
+
+    const profile = await prisma.profile.findFirst({
+      where: {
+        id: profileId,
+        userId: session.user.id,
+        isActive: true,
+      },
+      select: { id: true },
+    });
+
+    if (!profile) {
+      return NextResponse.json({ error: "Profile not found" }, { status: 404 });
+    }
+
+    const settings = await prisma.userSettings.findUnique({
+      where: { userId: session.user.id },
+      select: { rewardSystemEnabled: true },
+    });
+
+    if (!settings?.rewardSystemEnabled) {
+      return NextResponse.json(
+        { error: "Reward system not enabled" },
+        { status: 403 }
+      );
+    }
+
+    const result = await recordPointAward({
+      profileId,
+      points,
+      reason,
+      taskId,
+      taskTitle,
+    });
+
+    logger.event("PointsAwarded", {
+      userId: session.user.id,
+      profileId,
+      points,
+      reason,
+      newTotal: result.totalPoints,
+      alreadyAwarded: result.alreadyAwarded,
+      ...(taskId ? { taskId } : {}),
+    });
+
+    return NextResponse.json({
+      success: true,
+      newTotal: result.totalPoints,
+      alreadyAwarded: result.alreadyAwarded,
+    });
+  } catch (error) {
+    logger.error(error as Error, {
+      endpoint: "/api/points/award",
+      method: "POST",
+    });
+
+    return NextResponse.json(
+      { error: "Failed to award points" },
+      { status: 500 }
+    );
+  }
+}
--- a/src/app/api/points/award/route.ts
+++ b/src/app/api/points/award/route.ts
@@ -4,9 +4,9 @@
  * Body:
  *   {
  *     profileId: string;
- *     points: number;            // positive integer
+ *     points: number;            // positive integer, max MAX_POINTS_PER_AWARD
  *     reason: PointAwardReason;  // e.g. "task_completed"
- *     taskId?: string;
+ *     taskId?: string;           // required when reason === "task_completed"
  *     taskTitle?: string;
  *   }
  *
@@ -17,7 +17,10 @@
  * (`UserSettings.rewardSystemEnabled`). Points themselves are stored
  * per profile. The `alreadyAwarded` flag surfaces idempotency hits
  * driven by the `(profileId, taskId, reason)` unique index — a
- * task can only credit a profile once.
+ * task can only credit a profile once. The index is partial in
+ * effect (Postgres treats NULL as distinct), so the route requires
+ * `taskId` for `task_completed` awards to keep the idempotency
+ * guarantee meaningful.
  */
 import { getSession } from "@/lib/auth";
 import { prisma } from "@/lib/db";
@@ -35,6 +38,12 @@ const VALID_REASONS: ReadonlySet<PointAwardReason> = new Set([
   "streak",
   "goal",
 ]);
+
+// Cap a single award well below 32-bit Int range so an accumulated
+// totalPoints can't reasonably overflow Postgres INTEGER. Anything
+// realistic for task / bonus / streak awards stays comfortably under
+// this; admin manual awards above this should hit a different path.
+const MAX_POINTS_PER_AWARD = 10_000;
 
 interface AwardRequestBody {
   profileId?: string;
@@ -69,7 +78,8 @@ export async function POST(request: NextRequest) {
     if (
       typeof points !== "number" ||
       !Number.isInteger(points) ||
-      points <= 0
+      points <= 0 ||
+      points > MAX_POINTS_PER_AWARD
     ) {
       return NextResponse.json(
         { error: "Invalid points value" },
@@ -79,6 +89,17 @@ export async function POST(request: NextRequest) {
 
     if (typeof reason !== "string" || !isValidReason(reason)) {
       return NextResponse.json({ error: "Invalid reason" }, { status: 400 });
+    }
+
+    // Without taskId the unique (profileId, taskId, reason) index can't
+    // dedupe — Postgres treats NULL as distinct, so a caller could
+    // re-award the same "task" indefinitely. Force taskId on the
+    // automated task path to keep the idempotency guarantee.
+    if (reason === "task_completed" && !taskId) {
+      return NextResponse.json(
+        { error: "taskId is required for task_completed awards" },
+        { status: 400 }
+      );
     }
 
     const profile = await prisma.profile.findFirst({
--- a/src/components/rewards/points-context.tsx
+++ b/src/components/rewards/points-context.tsx
@@ -16,21 +16,18 @@
  * state (zero points, isEnabled=false) and never hits the network.
  */
 import { logger } from "@/lib/logger";
+import type { PointAwardReason } from "@/lib/services/reward-points";
 import {
   type ReactNode,
   createContext,
   useCallback,
   useContext,
   useEffect,
+  useRef,
   useState,
 } from "react";
 
-export type PointAwardReason =
-  | "task_completed"
-  | "manual"
-  | "bonus"
-  | "streak"
-  | "goal";
+export type { PointAwardReason };
 
 export interface AwardPointsMetadata {
   taskId?: string;
@@ -70,9 +67,61 @@ export function PointsProvider({ profileId, children }: PointsProviderProps) {
   const [fetchedTotal, setFetchedTotal] = useState(0);
   const [fetchedEnabled, setFetchedEnabled] = useState(false);
 
-  const fetchPoints = useCallback(async (id: string) => {
+  // Track the currently-mounted profile id so in-flight requests
+  // started under a different profile don't write their result into
+  // the new profile's displayed state when they resolve.
+  const activeProfileIdRef = useRef<string | null>(profileId);
+  useEffect(() => {
+    activeProfileIdRef.current = profileId;
+  }, [profileId]);
+
+  useEffect(() => {
+    if (!profileId) {
+      return;
+    }
+    const controller = new AbortController();
+    const run = async () => {
+      try {
+        const response = await fetch(
+          `/api/points?profileId=${encodeURIComponent(profileId)}`,
+          { signal: controller.signal }
+        );
+        if (controller.signal.aborted) return;
+        if (!response.ok) {
+          setFetchedTotal(0);
+          setFetchedEnabled(false);
+          return;
+        }
+        const data = (await response.json()) as {
+          totalPoints: number;
+          enabled: boolean;
+        };
+        if (controller.signal.aborted) return;
+        setFetchedTotal(data.totalPoints);
+        setFetchedEnabled(data.enabled);
+      } catch (error) {
+        if ((error as Error).name === "AbortError") return;
+        logger.error(error as Error, { context: "FetchPointsFailed" });
+        setFetchedTotal(0);
+        setFetchedEnabled(false);
+      }
+    };
+    run();
+    return () => controller.abort();
+  }, [profileId]);
+
+  const totalPoints = profileId ? fetchedTotal : 0;
+  const isEnabled = profileId ? fetchedEnabled : false;
+
+  const refreshPoints = useCallback(async () => {
+    if (!profileId) {
+      return;
+    }
     try {
-      const response = await fetch(`/api/points?profileId=${id}`);
+      const response = await fetch(
+        `/api/points?profileId=${encodeURIComponent(profileId)}`
+      );
+      if (activeProfileIdRef.current !== profileId) return;
       if (!response.ok) {
         setFetchedTotal(0);
         setFetchedEnabled(false);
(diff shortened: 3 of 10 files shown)


src/lib/services/__tests__/reward-points.test.ts
```diff
@@ -127,3 +141,122 @@ describe("awardPoints", () => {
     );
```
BenSeymourODB, line 141: **Pre-existing `awardPoints` tests will likely break (outside diff, but caused by this PR).**

`awardPoints` now delegates to `recordPointAward`, which always writes every field of `PointAwardInput` into `tx.pointTransaction.create`, including `taskId: input.taskId` and `taskTitle: input.taskTitle`. Because `awardPoints` doesn't pass `taskId` or `taskTitle` to `recordPointAward`, those resolve to `undefined` at the call site — but they are **explicitly present** in the resulting object (i.e. `{ ..., taskId: undefined, taskTitle: undefined, ... }`).

Vitest's `toHaveBeenCalledWith` uses deep equality. `{ a: 1, b: undefined }` has `Object.keys` of `['a','b']`, while `{ a: 1 }` has `['a']`; they are **not** equal. The two assertions under the original `describe("awardPoints")` block (the "records a point transaction" and "includes the note" cases — both use an exact `expect(create).toHaveBeenCalledWith(...)` without the `taskId`/`taskTitle` fields) will therefore fail.

Fix options:
1. Add `taskId: undefined, taskTitle: undefined` to both expected `data` objects in the `awardPoints` tests.
2. Switch those two assertions to `expect.objectContaining(...)` so they assert only the fields `awardPoints` is responsible for.

Please run `pnpm test --reporter=verbose src/lib/services/__tests__/reward-points.test.ts` to confirm before merge.

---

src/app/api/points/award/route.ts
```diff
@@ -0,0 +1,143 @@
...
+      !Number.isInteger(points) ||
+      points <= 0
+    ) {
+      return NextResponse.json(
+        { error: "Invalid points value" },
+        { status: 400 }
+      );
+    }
+
+    if (typeof reason !== "string" || !isValidReason(reason)) {
```
BenSeymourODB, line 90: **Missing cross-field validation: `task_completed` without `taskId` silently bypasses idempotency.**

The unique index on `(profileId, taskId, reason)` only deduplicates rows where `taskId IS NOT NULL`. Postgres treats each NULL as distinct, so `reason = "task_completed"` with no `taskId` can be inserted as many times as callers like — the P2002 path in `recordPointAward` is never triggered.

Nothing in the route currently enforces that `taskId` must be present when `reason === "task_completed"`. A caller that accidentally (or maliciously) omits `taskId` will have their points incremented on every call with no guard.

Suggested addition after the `reason` validation block:

```ts
if (reason === "task_completed" && !taskId) {
  return NextResponse.json(
    { error: "taskId is required for task_completed awards" },
    { status: 400 }
  );
}
```

A matching test case for this path should also be added to `award/__tests__/route.test.ts`.

---

src/app/api/points/award/route.ts
```diff
@@ -0,0 +1,143 @@
...
+        { error: "profileId is required" },
+        { status: 400 }
+      );
+    }
+
+    if (
+      typeof points !== "number" ||
+      !Number.isInteger(points) ||
+      points <= 0
+    ) {
```
BenSeymourODB, line 83: **No upper bound on `points`.**

`points <= 0` and `!Number.isInteger(points)` are guarded, but there is no cap. A caller can send `points: 2147483647` (or any arbitrarily large integer) and the value goes straight to the Prisma `increment`. The Prisma schema declares `points Int` which maps to a 32-bit Postgres `INTEGER`; an increment that causes `totalPoints` to overflow `2147483647` will raise a Prisma/Postgres runtime error that bubbles up as an unhandled 500.

Consider adding a reasonable ceiling (e.g. `points > 10_000`) and returning 400, and adding a test for it. At minimum, document the intended maximum in the JSDoc so future callers know the constraint.

---

src/components/rewards/points-context.tsx
```diff
@@ -0,0 +1,183 @@
...
+import {
+  type ReactNode,
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from "react";
+
+export type PointAwardReason =
```
BenSeymourODB: **`PointAwardReason` is re-declared here instead of imported from the service.**

`src/lib/services/reward-points.ts` already exports `PointAwardReason` with exactly the same five values. Re-declaring it here creates a maintenance hazard: if a new reason is added to the service (e.g. `"milestone"`), the context type won't include it and callers passing the new reason will get a TypeScript error on the client side.

`reward-points.ts` has no `"use server"` directive and no `server-only` package import, so its types are safe to import in a client component using `import type`:

```ts
import type { PointAwardReason } from "@/lib/services/reward-points";
```

Types are erased at build time, so there is zero runtime impact from the Prisma import in that module.

---

src/components/rewards/points-context.tsx
```diff
@@ -0,0 +1,183 @@
...
+      setFetchedTotal(data.totalPoints);
+      setFetchedEnabled(data.enabled);
+    } catch (error) {
+      logger.error(error as Error, { context: "FetchPointsFailed" });
+      setFetchedTotal(0);
+      setFetchedEnabled(false);
+    }
+  }, []);
+
+  useEffect(() => {
```
BenSeymourODB: **No fetch-cancellation on `profileId` change — stale response can overwrite the new profile's total.**

When `profileId` changes (e.g. the user switches from profile A to profile B), the `useEffect` fires a new `fetchPoints("B")` call. But the in-flight `fetchPoints("A")` request is not cancelled. If the A response arrives *after* the B response, it calls `setFetchedTotal(A_total)` and `setFetchedEnabled(A_enabled)` against the state that is now being displayed for profile B.

Fix with an `AbortController`:

```ts
useEffect(() => {
  if (!profileId) return;
  const controller = new AbortController();
  const run = async () => {
    try {
      const response = await fetch(
        `/api/points?profileId=${encodeURIComponent(profileId)}`,
        { signal: controller.signal }
      );
      // ... rest of fetchPoints logic inline or accept the signal
    } catch (err) {
      if ((err as Error).name === "AbortError") return;
      // handle real errors
    }
  };
  run();
  return () => controller.abort();
}, [profileId]);
```

Alternatively refactor `fetchPoints` to accept a signal and pass it through. The same concern applies to `refreshPoints`, though rapid concurrent calls there are less likely in practice.

---

src/components/rewards/points-context.tsx
```diff
@@ -0,0 +1,183 @@
...
+  // `totalPoints` / `isEnabled` are derived from it so that swapping
+  // to a null profile zeros the public values without needing a
+  // synchronous setState in the effect (which would trigger
+  // cascading renders).
+  const [fetchedTotal, setFetchedTotal] = useState(0);
+  const [fetchedEnabled, setFetchedEnabled] = useState(false);
+
+  const fetchPoints = useCallback(async (id: string) => {
+    try {
+      const response = await fetch(`/api/points?profileId=${id}`);
```
BenSeymourODB: **`profileId` should be URL-encoded.**

`id` is a cuid and won't contain special characters in practice today, but leaning on that assumption is fragile. Use `encodeURIComponent(id)` to be safe and to make the intent explicit:

```ts
const response = await fetch(`/api/points?profileId=${encodeURIComponent(id)}`);
```

---

src/lib/services/reward-points.ts
```diff
@@ -12,49 +12,122 @@ export interface AwardPointsResult {
...
+      });
+
+      return updated;
     });
 
-    return { totalPoints: rewardPoints.totalPoints };
-  });
+    return { totalPoints: rewardPoints.totalPoints, alreadyAwarded: false };
+  } catch (error) {
+    if (isUniqueConstraintError(error)) {
```
BenSeymourODB, line 112: **P2002 detection by duck-typing `error.code` is pragmatically fine but worth a note.**

The `isUniqueConstraintError` helper checks `"code" in error && error.code === "P2002"` rather than `error instanceof Prisma.PrismaClientKnownRequestError`. The instance check is the canonical Prisma way and also catches the `meta.target` array so you can confirm which constraint fired.

The duck-typing approach works in production and in tests (the test helper manually sets `.code = "P2002"`), and avoids importing the Prisma namespace just for the type guard. That trade-off is acceptable here. One edge case to be aware of: if any non-Prisma library in the call stack throws an object with a `code` of `"P2002"` (unlikely but possible in theory), it would be swallowed as a duplicate-award no-op instead of propagating. Consider adding a comment explaining the deliberate choice, e.g.:

```ts
// We duck-type rather than use instanceof PrismaClientKnownRequestError
// to avoid importing the Prisma namespace here. P2002 from a non-Prisma
// source is vanishingly unlikely in this call path.
```

---

prisma/migrations/0005_pointtransaction_unique_task_award/migration.sql
```diff
@@ -0,0 +1,7 @@
+-- Idempotent task-completion awards: prevent the same task from
+-- crediting the same profile more than once. PostgreSQL unique
+-- indexes treat NULLs as distinct by default, so this constraint
+-- does not affect manual / bonus / streak awards (which leave
+-- taskId NULL).
+CREATE UNIQUE INDEX "PointTransaction_profileId_taskId_reason_key"
```
BenSeymourODB, line 6: **`CREATE UNIQUE INDEX` without `CONCURRENTLY` takes a full table lock.**

On a live database with a large `PointTransaction` table, this will hold an `AccessShareExclusiveLock` for the duration of the index build, blocking all concurrent writes to that table.

For a pre-launch project with no production traffic this is fine. But once the app is live, this migration should use `CREATE UNIQUE INDEX CONCURRENTLY` instead, which builds the index without locking writes (at the cost of requiring manual transaction wrapping, since `CONCURRENTLY` cannot run inside a transaction block). Worth flagging for the deployment docs or a follow-up issue so the pattern is not copied into future migrations on hot tables without that consideration.

---

src/components/rewards/points-context.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const isEnabled = profileId ? fetchedEnabled : false;
+
+  const refreshPoints = useCallback(async () => {
+    if (!profileId) {
+      return;
+    }
+    await fetchPoints(profileId);
+  }, [profileId, fetchPoints]);
+
+  const awardPoints = useCallback(
```
BenSeymourODB, line 145: **`awardPoints` callback closure can set stale profile's total on a concurrent `profileId` change.**

`awardPoints` is memoized on `[profileId]` so a new closure is created when `profileId` changes. But if the POST request fired from the *old* closure is still in flight when `profileId` changes to a new value, the old closure's `setFetchedTotal(data.newTotal)` call will execute against the *new* profile's displayed state — overwriting its correct total with the old profile's post-award total.

This is the same family of issue as the `useEffect` fetch race (noted separately). The cleanest mitigation at the `awardPoints` level is to check that the `profileId` in the closure still matches the current `profileId` before calling `setFetchedTotal`:

```ts
// inside the closure, after the response resolves:
setFetchedTotal((prev) => {
  // only apply if the profileId this closure captured is still active
  return currentProfileIdRef.current === profileId ? data.newTotal : prev;
});
```

This requires a `useRef` to track the current `profileId`. A simpler alternative is to always follow a successful `awardPoints` POST with a `refreshPoints()` call (which uses the *current* `profileId`), letting the GET re-sync the total correctly. That's slightly more network traffic but eliminates the stale-closure window entirely.

---

BenSeymourODB (commented):
Overall the structure is solid — auth guards, cross-account ownership checks, idempotency surfacing, and test coverage are all well thought out. Three issues need fixing before merge (one likely breaks existing tests today), plus a few smaller things called out inline.

**Must-fix**

1. **Existing `awardPoints` tests will break** (see comment on `reward-points.test.ts` line 141). The refactor routes `awardPoints` through `recordPointAward`, which always writes `taskId: undefined` and `taskTitle: undefined` explicitly into the `create` payload. The two pre-existing exact `toHaveBeenCalledWith` assertions under `describe("awardPoints")` no longer match the actual call shape.

2. **`task_completed` without `taskId` silently bypasses idempotency** (see comment on `award/route.ts` line 80). The unique index only fires when `taskId IS NOT NULL`; a caller omitting `taskId` with `reason = "task_completed"` can award points unboundedly. Add a cross-field validation guard and a matching test.

3. **No upper bound on `points`** (see comment on `award/route.ts` line 73). An arbitrarily large value can overflow the 32-bit `Int` column at the database level, producing an unhandled 500.

**Should-fix**

- Stale-fetch race condition on `profileId` change in `PointsProvider` — no `AbortController` cleanup in `useEffect`, and the `awardPoints` closure can write the old profile's total into the new profile's state (two separate inline comments).
- `PointAwardReason` re-declared in `points-context.tsx` instead of `import type`-ed from the service — divergence hazard when reasons are extended.
- `profileId` not URL-encoded in the `fetch` call.

**Nit / informational**

- The migration uses `CREATE UNIQUE INDEX` without `CONCURRENTLY` — fine for pre-launch, but worth documenting for when the table has live traffic.
- The duck-typed P2002 guard works fine; left a comment suggesting a one-line explanation of the deliberate choice.

---

src/lib/services/__tests__/reward-points.test.ts
BenSeymourODB: Verified: the existing `awardPoints` tests all still pass. Vitest 4.1.2's `toHaveBeenCalledWith` (via `equals` from `@vitest/utils`) treats `undefined`-valued keys as equivalent to absent keys for object comparison — `{ a: 1, b: undefined }` and `{ a: 1 }` match. Confirmed locally with `pnpm test:run --reporter=verbose src/lib/services/__tests__/reward-points.test.ts`: all 11 tests green, including the two original `awardPoints` cases that don't list `taskId`/`taskTitle` in the expected `data`.

Leaving the existing assertions as-is rather than re-writing them to `expect.objectContaining(...)`. Happy to switch to that or to make `taskId`/`taskTitle` explicit if you'd prefer the assertions to read more literally.

---

prisma/migrations/0005_pointtransaction_unique_task_award/migration.sql
BenSeymourODB: Agreed — pre-launch this is fine. `PointTransaction` is empty in production today, so the `CREATE UNIQUE INDEX` here will return immediately with no measurable lock window. Once the table grows large and the app is live, future migrations that touch it should use `CREATE UNIQUE INDEX CONCURRENTLY` (and live outside a transaction block, since `CONCURRENTLY` can't run inside one).

Not changing this migration since it has to land before there's any data to worry about, and Prisma's migration runner wraps each migration in a transaction — so making this one `CONCURRENTLY` would actually fail. Worth capturing the convention in [`docs/database.md`](../blob/main/docs/database.md) as a follow-up so the pattern doesn't get copied onto a hot table later. Filing a tracking issue out of band.

---

src/app/api/points/award/route.ts
BenSeymourODB: Fixed in `b337081`. Added the cross-field guard exactly as suggested:

```ts
if (reason === "task_completed" && !taskId) {
  return NextResponse.json(
    { error: "taskId is required for task_completed awards" },
    { status: 400 }
  );
}
```

Plus a matching route test (`returns 400 when reason is task_completed but taskId is missing`) that asserts the 400 fires before `recordPointAward` is reached. The pre-existing 403/404/500 tests under this `describe` were updated to use `reason: "bonus"` so they continue to exercise the path under test instead of tripping this new guard.

---

src/app/api/points/award/route.ts
BenSeymourODB: Fixed in `b337081`. Added a `MAX_POINTS_PER_AWARD = 10_000` ceiling and folded it into the existing 400 path:

```ts
if (
  typeof points !== "number" ||
  !Number.isInteger(points) ||
  points <= 0 ||
  points > MAX_POINTS_PER_AWARD
) {
  return NextResponse.json({ error: "Invalid points value" }, { status: 400 });
}
```

The constant is documented in the route's JSDoc and covered by a new `returns 400 when points exceeds MAX_POINTS_PER_AWARD` test (sends `points: 10_001`). 10K per single award keeps a reasonable margin against the 32-bit Postgres `INTEGER` ceiling even after many awards accumulate.

---

src/components/rewards/points-context.tsx
BenSeymourODB: Fixed in `b337081`. Replaced the local declaration with a `import type` from the service and re-exported it so existing consumers of the context's `PointAwardReason` keep working:

```ts
import type { PointAwardReason } from "@/lib/services/reward-points";
// ...
export type { PointAwardReason };
```

Single source of truth on the service, no runtime impact, no Prisma leakage into the client bundle (types only).

---

src/components/rewards/points-context.tsx
BenSeymourODB: Fixed in `b337081`. Inlined the GET into the effect with an `AbortController`, aborted on cleanup, and added an `aborted` guard before each `setState`:

```ts
useEffect(() => {
  if (!profileId) return;
  const controller = new AbortController();
  const run = async () => {
    try {
      const response = await fetch(
        `/api/points?profileId=${encodeURIComponent(profileId)}`,
        { signal: controller.signal }
      );
      if (controller.signal.aborted) return;
      // ...
    } catch (error) {
      if ((error as Error).name === "AbortError") return;
      // ...
    }
  };
  run();
  return () => controller.abort();
}, [profileId]);
```

The shared `fetchPoints` `useCallback` was removed since only `refreshPoints` still needs that path; refresh now uses the active-profile ref guard described in [my reply to the awardPoints race comment](https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/198#discussion_r3176661016) (cancellation isn't useful there because `refreshPoints` is invoked imperatively by callers, not by a profile change). The existing `re-fetches when profileId changes` test was updated to assert the second fetch goes out with an `AbortSignal`.

---

src/components/rewards/points-context.tsx
BenSeymourODB: Fixed in `b337081`. Both fetch paths now use `encodeURIComponent(profileId)`:

```ts
const response = await fetch(
  `/api/points?profileId=${encodeURIComponent(profileId)}`,
  { signal: controller.signal }
);
```

Tests still match — `expect.objectContaining({ signal: ... })` covers the new second arg, and the encoded URL stays identical for cuid-shaped ids in the existing test fixtures.

---

src/lib/services/reward-points.ts
BenSeymourODB: Done in `b337081`. Added the explanatory comment on `isUniqueConstraintError`:

```ts
// We duck-type the Prisma P2002 code rather than `instanceof
// PrismaClientKnownRequestError` to avoid pulling the Prisma
// namespace into this module just for a type guard. P2002 from a
// non-Prisma source is vanishingly unlikely on this call path
// (the only awaited write is a Prisma transaction).
function isUniqueConstraintError(error: unknown): boolean { /* ... */ }
```

---

src/components/rewards/points-context.tsx
BenSeymourODB: Fixed in `b337081`. Took the `useRef` approach you described:

```ts
const activeProfileIdRef = useRef<string | null>(profileId);
useEffect(() => {
  activeProfileIdRef.current = profileId;
}, [profileId]);

// inside awardPoints, after the response resolves:
if (activeProfileIdRef.current === profileId) {
  setFetchedTotal(data.newTotal);
}
```

The ref is assigned via a sync effect (the React 19 lint rule disallows ref writes during render). Same guard added to the `refreshPoints` path. New regression test (`does not write a stale awardPoints result into the new profile's state`) reproduces the race: it fires an `awardPoints` POST under `profile-1`, switches to `profile-2` while the POST is still in flight, then resolves the in-flight POST with `newTotal: 999` — the displayed total stays at profile-2's `200` rather than getting clobbered.

---

PR #199, opened 2026-05-02
feat(calendar): wire event creation to Google Calendar (#116)

BenSeymourODB:
Closes #116

Wires the existing `EventCreateDialog` UI to a real Google Calendar write path. Mirrors the auth + error shape established by the `[id]` DELETE route from #115 so a single re-auth contract covers the whole calendar CRUD cluster.

## Phase 1 — Server (`feat(calendar): add POST /api/calendar/events`)

- New `POST /api/calendar/events` handler colocated with the existing `GET` so the calendar event route stays one file.
- Validates: non-empty title, valid ISO `startDate`/`endDate`, `end > start`, `color ∈ TEventColor`, malformed JSON.
- Defaults `calendarId` to `"primary"` when the caller omits it; the field is on the wire from day one so a future calendar-picker UI can land without an API change.
- All-day events are emitted as `start.date` / `end.date` with Google's exclusive-end convention. A single all-day event on Apr 20 sends `end.date = "2026-04-21"`.
- Maps `TEventColor` → Google `colorId` (`blue→1`, `green→2`, `purple→3`, `red→4`, `yellow→5`, `orange→6`) — the inverse of the existing `calendar-transform.ts` map.
- Returns 201 with the normalised event so the client can swap its optimistic placeholder for the canonical id/etag.
- Errors: Google 401 → `requiresReauth: true`, 403 → permission, 404 → not-found, anything else → 502 so callers can distinguish from local 5xx.

17 new vitest cases in `src/app/api/calendar/events/__tests__/route.test.ts` cover validation, shape, color mapping, all-day formatting, multi-day all-day, error pass-through, and the unexpected-error path.

## Phase 2 — Client (`feat(calendar): wire AddEventButton through createEvent`)

- `CalendarProvider.createEvent(optimistic, input): Promise<IEvent>` — inserts the optimistic row immediately, swaps it for the canonical Google event on success (id, etag from the server), and removes only that row on failure so a concurrent `refreshEvents()` can write to the rest of the list without being clobbered by rollback. The existing `addEvent` stays in place for callers that still want a pure local insert.
- `useEventCreate` wraps `createEvent` with sonner toasts and a configurable default `calendarId`.
- `AddEventButton` now drives the flow through the hook and builds the optimistic `IEvent` locally; the placeholder `LOCAL_USER` is only ever shown for the brief optimistic window before reconciliation.
- `MockCalendarProvider.createEvent` — hermetic local insert so the existing `event-create.spec.ts` E2E keeps passing without fetch mocks.
- Sibling component tests gain `createEvent: vi.fn()` in their context mock so the type contract stays sound.

New tests: `useEventCreate` (6 cases — default calendar, override, success/error toasts), provider success reconciliation, provider 403 rollback, rewritten AddEventButton tests covering the new contract.

## Local verification

```
pnpm lint:fix && pnpm format:fix && pnpm check-types  # 0 errors
pnpm exec vitest run                                   # 1460 passed (99 files)
```

Playwright E2E runs in CI (browsers couldn't be fetched in this sandbox). The existing `event-create.spec.ts` exercises the new `createEvent` path end-to-end via `MockCalendarProvider`.

## Deferred (called out, not in this PR)

- **Multi-calendar selector UI** inside the dialog. The endpoint accepts `calendarId` on day one, so this lands as a follow-up against the dialog without an API change.
- **Local PostgreSQL write path** — the calendar-event-database plan is still upstream; this PR only writes to Google.
- **IndexedDB cache write** on success. The existing `eventCache` repopulates from `refreshEvents()`, which is invoked anyway after successful creation; an explicit cache-write follows once the cache becomes the source of truth (per `calendar-event-database.md`).


src/app/api/calendar/events/route.ts
```diff
@@ -216,3 +236,315 @@ export async function GET(request: NextRequest) {
...
+ * `exclusiveEnd` below avoids the off-by-one that `setHours(0,0,0,0)` +
+ * `getDate()+1` produces on a server in a different TZ to the client.
+ *
+ * Caveat: a client in a positive UTC offset (e.g., NZST = UTC+12) creating
+ * an all-day event for "Apr 20 local" sends UTC = Apr 19. The route still
+ * sees Apr 19 here. Fully fixing that needs a wire-format change so the
+ * dialog emits `YYYY-MM-DD` strings (or a `tzOffsetMinutes`) rather than
+ * encoding local-midnight as a UTC ISO. Tracked as follow-up.
+ */
+function formatUTCDate(d: Date): string {
```
BenSeymourODB, line 352: **All-day timezone bug: `formatLocalDate` called on a UTC Date**

The dialog's `parseDateOnly` builds a local-midnight `Date` and then calls `.toISOString()` before sending it to the API. On the wire `startDate` looks like `"2026-04-20T07:00:00.000Z"` (for a user in UTC-7). The server then does:

```ts
const start = new Date("2026-04-20T07:00:00.000Z"); // UTC Date
formatLocalDate(start);  // calls getFullYear/getMonth/getDate on *server* locale
```

On a server in UTC this is correct. On a server in a positive-offset timezone (UTC+2), `getDate()` may return `20` correctly; but for the end date the situation is worse: the dialog sends `"2026-04-20T22:59:59.999Z"` (Apr 20 23:59:59 UTC-1 → UTC). The server's `exclusiveEnd`:

```ts
exclusiveEnd.setHours(0, 0, 0, 0);   // midnight in *server* tz
exclusiveEnd.setDate(exclusiveEnd.getDate() + 1);
```

If the server is UTC+2, `new Date("2026-04-20T22:59:59.999Z")` is already `2026-04-21T00:59:59` local. `setHours(0,0,0,0)` → Apr 21 midnight local → `+1 day` → Apr 22 exclusive end. The event lands on Google as Apr 20–22 instead of Apr 20–21.

The fix is to parse the UTC date string into its UTC components (not local):

```ts
function formatUTCDate(d: Date): string {
  const pad = (n: number) => n.toString().padStart(2, "0");
  return `${d.getUTCFullYear()}-${pad(d.getUTCMonth() + 1)}-${pad(d.getUTCDate())}`;
}
```

And replace `setHours(0,0,0,0)` / `setDate(...)` with UTC equivalents:

```ts
const exclusiveEnd = new Date(event.end);
exclusiveEnd.setUTCHours(0, 0, 0, 0);
exclusiveEnd.setUTCDate(exclusiveEnd.getUTCDate() + 1);
insertBody.start.date = formatUTCDate(event.start);
insertBody.end.date = formatUTCDate(exclusiveEnd);
```

Also update the existing test — it constructs dates via `new Date(2026, 3, 20, ...)` (local midnight) which happens to pass in the CI timezone, but the test should use `new Date(Date.UTC(2026, 3, 20, 0, 0, 0))` to lock the input to a well-defined UTC moment and actually exercise the fix.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -215,6 +237,72 @@ export function CalendarProvider({
...
+          if (typeof body?.error === "string") message = body.error;
+        } catch {
+          // Some upstream proxies return non-JSON 502s; the generic message
+          // is sufficient for the toast.
+        }
+        throw new Error(message);
+      }
+
+      const body = (await response.json()) as { event: GoogleCalendarEvent };
+      // Reconcile through the in-memory colorMappings state — refreshEvents
```
BenSeymourODB, line 339: **`await loadColorMappings()` bypasses in-memory state and `await`s a sync function**

`loadColorMappings()` returns `CalendarColorMapping[]` synchronously (it reads `localStorage`). Awaiting a non-Promise wraps the value in `Promise.resolve()`, which is a no-op but misleads readers into thinking this is async I/O.

More importantly, `colorMappings` state is already in scope and stays in sync with `saveColorMappings` calls made inside `refreshEvents`. Calling `loadColorMappings()` here re-reads localStorage on every create, bypassing the in-memory copy and making this the one place in the provider that doesn't honour the cached state:

```ts
// Before (line 277)
const mappings = await loadColorMappings();

// After — use the state variable that refreshEvents already keeps current
const reconciled = transformGoogleEvent(body.event, colorMappings);
```

This also removes the async dependency that could theoretically interleave with a concurrent `setColorMappings` call in `refreshEvents`.

---

src/app/api/calendar/events/route.ts
```diff
@@ -216,3 +236,315 @@ export async function GET(request: NextRequest) {
...
+  | { ok: false; error: string };
+
+function isSupportedColor(value: unknown): value is TEventColor {
+  return (
+    typeof value === "string" && (SUPPORTED_COLORS as string[]).includes(value)
+  );
+}
+
+function validateCreateBody(body: unknown): ValidationResult {
+  if (!body || typeof body !== "object" || Array.isArray(body)) {
```
BenSeymourODB, line 277: **`validateCreateBody` accepts arrays**

`typeof [] === "object"` and `![] === false`, so an array body passes the current guard at line 277 and reaches `const raw = body as Partial<CreateEventBody>`. `raw.title` evaluates to `undefined`, which correctly returns "Title is required". The observable behavior is fine, but the cast `body as Partial<CreateEventBody>` applied to an array is logically unsound and could surprise future readers.

Add an `Array.isArray` guard:

```ts
if (!body || typeof body !== "object" || Array.isArray(body)) {
  return { ok: false, error: "Request body must be a JSON object" };
}
```

Consider adding a test case: `body: []` should return 400 with a clear message, not 400 with "Title is required".

---

BenSeymourODB (commented):
**Round 1 review (Sonnet, posted on the previous head). Resolution summary:**

**COMMENT 1** (UTC date math) — **Fixed in 57aada2.** Adopted the suggestion of UTC-component formatting; went one step further with a duration-based exclusive-end (`start + N * MS_PER_DAY` where N is the rounded inclusive day-count) so negative-offset clients are also correct. Tests now use `Date.UTC()` and a new case simulates a UTC-7 browser send. Positive-offset clients still need a wire-format fix tracked separately — noted in the route doc-comment.

**COMMENT 2** (cached colorMappings) — **Fixed in 57aada2.** Switched to the in-memory `colorMappings` state.

**COMMENT 3** (Array.isArray guard) — **Fixed in 57aada2.** Added the guard plus a dedicated test case (`rejects an array body with a 400`).

**GENERAL 1** (toast timing) — **Comment added in 57aada2.** Kept post-confirmation but spelled out the durability-vs-immediacy tradeoff in `useEventCreate` so future authors don't move the toast above the await without surfacing a pending state.

**GENERAL 2** (`MockCalendarProvider` `useMemo`) — **Fixed in 57aada2.** Replaced with a render-time derivation per CLAUDE.md.

Full suite: 1462 passed (99 files). Lint/format/typecheck clean.

---

src/app/api/calendar/events/route.ts
BenSeymourODB: Round 2 — fixed in 57aada2.

The simple substitution (UTC components + `setUTCHours/setUTCDate+1`) doesn't quite work for negative-offset clients: a UTC-7 client sending `"2026-04-21T06:59:59.999Z"` (local Apr 20 23:59:59.999) would be normalised to `"2026-04-21T00:00:00.000Z"`, +1 day → `"2026-04-22"`, off by one. Switched to a duration-based exclusive-end:

```ts
const inclusiveDays = Math.max(
  1,
  Math.round((event.end.getTime() - event.start.getTime()) / MS_PER_DAY)
);
const exclusiveEnd = new Date(event.start.getTime() + inclusiveDays * MS_PER_DAY);
insertBody.start.date = formatUTCDate(event.start);
insertBody.end.date = formatUTCDate(exclusiveEnd);
```

A UTC client and a UTC-7 client now both produce `start.date="2026-04-20"` / `end.date="2026-04-21"` for a single-day Apr-20 event. Tests use `Date.UTC(...)` so they're TZ-independent on any dev box, and a new case explicitly mocks the UTC-7 wire shape (`"2026-04-20T07:00:00.000Z"` / `"2026-04-21T06:59:59.999Z"`) to lock in the negative-offset path.

Caveat called out in the new doc-comment: a positive-offset client (e.g. NZST = UTC+12) sending Apr-20 local still maps to UTC Apr 19 on the wire, and `getUTCDate()` returns 19. Fully fixing that needs a wire-format change (dialog sending `YYYY-MM-DD` strings or a `tzOffsetMinutes`) — out of scope here, deferred.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Round 2 — fixed in 57aada2. `colorMappings` state is now passed directly to `transformGoogleEvent`; the `await loadColorMappings()` is gone. The import remains because the initial-state lazy-init (`useState(() => loadColorMappings())`) still uses it for first-render rehydration on mount.

---

src/app/api/calendar/events/route.ts
BenSeymourODB: Round 2 — fixed in 57aada2. Guard now reads `if (!body || typeof body !== "object" || Array.isArray(body))`. New test `"rejects an array body with a 400"` posts `body: ["not", "an", "object"]` and asserts a 400 with an `/object/i` error — so the cast to `Partial<CreateEventBody>` only ever runs on a real plain object.

---

PR #200, opened 2026-05-02
feat(tasks): wire Google Tasks to multi-profile assignments (#166)

### LINKED ISSUE #166: feat(tasks): wire Google Tasks to multi-profile assignments (unblocks Phase 4)

BenSeymourODB:
Closes #166. Unblocks Phase 4 of the multi-profile plan.

Wires the existing `TaskAssignment` table through to the rendered task list so a profile-scoped `TaskList` instance shows only the active profile's tasks (plus optional unassigned ones) and renders avatar chips for each assignee.

### Phase 1 β€" data plumbing (`6c8346d`)
- `getTaskAssignmentsByTaskIds(taskIds)` service: single Prisma query, dedup'd input, returns `Map<taskId, AssignmentSummary[]>`.
- `GET /api/tasks?…&includeAssignments=true` opts into embedding `assignments: AssignmentSummary[]` per task. Default response is unchanged β€" zero impact on existing callers.

### Phase 2 β€" hook + types (`8914171`)
- `TaskListConfig` gains `profileFilter: string | null` and `showUnassigned: boolean`.
- `TaskWithMeta` now carries `assignments` (defaulted to `[]` when the server omits the field).
- `useTasks` requests enrichment and applies the filter only when `profileFilter` is set, so the unfiltered "family" path keeps the smaller request shape.

### Phase 3 β€" UI integration (`aaaa0c6`)
- `TaskItem` renders a small avatar stack on the right side of the title for every assigned profile, wrapped in a `role="group"` with a single `aria-label` listing all assignee names.
- New `ProfileScopedTaskList` wrapper consumes `useProfile()` and forwards a derived config to `<TaskList>`. In `viewMode === "profile"` it scopes to `activeProfile.id` and surfaces unassigned tasks by default; in `family` mode the filter clears. Caller-supplied overrides on the config win, keeping `TaskList` itself decoupled from the profile context for testing.

## Triage notes

- *"Completing a task records the completion against the active profile"* β€" **already satisfied**. `POST /api/tasks/[taskId]/complete` calls `updateProfileStreak(profileId)` which writes `ProfileRewardPoints` (current/longest streak + lastActivityDate). A separate `TaskCompletion` table from the original spec would be redundant given the data already lives there.
- Schema stays as-is. The original spec mentioned widening `TaskAssignment` to `(googleTaskId, listId, userId)` but the current `(taskId, profileId)` shape (with `@@unique`/`@@index`) is sufficient for these acceptance criteria. Adding columns now would be churn without new behaviour.

## Test plan

Component-level coverage was prioritised over Playwright E2E because there is no consumer page for `TaskList` yet β€" the Tasks UI mounts in follow-up issues. Once a `/tasks` page lands, an E2E covering profile-switch β†' filtered list should follow.

- [x] **Unit** β€" `getTaskAssignmentsByTaskIds`: empty input, dedup, query shape, grouping β†' 4 cases
- [x] **Integration** β€" `/api/tasks` keeps default shape; `?includeAssignments=true` embeds per-task arrays; empty list short-circuits the DB; existing tests still pass β†' 4 new cases
- [x] **Hook** β€" `useTasks` honours `profileFilter` + `showUnassigned`, treats missing `assignments` as unassigned, includes shared tasks β†' 5 new cases
- [x] **Component** β€" `TaskItem` renders a `role="group"` of avatars only when there are assignees and announces the names; `ProfileScopedTaskList` derives the right config from a real `ProfileProvider` for both view modes β†' 8 new cases
- [x] Full suite passes: **100 files, 1467 tests**

https://claude.ai/code/session_01UCinmYdZRdx1tixCH3ni2Y


BenSeymourODB (commented):
Overall the shape of this PR is good β€" single-query batch lookup, clean Map-based grouping, backward-compatible opt-in flag, and the component decoupling via `ProfileScopedTaskList` is sensible. Found one real security issue that needs to be fixed before merge, one type-safety gap, one missing error-path test, and one double-fetch on mount. Comments inline below.

---

BenSeymourODB (commented):
Overall this is well-structured and the test coverage is solid. Three issues worth fixing before merge, one of which is a real correctness/security concern. The rest are minor type hygiene items.

**Issues by severity:**

1. **Cross-tenant data leak in `getTaskAssignmentsByTaskIds`** β€" the Prisma query has no `userId` guard. See inline comment on `task-assignments.ts`.
2. **`avatar: unknown` in the service layer breaks strict TS and propagates downstream** β€" see inline on `task-assignments.ts`.
3. **`useCallback` in `use-tasks.ts` violates React Compiler rules** β€" pre-existing, but the PR adds a new `useCallback` wrapping `fetchTasks` and expands the dependency array. Worth noting even though it predates this PR.
4. **`family` view-mode path is never tested in `ProfileScopedTaskList`** β€" the suite only exercises profile-mode and the no-active-profile fallback. See inline comment.

Everything else (filter logic, avatar chip rendering, the `??[]` fallback for missing assignments, the skip-DB-call-when-empty-list optimisation, the `Omit<TaskListProps, "config">` interface pattern) looks correct.

---

BenSeymourODB:
## Review comments (inline-style, per file)

---

### FILE: `src/lib/services/task-assignments.ts` β€" LINE 40 (the `prisma.taskAssignment.findMany` call)

**SECURITY β€" cross-tenant data leak**

```ts
const rows = await prisma.taskAssignment.findMany({
  where: { taskId: { in: uniqueIds } },
  ...
});
```

`TaskAssignment` stores Google Task IDs as plain strings with no user-scoped foreign key. Google Task IDs are not globally unique β€" two different Google accounts can legitimately have a task with the same ID (e.g. `MTEzMzQ1NjQ4NTQ4OTM0MjMzODI6MDow`). If user A and user B both have a task `"task-xyz"` and user A has assigned it to a profile, this query will return user A's assignment row when user B requests `?includeAssignments=true` for their own `"task-xyz"` task.

The precedent in `GET /api/tasks/[taskId]/assignments/route.ts` has the same gap β€" it also queries by `taskId` alone without a `userId` filter β€" so this PR is consistent with the existing code. But that consistency means the gap is now surfaced to the client rather than living only in an internal management endpoint.

**The fix** is to thread `session.user.id` from `route.ts` into the service and join through `Profile.userId`:

```ts
// service signature change
export async function getTaskAssignmentsByTaskIds(
  taskIds: string[],
  userId: string
): Promise<Map<string, AssignmentSummary[]>>

// query change
where: {
  taskId: { in: uniqueIds },
  profile: { userId },   // scopes to the authenticated user's profiles only
},
```

This is a one-line schema addition to the `where` clause and keeps the rest of the function identical.

---

### FILE: `src/lib/services/task-assignments.ts` β€" LINE 14 (`avatar: unknown`)

**TYPE SAFETY β€" `avatar: unknown` violates strict-TS rules and causes a downstream type error**

```ts
export interface AssignmentProfileSummary {
  id: string;
  name: string;
  color: string;
  avatar: unknown;   // <-- forbidden by project rules (no `any`/`unknown` escape hatches)
}
```

This leaks: `task-assignments.ts` exports `AssignmentSummary` with `profile: AssignmentProfileSummary`, that type is assigned into `TaskWithMeta.assignments` in `use-tasks.ts`, and `task-item.tsx` then passes `a.profile` directly to `<ProfileAvatar profile={a.profile} />`. `ProfileAvatar` expects `avatar: ProfileAvatarType` (which is `{ type: "initials"|"photo"|"emoji"; value: string; backgroundColor?: string }`). TypeScript accepts it today only because the `unknown` avatar field is structurally compatible with nothing β€" it will likely require a cast or will fail under stricter inference in a future TS bump.

Meanwhile `src/components/tasks/types.ts` already defines a correctly-typed `AssignmentProfileSummary` with `avatar: { type: "initials" | "photo" | "emoji"; value: string }`. The service layer should re-use that type rather than introducing a second, looser definition:

```ts
// task-assignments.ts β€" import from the shared types instead of redefining
import type { AssignmentSummary } from "@/components/tasks/types";
// or, if the circular import is a concern, extract both interfaces to a
// shared src/types/task-assignments.ts and import from there in both places.
```

The PR description mentions wanting to keep this module "free of cross-feature imports", which is a reasonable goal, but `unknown` is not the right escape hatch. A dedicated shared types file (or simply using the exact Prisma-inferred type for `avatar`) solves the layering concern without introducing a type hole.

---

### FILE: `src/components/tasks/use-tasks.ts` β€" LINE 49 (`useCallback`)

**CLAUDE.md VIOLATION β€" `useCallback` is forbidden (React Compiler handles memoization automatically)**

```ts
const fetchTasks = useCallback(async () => { ... }, [config]);
```

The `useCallback` wrapping `fetchTasks` (and the one wrapping `updateTask` at line 133) pre-date this PR β€" they are in `main` already. However this PR adds new logic inside `fetchTasks` and expands what it closes over, so it's a good moment to flag: per CLAUDE.md, manual memoization with `useCallback`/`useMemo`/`React.memo` is explicitly forbidden because React Compiler handles it. Removing both `useCallback` wrappers and writing the functions inline (or as plain `async function` declarations inside `useTasks`) is the correct fix. The `useEffect` dependency on `fetchTasks` would become a dependency on `config` directly, which is cleaner.

This isn't introduced by this PR but the PR touches those lines, so now is the right time to fix it.

---

### FILE: `src/components/tasks/__tests__/profile-scoped-task-list.test.tsx` β€" LINE 88 (the `describe` block)

**MISSING TEST CASE β€" `viewMode === "family"` path is never exercised**

The component has two distinct code paths:

```ts
viewMode === "family" || !activeProfile
  ? { ...config, profileFilter: null }   // family path
  : { ...config, profileFilter: ..., showUnassigned: ... }  // profile path
```

Every test in this file exercises the profile path (the default `viewMode` from `ProfileProvider` is `"profile"`). There is no test that sets `viewMode` to `"family"` and asserts that `profileFilter` is forced to `null` regardless of what was on the incoming `config`. The "clears the filter when no profile is active" case covers `!activeProfile`, not the family-mode branch.

A missing test for this is a gap because the family-mode check is the primary entry point for `ProfileScopedTaskList` in the family calendar view β€" it's the whole reason the component exists as a wrapper rather than inlining the logic in `TaskList`. The test should set `viewMode` via `ProfileProvider`'s `setViewMode` (or mock `useProfile` directly) and confirm `config.profileFilter === null` even when `activeProfile` is non-null.

---

BenSeymourODB:
Thanks for the review β€" addressed the four items in `73cc251`.

**1. Cross-tenant data leak β€" fixed.** `getTaskAssignmentsByTaskIds` now requires the caller's `userId` and adds `profile: { userId }` to the Prisma `where`. `/api/tasks` threads `session.user.id` through to the service. Two tests pin the new behaviour: a query-shape assertion proving the `profile.userId` filter is on every call, and a dedup test repeating the same assertion.

**2. `avatar: unknown` β€" fixed.** The structured `AssignmentProfileSummary` type now lives in `src/components/tasks/types.ts` (with the optional `backgroundColor` field added) and the service `import type`s + `export type`s it. The Prisma `JsonValue` is cast to that shape at the boundary β€" the only point we can't avoid an assertion β€" so downstream consumers see strict types.

**3. Missing tests β€" added two.**
- Service: error propagation when Prisma rejects (`connection refused`).
- Route: 500 path when the assignments lookup throws after a successful Google fetch β€" route surfaces the catch-all rather than returning a half-populated tasks array.
- Bonus from the second review summary: `ProfileScopedTaskList` now has a family-view-mode test that flips `viewMode` via `setViewMode("family")` and asserts `profileFilter` becomes `null` even with an active profile β€" the path the previous five tests didn't cover.

**4. The earlier review's "double-fetch on mount" / second review's `useCallback` flag β€" deferred.** Both `useCallback` wrappers in `use-tasks.ts` predate this PR (`main` has them on lines 49 and 133). I expanded the inner body of the existing wrapper but didn't add new memoization. Removing them touches behaviour beyond this issue's scope and would risk destabilising the existing 30+ `useTasks` tests, so I'd rather file a follow-up that removes both manual-memoization sites along with any other React-Compiler-rule violations across the repo. Happy to either open that issue or fold it in here if you'd prefer.

Full suite: **100 files, 1470 tests** passing locally; CI should confirm shortly.

---

PR #225, opened 2026-05-02
feat(calendar): add useDateNow hook with midnight-tick refresh (#204)

BenSeymourODB:
Closes #204.

## Why

A 24/7 wall display left open across midnight currently shows stale
"today" affordances — `today`-highlight stays on yesterday, the
`Today`-button enable state sticks, and labels like
"Today / Tomorrow / Yesterday" lag a day behind. The same pattern was
flagged as deferred follow-up in PRs #140, #145, and #149.

### Phase 1 — `useDateNow` hook (`src/lib/hooks/use-date-now.ts`)
- `useDateNow()` returns the current `Date` and re-renders subscribers
  on the next local-midnight tick.
- `useTodayStartOfDay()` convenience for `isSame*` comparisons.
- `useSyncExternalStore` with **one shared `setTimeout`** chained to
  next midnight. The timer is cleared when the last subscriber
  unmounts and re-armed on remount, so an idle calendar app never
  burns timers.
- 9 unit tests with `vi.useFakeTimers()`: initial value, midnight
  rollover, multi-subscriber notification, re-scheduling next
  midnight, single-shared-timer guarantee, last-unmount cleanup,
  remount re-scheduling, and `startOfDay` helper behaviour.

### Phase 2 — wired into all `today`-aware surfaces
Replaced module-level / per-render `today = new Date()` captures in:
- `DayCalendar` — Today button disabled state + Today/Tomorrow/Yesterday label + today-column highlight
- `WeekCalendar` — Today button + today-cell highlight + current-week flag
- `SimpleCalendar` — today cell `aria-current` + bg highlight + current-month flag
- `YearCalendar` — `data-today` on each day cell + current-year flag
- `MiniCalendarSidebar` — today highlight in the mini grid
- `AnalogClockView` — all-day-today filter + heading label

Integration tests in
`src/components/calendar/__tests__/midnight-rollover.test.tsx`
mount each consumer before midnight, advance the fake clock past the
boundary, and assert the today affordance flips to the new day.

## Tests
- 1485 / 1485 vitest cases pass (16 new, 1469 existing)
- `pnpm lint:fix && pnpm format:fix && pnpm check-types` clean

## Test plan
- [x] Phase 1 hook tests pass
- [x] Phase 2 integration tests pass for all six consumers
- [x] Full suite passes
- [x] Type-check clean

## Notes
- No per-component `setInterval` is introduced; existing `NowLine`
  minute-tickers in `DayCalendar` / `WeekCalendar` are independent and
  unaffected (they're for the red current-time line, not the day-level
  today affordances).
- React Compiler-friendly: no `useMemo` / `useCallback`.


BenSeymourODB (commented):
## Code Review — `useDateNow` midnight-tick hook (#204)

Overall the implementation is clean and the approach is sound. There are three issues worth addressing before merge — one correctness bug (StrictMode double-subscribe), one snapshot-stability problem that can cause unnecessary re-renders, and one false-positive test — plus a few smaller observations.

---

### 1. Bug: React StrictMode double-subscribe leaves a leaked timer (correctness)

**File:** `src/lib/hooks/use-date-now.ts`, `subscribe` function (~line 44)

In React 18/19 StrictMode, `useSyncExternalStore` calls `subscribe` twice with two *different* callbacks during the same component mount, then immediately calls the first unsubscribe. The sequence is:

```
subscribe(cb1)   → subscribers = {cb1}, timer armed ✓
subscribe(cb2)   → subscribers.size > 0, so NO refresh, NO new timer
unsubscribe(cb1) → subscribers = {cb2}, size > 0 → timer NOT cleared ✓ (ok)
```

That sequence is actually fine. The real problem is a rapid **last-unmount / remount** race under StrictMode: StrictMode mounts, unmounts (clearing the timer + emptying `subscribers`), then remounts. The re-arm on the second mount calls `scheduleTick()` correctly. However there is a window where `currentDate` could be stale if a tick fires between the unmount and the remount *and* the `subscribers.size === 0` guard in `subscribe` only refreshes `currentDate` at first subscription — that part is actually handled. So StrictMode itself is safe.

The real concern is subtler: `scheduleTick` captures `new Date()` at the moment the timer is *scheduled* (line 41), not when the tick fires. If the timer is re-armed after an unmount/remount cycle that crosses midnight (e.g., the calendar goes offscreen and back on), `msUntilNextLocalMidnight` is computed correctly each time from `new Date()`, so this is fine.

**Actual bug:** `subscribe` checks `if (timeoutId === null) scheduleTick()` (line 52–54), but `scheduleTick` itself also checks `if (timeoutId !== null) return` (line 35). These two guards are redundant — either is sufficient alone. The redundancy is harmless, but it obscures intent. More importantly, this means the `setTimeout` spy test (`"uses a single shared timer"`) will pass even if the inner guard is removed, giving false confidence. Not a runtime bug, but worth a comment.

---

### 2. Snapshot instability: `useTodayStartOfDay` creates a new `Date` object every render (correctness / performance)

**File:** `src/lib/hooks/use-date-now.ts`, line 83

```ts
export function useTodayStartOfDay(): Date {
  return startOfDay(useDateNow());
}
```

`startOfDay` always returns a **new `Date` object**. `useSyncExternalStore` guarantees stability of `getSnapshot`'s return value only for `currentDate` itself — but `useTodayStartOfDay` wraps it in `startOfDay`, producing a fresh object on every render. React Compiler will not memoize across a function boundary here.

**Consequence:** any component using `useTodayStartOfDay()` that passes `today` as a prop or into a comparison will always receive a new object reference, preventing React Compiler's automatic bail-out optimisation. For a wall display with many mounted components this compounds re-render cost.

**Fix:** either cache the `startOfDay` result in the store alongside `currentDate`, or make `useTodayStartOfDay` compute once per tick:

```ts
let currentDateSOD: Date = startOfDay(new Date());

// Inside the tick callback:
currentDate = new Date();
currentDateSOD = startOfDay(currentDate);

function getSnapshotSOD(): Date { return currentDateSOD; }

export function useTodayStartOfDay(): Date {
  return useSyncExternalStore(subscribe, getSnapshotSOD, getSnapshotSOD);
}
```

This ensures the returned `Date` reference is stable between ticks, matching the contract consumers expect.

---

### 3. False-positive test: "single shared timer" test proves nothing about sharing (test quality)

**File:** `src/lib/hooks/__tests__/use-date-now.test.ts`, lines 83–94

```ts
it("uses a single shared timer across many subscribers", () => {
  const setTimeoutSpy = vi.spyOn(globalThis, "setTimeout");
  setTimeoutSpy.mockClear();

  const hooks = Array.from({ length: 5 }, () =>
    renderHook(() => useDateNow())
  );
  expect(hooks).toHaveLength(5);      // ← this assertion is trivially true
  expect(setTimeoutSpy).toHaveBeenCalledTimes(1);
  setTimeoutSpy.mockRestore();
});
```

This test passes because `__resetUseDateNowForTests()` in `beforeEach` clears `timeoutId`, so the *first* `renderHook` arms the timer and the remaining four skip it. But `vi.spyOn` on `setTimeout` with `vi.useFakeTimers()` active may intercept Vitest's internal timers too, making the call count unreliable if Vitest itself calls `setTimeout` during `renderHook`. The spy is installed *before* the hooks are rendered, so any framework-internal timer calls between `setTimeoutSpy.mockClear()` and the assertion will inflate the count. 

**Additionally**, `expect(hooks).toHaveLength(5)` is a tautology — `Array.from({length: 5}, ...)` always produces 5 elements regardless of hook correctness.

**Fix:** Install the spy before rendering, capture the call count after all renders, and assert it equals 1 minus however many calls happened during the `renderHook` setup itself. Or restructure to spy *only* on `scheduleTick` via a module spy. At minimum remove the tautological assertion.

---

### 4. `__resetUseDateNowForTests` exported from production code (minor)

**File:** `src/lib/hooks/use-date-now.ts`, lines 90–97

Exporting a `__`-prefixed test helper from a `"use client"` production module is the pragmatic choice given the module-level singleton state and no dependency-injection seam. It's a common pattern. However, it will be included in every client bundle. 

If bundle size becomes a concern later, this could be moved behind a `process.env.NODE_ENV === "test"` guard or extracted to a separate test-utility file that re-exports from the module's internals via a vitest `vi.importActual` approach. For now this is acceptable — just noting it is shipped to production.

---

### 5. `getServerSnapshot` is identical to `getSnapshot` — semantically correct but worth a comment

**File:** `src/lib/hooks/use-date-now.ts`, line 74

```ts
return useSyncExternalStore(subscribe, getSnapshot, getSnapshot);
```

Passing the same function for both `getSnapshot` and `getServerSnapshot` is correct here because the module is `"use client"` — SSR will never call `subscribe`, so `getServerSnapshot` is essentially dead code. A one-line comment would clarify intent for future readers and prevent someone from "fixing" it to return a hardcoded sentinel.

---

### 6. DST / clock-skew resilience (minor)

**File:** `src/lib/hooks/use-date-now.ts`, `msUntilNextLocalMidnight`

During a DST spring-forward, a 23-hour day means midnight could arrive 1 hour earlier than computed. The `Math.max(1, ...)` guard handles the edge case of a zero/negative delay, and after the tick fires `scheduleTick` recomputes from `new Date()` so the next interval is always correct. The only pathological case is a forward clock change of more than `msUntilNextLocalMidnight` milliseconds during the wait — in practice DST shifts are 1 hour max, so this is safe. No action needed, but worth a comment near `msUntilNextLocalMidnight`.

---

### 7. Integration test for `MiniCalendarSidebar` is fragile (test quality)

**File:** `src/components/calendar/__tests__/midnight-rollover.test.tsx`, lines 153–168

```ts
const todayBefore = document.querySelector('[data-today="true"]');
expect(todayBefore?.textContent).toContain(String(BEFORE_MIDNIGHT.getDate()));
```

`BEFORE_MIDNIGHT.getDate()` is `2` (May 2). `NEXT_DAY.getDate()` is `3`. These are distinct numbers so the test passes correctly here. However if the test were run with dates where the "before" and "after" day numbers overlap in string content (e.g., day 1 and day 21 — `"21".includes("1")` is true), the `textContent.toContain(String(...))` assertion would be a false positive. The fixture dates are hardcoded so this isn't a live issue, but prefer `toBe(String(...))` over `toContain` for day-number checks to be safe.

---

### Component wiring — no issues found

- `DayCalendar`: `startOfDay` is correctly retained for `eventCoversDay` helper; `useTodayStartOfDay` used for all `today` comparisons. The module-level `const today = startOfDay(new Date())` stale capture is fully removed.
- `WeekCalendar`: `startOfDay` import correctly removed (no remaining local uses); `useTodayStartOfDay` covers all comparison sites.
- `SimpleCalendar`, `MiniCalendarSidebar`, `YearCalendar`, `AnalogClockView`: all switch from bare `new Date()` to `useDateNow()`. No missed call sites found.
- No `useMemo`/`useCallback` added — React Compiler compliant.
- All six files are already `"use client"` — the hook's client-only module state is safe.
- No `any` types introduced.
- Tailwind colors are all standard palette.

---

### Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 2 | Medium | `use-date-now.ts:83` | `useTodayStartOfDay` returns new object every render; breaks reference stability |
| 3 | Medium | `use-date-now.test.ts:83` | Timer-sharing test has a tautological assertion and may count framework timers |
| 4 | Low | `use-date-now.ts:90` | Test helper exported into production bundle |
| 5 | Low | `use-date-now.ts:74` | `getServerSnapshot` reuse unexplained |
| 7 | Low | `midnight-rollover.test.tsx:157` | `toContain` on day number is fragile; prefer `toBe` |

Issues 1 (StrictMode), 5, 6 are informational — no code change required. Issues 2 and 3 should be fixed before merge; 4 and 7 are at the author's discretion.

---

BenSeymourODB:
Second-round response to the review (commit cb0f430):

| # | Review item | Resolution |
|---|---|---|
| 1 | StrictMode double-subscribe (informational) | No change — review confirms current code is safe under StrictMode. |
| 2 | `useTodayStartOfDay` returned a new `Date` every render | **Fixed.** Cached `currentStartOfDay = startOfDay(currentDate)` in the module store and added a dedicated `getStartOfDaySnapshot`; `useTodayStartOfDay` now goes through `useSyncExternalStore` directly so the reference is stable between midnight ticks. |
| 3 | Tautological + framework-noise prone "single shared timer" test | **Fixed.** Mount one subscriber first to absorb framework-internal `setTimeout` calls during `renderHook`, *then* install the spy and mount four more subscribers and assert the spy was never invoked. Tautological `expect(hooks).toHaveLength(5)` removed. |
| 4 | `__resetUseDateNowForTests` shipped to production bundle | No change for now — pragmatic given module-level singleton. Filed mentally as future cleanup; happy to gate behind `NODE_ENV` or move to a side module if you'd prefer. |
| 5 | `getServerSnapshot` reuse not commented | **Fixed.** Added a short comment inside `useDateNow` explaining the SSR path never subscribes but the argument can't be undefined. |
| 6 | DST / clock-skew resilience (informational) | **Comment added** to `msUntilNextLocalMidnight` noting that ±1h DST shifts are absorbed because each interval is recomputed from `new Date()` after every tick. |
| 7 | `MiniCalendarSidebar` `toContain(String(day))` is fragile | **Fixed.** Tightened to `expect(...trim()).toBe(String(...))` so overlapping numerals can't mask a wiring break. |

All 1485 vitest cases still pass; `pnpm check-types` clean. Ready for re-review.

---

PR #228, opened 2026-05-02
fix(calendar): include Saturday-afternoon events in week view (#201)

BenSeymourODB:
Closes #201.

## Summary

`getEventsForWeek` used `weekDates[6]` as the upper bound of the week. `weekDates[6]` is the *start* of Saturday (00:00:00), so any event whose `startDate` was at 00:00:01 or later was silently excluded — that's the entire Saturday.

The fix replaces the bound with `endOfDay(weekDates[6])` so the comparison covers the whole final day. This mirrors the semantics of `getEventsForMonth` (`endOfMonth`) and `getEventsForYear` (`endOfYear`).

## Tests (TDD: red → green)

Added three regression tests in `src/lib/__tests__/calendar-helpers.test.ts > getEventsForWeek`:

1. **`includes Saturday-afternoon events on the last day of the week (regression: #201)`** — covers Saturday 08:00, 14:00, and 23:30. Failed on the previous code (returned `[]`); now passes.
2. **`excludes events that start on Sunday of the next week`** — boundary check the other direction (Sun 00:00 of week N+1 must not leak into week N).
3. **`excludes events fully before the week (Saturday of the prior week)`** — symmetric prior-week boundary.

Full suite: **1575/1575 passing**, lint clean, types clean, format clean.

## Drive-by

`src/components/calendar/__tests__/AgendaList.test.tsx` was failing `pnpm check-types` on `main` because the `makeContext` mock was missing the `deleteEvent` field added to `ICalendarContext`. Added the missing field (one line) so the type-check gate passes.

## Acceptance criteria from #201

- [x] `getEventsForWeek` includes all events whose start falls anywhere on the final day of the week
- [x] Unit test covers an event at Saturday 14:00 — fails on old code, passes on new
- [ ] Verified visually that an event scheduled for Saturday afternoon shows in `WeekCalendar` — *deferred: pure helper change, exercised by `WeekCalendar.tsx:128` (`getEventsForWeek(events, selectedDate)`); no rendering logic changed. Manual visual check welcomed in review.*

---
_Generated by [Claude Code](https://claude.ai/code/session_014914HntGoaPxJJcYvc1RCt)_


BenSeymourODB (commented):
Overall: the fix is correct and well-reasoned. Three focused comments below.

---

**1. Simpler alternative: `endOfWeek` is already imported — use it instead of `endOfDay(weekDates[6])`**

`endOfWeek` from date-fns does exactly this (returns 23:59:59.999 of the last day), it already respects `weekStartsOn`, and it completely removes the dependency on `getWeekDates` for this purpose. The current approach works but adds a subtle two-step indirection (`getWeekDates` → `[6]` → `endOfDay`) that could mislead a future reader into thinking `weekDates[6]` is semantically meaningful beyond "the last day of the array".

Suggested replacement (line 364):
```ts
// before
const endOfWeekDate = endOfDay(weekDates[6]);

// after — more explicit, no array indexing, consistent with WEEK_STARTS_ON usage everywhere else
const endOfWeekDate = endOfWeek(date, { weekStartsOn: WEEK_STARTS_ON });
```
`endOfWeek` is already imported (line 18) and used at line 86 for exactly this pattern. Both produce the same millisecond value (`T23:59:59.999`), so behaviour is identical; this is purely a readability/maintainability note.

---

**2. The `AgendaList` mock uses `vi.fn()` for `deleteEvent`, but every other test file uses `vi.fn().mockResolvedValue(undefined)` to match the `Promise<void>` signature**

(`ViewSwitcher.test.tsx:50`, `CalendarFilterPanel.test.tsx:42`, `SimpleCalendar.test.tsx:75`, `AddEventButton.test.tsx:53` all use `.mockResolvedValue(undefined)`.)

The bare `vi.fn()` returns `undefined` synchronously, so any code that calls `deleteEvent(...).then(...)` or `await deleteEvent(...)` will throw a runtime error in tests. It passes type-check today only because the mock is typed via `Partial<ICalendarContext>` and vitest doesn't validate promise return types at that level.

```ts
// line 65 — change to:
deleteEvent: vi.fn().mockResolvedValue(undefined),
```

---

**3. Tests are thorough for the happy path; consider one multi-day spanning event**

The three new regression tests cover single-day Saturday events well. One gap: a multi-day event that *ends* on Saturday afternoon (e.g. `startDate: "2024-03-14T22:00:00"`, `endDate: "2024-03-16T15:00:00"`) was also broken by the old bound — `eventEnd >= startOfWeekDate` would pass, but `eventStart <= endOfWeekDate` is what changed. Worth one extra case to lock in the full overlap logic. Not a blocker.

---

BenSeymourODB:
Addressed all three review points in `7eba961`:

**1. `endOfWeek` instead of `endOfDay(weekDates[6])`** — done. Replaced both bounds with `startOfWeek(date, { weekStartsOn: WEEK_STARTS_ON })` / `endOfWeek(date, { weekStartsOn: WEEK_STARTS_ON })`, dropping the `getWeekDates` indirection and the now-unused `endOfDay` import. Mirrors the pattern at `calendar-helpers.ts:84-86`.

**2. `deleteEvent` mock returns a real Promise** — done. `vi.fn().mockResolvedValue(undefined)` so `await`/`.then()` paths work, matching `ViewSwitcher`/`CalendarFilterPanel`/`SimpleCalendar`/`AddEventButton`.

**3. Multi-day spanning event tests** — added two:
- Thu 22:00 → Sat 15:00 (locks in the end-of-overlap case)
- Sat 14:00 → Mon 09:00 (this one was actually broken by the old bound — `eventStart` 14:00 > old upper bound 00:00, so it would have been dropped despite obviously belonging to the week)

Tests: 1577/1577 passing. Lint, format, types all clean.

---

**Note on CI**: the failing `ci` check is the *Validate Prisma migrations are up to date* step — this is pre-existing on `main` (the equivalent local command `npx prisma migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma --exit-code` exits 0 here). Not caused by this PR — flagging only because the gate looks red.

---

PR #151, opened 2026-04-21
feat(calendar): add compact settings panel with week-start toggle (#86)

BenSeymourODB:
Closes #86.

## Summary

Introduces `CalendarSettingsPanel` — a compact popover triggered by a gear button in the calendar toolbars — that exposes the four calendar display preferences owned by `CalendarProvider`:

- Event badge style (colored pills vs dots)
- 24-hour vs 12-hour time format
- Agenda mode group-by (date / color)
- Week-start day (Sunday / Monday)

`weekStartDay` is a new setting for this issue; the other three were already wired up and just needed UI. All four persist to `localStorage` via the existing `useLocalStorage` hook in `CalendarProvider`, with a `??` fallback so pre-#86 payloads rehydrate cleanly.

The panel is mounted next to the theme toggle on `/calendar` and next to the view switcher on `/test/calendar?sidebar=true`.

## Implementation notes

- **Provider**: extended `ICalendarContext` + `CalendarSettings` with `weekStartDay` / `setWeekStartDay`. Defaults to Sunday (`0`).
- **Helpers**: `getShortWeekdayLabels()` now takes an optional `weekStartsOn: Day` argument (defaults to the project-wide `WEEK_STARTS_ON` constant so unrelated callers are unaffected).
- **Views**: `SimpleCalendar` and `MiniCalendarSidebar` now consume `weekStartDay` from the provider and pass it to `getShortWeekdayLabels` / `startOfWeek` / `endOfWeek` / the leading-padding calculation so toggling the setting actually reorders the weekday row *and* shifts the 1st-of-the-month into the correct column in both grids.
- **Tailwind**: panel uses only default palette tokens (`text-foreground`, `text-muted-foreground`, `border-border`, etc.).
- **React Compiler**: no manual memoization; derived values (`rotatedDowLabels`, `weekdayHeaders`) are recomputed per render.

## TDD coverage

- **`CalendarProvider.test.tsx`** (new, 5 tests) — defaults, `weekStartDay` persistence, multi-field persistence, rehydration from `localStorage`, legacy-payload fallback.
- **`CalendarSettingsPanel.test.tsx`** (new, 7 tests) — gear label, popover exposes all 4 controls, reflects provider values, toggles each control.
- **`MiniCalendarSidebar.test.tsx`** — added Monday-first header-order regression.
- **`SimpleCalendar.test.tsx`** — added Monday-first header-order regression.
[...]

PR #229, opened 2026-05-02
feat(calendar): respect user weekStartDay in calendar helpers (#205)

BenSeymourODB:
## Summary

Wires the user-configurable `weekStartDay` setting (introduced for the settings panel in PR #151) all the way through the helper layer so every week-mode view, helper, and counter respects the user's preference.

Closes #205.

### Phase 1 — helpers (`src/lib/calendar-helpers.ts`)
`rangeText`, `getEventsCount`, `getCalendarCells`, `getEventsForWeek`, `getWeekDates`, and `getEventsByMode` now take an optional `weekStartsOn: Day = WEEK_STARTS_ON`. The default keeps every existing caller working unchanged; new callers pass `weekStartDay` from `CalendarProvider`. Five new unit tests cover both Sunday-start and Monday-start branches.

Also patches a pre-existing TypeScript error in `AgendaList.test.tsx` (missing `deleteEvent` mock) so the type-check gate passes — same one-line fix that PR #228 happens to apply.

### Phase 2 — wire `WeekCalendar`
`WeekCalendar` now reads `weekStartDay` from `useCalendar` and threads it through `getShortWeekdayLabels`, `getWeekDates`, `getEventsForWeek`, and the direct `startOfWeek/endOfWeek/isSameWeek` calls. A component test asserts the Monday-first range, header label, and day-column test ids; an E2E test toggles the setting from the popover with the week view open and asserts the range and weekday headers update.

`WeekCalendar` was the only production caller of any of the parameterised helpers — the rest (`rangeText`, `getEventsCount`, `getCalendarCells`, `getEventsByMode`) had no production callers, so wiring is gated behind their first use.

## Coordination with #201 / PR #228

`getEventsForWeek` currently uses `weekDates[6]` (start-of-last-day) as the upper bound, which is the off-by-one PR #228 fixes. To avoid duplicating that fix, the new Monday-first test for `getEventsForWeek` exercises only the leading-Sunday inclusion direction. When PR #228 lands, the rebase is trivial — the bound switch composes cleanly with the new parameter.

## Test plan

- [x] `pnpm test` — 1579 unit/component tests pass (5 new helper tests + 1 new WeekCalendar test)
- [ ] CI Playwright run: `e2e/calendar-settings-panel.spec.ts` covers the week-view flip on settings change
- [ ] Manual verification deferred to local dev — the headless E2E exercises the same path

## Out of scope (deferred)

- `calendar-keyboard.ts` keyboard navigation (Home/End → start/end of week) still uses `WEEK_STARTS_ON`. The issue scope is the five helpers in `calendar-helpers.ts`; keyboard wiring is a small follow-up that can ride on top once this lands.


BenSeymourODB (commented):
Solid, well-scoped PR. The helper parameterisation is clean, defaults are backward-compatible, `WeekCalendar` wiring is complete, and all 1579 tests pass locally. Four substantive comments below — one real bug, two worth tracking, one minor nit.

---

### 1. `getEventsForWeek` Monday-first test passes for the wrong reason (latent bug) — `src/lib/__tests__/calendar-helpers.test.ts` ~line 430

The test correctly notes the end-of-week boundary is deferred to PR #228. But the risk is subtler than the comment implies: `weekDates[6]` is the *start* of the last day (midnight). For the Monday-first week in the test (`Mar 11–17`), `weekDates[6]` = `2024-03-17T00:00:00`. A noon event on that Sunday (e.g. `"2024-03-17T12:00:00"`) would be **excluded** by the current filter (`eventStart <= endOfWeekDate`) even though it falls within the user's displayed week. This means the Monday-first path has the same midnight-boundary bug as the Sunday-first path — it just isn't visible yet because no test covers that slot.

The current test only exercises the *leading* boundary (Sunday Mar 10 excluded in Mon-first), which is genuinely clean. But once PR #228 lands and widens the bound to `endOfDay(weekDates[6])`, a noon event on the *final* day of a Monday-first week will start being included — which is correct but untested here.

Suggested addition to the TODO note in this test:

```ts
// TODO(#201/PR #228): when endOfWeekDate becomes endOfDay(weekDates[6]),
// also assert that a Mon-first week includes a noon event on Sun Mar 17
// (currently excluded by the midnight bound, same as the Sun-first case).
```

Without this, whoever merges second is likely to forget the symmetric case.

---

### 2. `rangeText` is parameterised but not wired to `WeekCalendar` — `src/components/calendar/WeekCalendar.tsx` line 157

`WeekCalendar` formats its own range header inline with `format(weekStart, …) + "–" + format(weekEnd, …)` rather than calling `rangeText`. That is fine for now (no production caller exists), but the consequence is that `rangeText`'s new `weekStartsOn` parameter has **zero production callers** — it is tested in isolation only. The PR description correctly notes this, but the issue acceptance criteria (#205) lists `rangeText` as one of the helpers to fix. If there is an upstream consumer of `rangeText` (e.g. a header component, API route, or future view switcher) it will silently continue using `WEEK_STARTS_ON`.

This is not a blocker for merging, but if `rangeText` is expected to replace the inline formatting in `WeekCalendar` at some point, a follow-up issue to thread it through would prevent the parameter from becoming dead code. The issue spec also asks that "week-view … flip correctly" — the range text visually does flip (via the inline format), just not via `rangeText`.

---

### 3. `calendar-keyboard.ts` deferral is appropriate but deserves a tracking issue — `src/lib/calendar-keyboard.ts` lines 75–77

The PR description acknowledges that `Home`/`End` keyboard navigation (`WEEK_START`/`WEEK_END` actions) still use `WEEK_STARTS_ON`. This is a real UX hole: a user who sets Monday-first and presses `Home` in the month grid will jump to Sunday, not Monday. The PR body says "small follow-up" but there is no linked issue.

Given the project tracks everything on the board, consider filing a follow-up issue now so it doesn't slip. The fix is one-line (`applyCalendarKeyboardAction` needs a `weekStartsOn` argument), but `calendarKeyboard` is called from `SimpleCalendar` which already reads `weekStartDay` — so the plumbing is ready.

---

### 4. Nit: `getEventsCount` Monday-first test comment slightly off — `src/lib/__tests__/calendar-helpers.test.ts` ~line 176

The comment says:
```
// testDate = Fri Mar 15 2024. With weekStartsOn = 1 (Monday), the week
// runs Mon Mar 11 – Sun Mar 17. Mar 10 falls into the prior week.
```

That is all correct. But `getEventsCount` uses `isSameWeek` (not the `weekDates` array), so this test does *not* exercise the midnight-boundary issue at all — `isSameWeek` compares calendar weeks, not timestamps. Worth one sentence in the comment to prevent a future reader from conflating it with the `getEventsForWeek` boundary discussion:

```ts
// Note: getEventsCount uses isSameWeek, which is boundary-safe (compares
// calendar-week membership, not raw timestamps).
```

Minor — not blocking.

---

BenSeymourODB (commented):
Thanks for the careful read. Addressed all four comments in `0c67776`:

**#1 — `getEventsForWeek` Mon-first end-of-week TODO.** Added the suggested `TODO(#201/PR #228)` directly above the test note, calling out the symmetric noon-on-Sun-Mar-17 case so it isn't lost when the bound widens.

**#2 — `rangeText` has zero production callers.** Filed #230 as the follow-up. Didn't inline it here because `WeekCalendar` uses an em-dash and `rangeText` uses an ASCII hyphen, so a swap means glyph reconciliation + updating the existing `rangeText` unit tests — bigger than this PR's scope.

**#3 — `calendar-keyboard.ts` deferral.** Filed #231 as the follow-up. Agreed it's a real UX hole, but kept it out of #205's scope per the issue spec (lists the five `calendar-helpers` helpers explicitly).

**#4 — `getEventsCount` test comment clarity.** Added a one-sentence note that `isSameWeek` is boundary-safe so future readers don't conflate it with the `getEventsForWeek` midnight-bound discussion.

CI is green on `cd210f6`; the docs-only commit `0c67776` is in flight.

---

PR #232, opened 2026-05-02
chore(http): wire fetchWithRetry into Google Calendar API routes (#217)

BenSeymourODB:
## Summary

Adopts `fetchWithRetry` (added in #160) at the call sites that #217 explicitly identified — the four Google Calendar API routes plus the OAuth token-refresh inside the NextAuth `session` callback. Transient failures (429 / 5xx / `fetch`-level network errors) now get retried with exponential backoff + jitter instead of bubbling straight to the client.

Closes #217

## Phase 1 — Calendar API routes

- `src/app/api/calendar/events/route.ts`
- `src/app/api/calendar/events/[id]/route.ts`
- `src/app/api/calendar/calendars/route.ts`
- `src/app/api/calendar/colors/route.ts`

Each was a single `fetch(...)` → `fetchWithRetry(...)` swap; success-path response shapes are unchanged. Existing route tests continue to stub `global.fetch`, which `fetchWithRetry` still calls under the hood — they pass without modification, except for two that mocked a 5xx response (which now goes through retries). Those got `headers: new Headers()` and fake timers so the retry sleeps don't slow the suite.

A transient-failure regression test on `/api/calendar/events` asserts that a 503 followed by a 200 yields the 200 response without a second client-visible failure.

### Pre-existing tsc fix

`src/components/calendar/__tests__/AgendaList.test.tsx` was missing `deleteEvent` from its mock `ICalendarContext` after the field was added to the provider in #197. Folded a one-line addition in to keep CI green.

## Phase 2 — Auth token refresh

The Google OAuth refresh-token grant lived inline in the NextAuth `session` callback, which made it both untestable in isolation and a holdout from the `fetchWithRetry` adoption. Extracted into `src/lib/auth/refresh-google-token.ts` (`refreshGoogleAccessToken`) so the actual `fetch` call is a discrete, testable surface and inherits transient-failure retries.

Behavioural surface to the callback is preserved:

- tokens come back on success (`{ access_token, expires_in, refresh_token? }`)
- the parsed error envelope is thrown on non-OK (matching the legacy `tokensOrError` throw the callback already catches)
- Google's "no new refresh_token" rotation policy is left to the call site to fall back to the existing plaintext refresh token

Tests in `src/lib/auth/__tests__/refresh-google-token.test.ts` cover:

1. Argument shape (URL, method, headers, URL-encoded body fields)
2. Success-path token parsing
3. Throw-on-4xx error envelope (non-transient — no retry)
4. Transient 503 → 200 regression returns the recovered tokens with two underlying fetch attempts

## Test plan

- [x] `pnpm exec vitest run` — 108 files / 1577 tests pass
- [x] `pnpm lint:fix` — no errors (5 pre-existing warnings)

https://claude.ai/code/session_01Fp4EqwDBu8ZdWY2PEW4rEE


BenSeymourODB:
**Code review — first pass**

CI is still in progress at time of review, so the test-run numbers in the PR description are from a local run; I haven't verified CI output independently.

---

### Issue 1 — `src/app/api/auth/refresh-token/route.ts` is still on bare `fetch` [blocker]

`src/lib/auth/auth.ts` had its token-refresh fetch extracted into `refreshGoogleAccessToken` (which now uses `fetchWithRetry`), but there is a second server-side token-refresh route at `src/app/api/auth/refresh-token/route.ts` that makes the identical `POST https://oauth2.googleapis.com/token` call with `URLSearchParams` and is still on bare `fetch` (line 33). It is in scope for #217 for the same reason the auth.ts call was: a transient 5xx from Google drops the client's session refresh silently. That route could call `refreshGoogleAccessToken` directly (passing `NEXT_PUBLIC_GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`), or at minimum adopt `fetchWithRetry` inline. Either way, this is a missed migration that the PR description does not acknowledge as a known exclusion.

---

### Issue 2 — `mockFetch.toHaveBeenCalledTimes(2)` in the events retry test is correct but asserts the wrong thing [minor]

`src/app/api/calendar/events/__tests__/route.test.ts` (the new `"retries a transient 503"` test) correctly checks `expect(mockFetch).toHaveBeenCalledTimes(2)`. That count is right *because* the route fetches exactly one calendar (`calendarId=primary`), making one attempt + one retry = two calls. But if `calendarIds` were ever expanded (e.g. via `?calendarIds=primary,work`) that same assertion would over-count. More concretely, the assertion proves two `fetch` calls happened but does **not** prove that the second call was a **retry of the same URL** rather than a second calendar fetch. Consider adding `expect(mockFetch.mock.calls[0][0]).toBe(mockFetch.mock.calls[1][0])` (same URL string both times) so the test actually pins the retry-wiring rather than just the call count. As written the test would pass even if retry were disabled and a second calendar call happened to succeed.

---

### Issue 3 — `refreshGoogleAccessToken` test: the 4xx throw shape is verified but the auth.ts `catch` branch is not [minor]

`src/lib/auth/__tests__/refresh-google-token.test.ts` checks that a 400 response causes the function to throw the parsed error body (`rejects.toEqual(errorBody)`), which matches the old `throw tokensOrError` in auth.ts. Good. What is not tested is the **network-error / `TypeError` path** — if `fetchWithRetry` itself throws (e.g. DNS failure after all retries exhausted), auth.ts's outer `catch` receives a `TypeError`, logs it, and sets `session.error = "RefreshTokenError"`. That branch works correctly without any code change, but a test that lets `mockFetch` throw `new TypeError("fetch failed")` and asserts the rejection propagates would close the coverage gap that currently leaves the retry-exhausted path unverified in this helper's own test suite. Not a correctness bug in the PR, but worth noting since the PR explicitly targets wiring test coverage.

---

**Everything else looks good:** the four route swaps are mechanically correct, the fake-timer pattern with `try/finally` is the right way to wrap tests that use `vi.useFakeTimers()`, `headers: new Headers()` correctly prevents the `Retry-After` lookup from throwing on the partial mock objects, the `deleteEvent: vi.fn().mockResolvedValue(undefined)` shape matches `(eventId: string, calendarId: string) => Promise<void>`, and the `refreshGoogleAccessToken` extraction is behaviour-preserving (same URL, method, body encoding, throw shape, and return type as the deleted inline code). No `as any` casts or manual memoization introduced.

---

src/app/api/calendar/events/__tests__/route.test.ts
```diff
@@ -597,5 +597,60 @@ describe("/api/calendar/events", () => {
...
+                  items: [
+                    {
+                      id: "event-1",
+                      summary: "Recovered",
+                      start: { dateTime: "2024-03-15T10:00:00Z" },
+                      end: { dateTime: "2024-03-15T11:00:00Z" },
+                    },
+                  ],
+                  summary: "Primary Calendar",
+                }),
```
BenSeymourODB, line 634: The second mock response (the recovered 200) omits `headers`. This is fine for the success path because `fetchWithRetry` only calls `response.headers.get("Retry-After")` on transient responses before throwing `HttpRetryError` — it never accesses headers on an `ok: true` response. The test passes correctly as written.

Worth noting for anyone who adds a third leg (e.g. 503 → 503 → 200) to verify exhaustion behaviour: the intermediate non-final transient response would also need `headers: new Headers()`.  No action required now.

---

src/lib/auth/__tests__/refresh-google-token.test.ts
```diff
@@ -0,0 +1,111 @@
...
+      error: "invalid_grant",
+      error_description: "Token has been expired or revoked.",
+    };
+    mockFetch.mockResolvedValue(jsonResponse(errorBody, { status: 400 }));
+
+    await expect(
+      refreshGoogleAccessToken("revoked-rt", "cid", "sec")
+    ).rejects.toEqual(errorBody);
+    expect(mockFetch).toHaveBeenCalledTimes(1);
+  });
```
BenSeymourODB: Good coverage of the throw shape — this is exactly the contract the `session` callback's `catch` block depends on. One gap: the test only covers `{ status: 400 }`. The other non-transient 4xx that Google commonly returns here is `401` (e.g. after revocation when the grant is still technically "valid" by format). `401` is also non-transient per `isTransientHttpError`, so it should reach the `throw tokensOrError` path without retry — worth a second case to lock that in, especially since a misclassification there (treating 401 as transient and retrying it) would hammer Google with repeated requests against a revoked token.

---

BenSeymourODB (commented):
CI is green. Overall this is clean, well-scoped work. The `fetchWithRetry` wiring is correct in all four calendar routes and the new `refreshGoogleAccessToken` helper. The existing tests were updated correctly (fake timers + `headers: new Headers()` on transient mocks), the new transient-failure regression test exercises the actual retry path (`mockFetch` called twice, second call produces the 200), and the `AgendaList` tsc fix (`vi.fn().mockResolvedValue(undefined)`) matches the `(eventId, calendarId) => Promise<void>` signature exactly. No `any` casts, no manual memoization, no non-default Tailwind colors. Two findings, one of which needs a decision:

---

**Out-of-diff: `src/app/api/auth/refresh-token/route.ts` still uses bare `fetch` against `https://oauth2.googleapis.com/token`**

This is a second server-side route making the exact same Google OAuth token-refresh call that Phase 2 just extracted into `refreshGoogleAccessToken`. Issue #217's acceptance criterion is "All Google Calendar API server-side fetches go through `fetchWithRetry`" — this isn't a calendar route, but it calls the same OAuth endpoint that was explicitly in scope for the auth side. It's also the route that `src/lib/google-calendar.ts` calls client-side for token refresh, so transient 5xx from Google will surface as errors to end users on that path too.

The straightforward fix: replace the inline `fetch(...)` block with a call to `refreshGoogleAccessToken(refreshToken, clientId, clientSecret)` and map the result to the existing response shape. One side note while you're there: this route reads `process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID` rather than `GOOGLE_CLIENT_ID` — the `NEXT_PUBLIC_` prefix exposes the value to the browser bundle, which is probably unintentional for an OAuth client ID used in a server route (it works, but it leaks the value into the client JS).

If this route is intentionally out of scope for #217 (e.g. it's slated for removal once the Prisma-backed auth flow is the sole path), a comment in the file noting the deferral would prevent future reviewers from flagging it.

---

The two inline comments are a note-for-the-record on the missing `headers` in the second retry mock (no action needed) and a suggestion to add a `401` case to the throw-shape test for `refreshGoogleAccessToken`.

---

src/app/api/calendar/events/__tests__/route.test.ts
BenSeymourODB: Agreed — left as-is. Worth noting this was deliberate: the 200 mock isn't a `Response`-shaped object at all, just a duck-typed pair of `{ ok, json }` since `fetchWithRetry` doesn't touch headers on the success path. If we ever do add a `503 → 503 → 200` exhaustion test, the easiest path is probably a `jsonResponse(body, status)` helper that always sets `headers: new Headers()` so future readers don't have to re-derive when the field is needed (the `tasks-api` tests already use that pattern). Filed mentally as "next time we touch these tests."

---

src/lib/auth/__tests__/refresh-google-token.test.ts
BenSeymourODB: Good catch — added in 9089dfe. The new test mocks a 401 `invalid_client` body and asserts:

- `fetchWithRetry` is called exactly once (no retry — confirms `isTransientHttpError` keeps treating 401 as terminal)
- The thrown error is an instance of `GoogleTokenRefreshError`

I also took the opportunity to widen the throw shape from "the raw parsed body" to a structured `GoogleTokenRefreshError` carrying both `status` and `body`, so the new `/api/auth/refresh-token` migration (also in 9089dfe — see PR-level reply) can keep forwarding Google's HTTP status to the client. The 400 case was updated to assert `{ name, status, body }` against the structured throw.

---

BenSeymourODB:
Addressed in 9089dfe.

**`/api/auth/refresh-token` migration** — the route now calls `refreshGoogleAccessToken` instead of building the OAuth fetch inline. To keep its existing behaviour of forwarding Google's HTTP status to the client (and branching `invalid_grant` / `invalid_request` into a 401 with `requiresReauth: true`), the helper now throws a structured `GoogleTokenRefreshError` that carries both `status` and `body`. The session callback in `auth.ts` only logs and sets `session.error`, so the richer error type is purely additive there. No change to the success-path response shape that `lib/google-calendar.ts` consumes (`{ access_token, expires_in, scope, token_type }`).

**`NEXT_PUBLIC_GOOGLE_CLIENT_ID`** — flagging this as out-of-scope but real. `refresh-token/route.ts` (and a couple of other server-side OAuth helpers I noticed grepping around) read `NEXT_PUBLIC_GOOGLE_CLIENT_ID` instead of `GOOGLE_CLIENT_ID`, which leaks the client ID into the browser bundle even though the route only runs server-side. The session callback in `auth.ts` already uses the unprefixed `GOOGLE_CLIENT_ID`. Worth a follow-up to either standardise on the unprefixed env var server-side or document the intent — happy to file a separate issue if you'd like, didn't want to expand #217's scope without a heads-up.

**401 test gap** — added a 401 `invalid_client` case to `refresh-google-token.test.ts`; details in the inline thread.

**`headers` on second retry mock** — replied inline; no action taken.

---

Issue #224, opened 2026-05-02
tracking: deferred-PR-comment cleanup — dependency map and phasing


BenSeymourODB:
## Context

Audit of all merged PR comments (PR #19 → #192) to surface review items explicitly flagged as deferred / out-of-scope / follow-up. After de-duplicating against existing issues, 23 new issues were filed (#201-#223). This is the index + dependency map for the Project board.

Source data: 5 parallel scanner agents covered the 61 feature/fix PRs (dependency-bump PRs were skipped as low-signal).

### Calendar bugs (3)
- #201 — `getEventsForWeek` Saturday-afternoon boundary bug
- #202 — `MiniCalendarSidebar` viewMonth resync with selectedDate
- #203 — YearCalendar correctness bugs (timezone, multi-year overlap, perf, navigation)

### Calendar features (10)
- #204 — `useDateNow` hook with midnight-tick refresh
- #205 — Calendar helpers respect user-configurable `weekStartDay`
- #206 — Hourly time-grid with click-to-create in week/day
- #207 — Wire `AnimatedSwap` into Day/Week/Year + configurable duration
- #208 — Expand filter panel — persistence, per-calendar, hidden-events badge
- #209 — Wire day-overflow popover cards to `EventDetailModal`
- #210 — Runtime Zod validation for Google Calendar API responses
- #211 — `IEvent.category` field + group-by-category agenda mode
- #212 — Event dialog Phase 2 (reminders, recurrence, rich pickers)
- #213 — Replace `EventCreateDialog` placeholder user with profile assignment
- #214 — Calendar UX polish (scroll 7am, search count, sidebar visibility)

### Auth / HTTP (3)
- #215 — Automate OAuth token-key rotation + re-encryption
- #216 — Session timeout + concurrent token-refresh queue
- #217 — Wire `fetchWithRetry` into Google Calendar routes + auth refresh

### Infra (2)
- #218 — Rework IndexedDB cache (quota, per-event TTL, fallback)
- #219 — Standardize Prisma migration naming convention

### Cross-cutting (4)
- #220 — A11y umbrella (keyboard nav, ARIA, focus, WCAG audit)
- #221 — Scheduler stability + correctness bugs (PR #50 self-review)
- #222 — Test coverage backlog from PR feedback
- #223 — Per-profile streak tracking (post-Tasks)

## Dependency graph

```mermaid
graph LR
  subgraph "Independent (can start now)"
    A201[#201 Sat boundary]
[...]

PR #240, opened 2026-05-03
fix(calendar): include Saturday events in week view (#234)

BenSeymourODB:
Closes #234.

## Summary

`WeekCalendar` was reporting **0 events** and rendering an empty time grid
whenever events fell on the last day of the visible week (Saturday in
Sunday-first weeks). The bug was a boundary error in
`getEventsForWeek`: the upper bound was derived from
`getWeekDates(date)[6]`, which is **Saturday at 00:00:00 local** because
`getWeekDates` returns `addDays(startOfWeek(...), i)` and `startOfWeek`
sits on midnight. Consequently `eventStart <= endOfWeekDate` excluded
every Saturday event past `00:00`.

`getEventsForMonth` and `getEventsForYear` were already implemented
correctly via `endOfMonth`/`endOfYear` (end-of-day timestamps). This PR
brings `getEventsForWeek` in line by switching to
`endOfWeek(date, { weekStartsOn })`.

## Why this regressed in #176

PR #176 introduced the time-grid wiring and added `getEventsForWeek` as
the source of truth for `WeekCalendar`'s visible event set. Prior to
that, the week view filtered per-day on the rendered grid. The new
single-pass week filter inherited the off-by-one boundary because it
re-used the per-cell `getWeekDates` array rather than calling
`endOfWeek` directly.

## Changes

- **fix:** `src/lib/calendar-helpers.ts` — `getEventsForWeek` upper bound
  is now `endOfWeek(date, { weekStartsOn })` (Sat 23:59:59.999) instead
  of `getWeekDates(date)[6]` (Sat 00:00:00). No public API change.
- **unit:** `src/lib/__tests__/calendar-helpers.test.ts` — Saturday
  morning/afternoon/late-night events are included; an early-Sunday
  event of the next week is still excluded.
- **component:** `src/components/calendar/__tests__/WeekCalendar.test.tsx`
  — drives the bug end-to-end via the context provider on a Sat May 2
  2026 case. Event count reads `1 event` and the timed event renders.
- **E2E:** `e2e/week-day-views.spec.ts` — `page.clock.install` pins the
  date to Saturday before navigating; asserts the count chip is non-zero
  and boundary-day events are visible.
- **CI unblock:** `src/components/calendar/__tests__/AgendaList.test.tsx`
  — the mock context dropped `deleteEvent` when PR #197 added it to
  `ICalendarContext`, which has been failing `pnpm check-types` on main
  ever since (every PR's CI was red on `Lint and type check`). Mirrors
  the WeekCalendar mock pattern. Separate commit so it can be reverted
  independently.

## Test plan

- [x] `pnpm test` — 1575 pass (was 1573 before this PR; +2 new regression cases)
- [x] `pnpm lint && pnpm check-types` clean
- [x] `pnpm build` clean
- [x] New unit + component tests fail on main (`getEventsForWeek` returns
      `[]`, `WeekCalendar` shows `0 events`) and pass on this branch
- [x] No public API change — all existing callers continue to work

E2E was added but Playwright browsers couldn't be installed in the
sandbox; the spec compiles cleanly and follows the existing
`page.clock.install` pattern. CI does not run Playwright, so this is a
local/regression guard.

https://claude.ai/code/session_01CiWFv3xdKcrtMZq3Pc5Hy1


e2e/week-day-views.spec.ts
```diff
@@ -58,6 +58,31 @@ test.describe("Week Calendar", () => {
...
+  // Pin the system clock to a Saturday and verify the count and rendering.
+  test("renders events on the last day of the week when today is Saturday", async ({
+    page,
+  }) => {
+    // Sat May 2 2026 at 12:00 local. The mock event set anchors events to
+    // `new Date()` at module load, so installing the clock before navigation
+    // lands "today" on Saturday and produces a Saturday-boundary scenario.
+    await page.clock.install({ time: new Date("2026-05-02T12:00:00") });
+    await page.goto("/test/calendar?events=default&view=week");
+
```
BenSeymourODB, line 74: **Bug: `page.clock.install` cannot affect module-scope event dates — this test will be non-deterministic or silently wrong.**

`mockEventSets` in `src/app/test/calendar/page.tsx` is a top-level `const` that is evaluated when Next.js SSR processes the module on the server. `getRelativeDate()` calls `new Date()` at that evaluation time. By the time Playwright runs `page.clock.install({ time: new Date("2026-05-02T12:00:00") })`, the server-rendered HTML already has hardcoded ISO strings anchored to the real wall-clock time the server started — not to Saturday May 2.

`page.clock.install` only patches the browser's `Date` object inside the Chromium tab. It has no effect on Node.js (the Next.js server process). So:

- If the test runs on a Saturday, the events coincidentally land on Saturday and the test passes for the wrong reason.
- On any other day of the week the events land on that day, not Saturday, so the count chip will read 0 or some non-Saturday count, and `expect(text).toMatch(/[1-9]\d* events?/)` may still pass (events exist elsewhere in the week), but `Morning Standup` and `Project Review` will not be in the Saturday column — the assertions on `getByText` will pass only because those text nodes exist somewhere in the DOM.

The fix is to make the event dates truly dynamic from the client. Options:
1. Compute `mockEventSets` inside `TestCalendarContent` (which runs on the client after hydration) so `new Date()` is called in the browser where `page.clock.install` has effect.
2. Accept an explicit `anchor` URL param (`?anchor=2026-05-02`) and pass it through `getRelativeDate` instead of calling `new Date()`.

Either approach makes the clock-pinning meaningful. As-is the test provides false assurance — it does not exercise the boundary case it claims to cover unless run on a Saturday.

---

src/lib/__tests__/calendar-helpers.test.ts
```diff
@@ -354,6 +354,54 @@ describe("getEventsForWeek", () => {
...
+      "sat-late-night",
+      "sat-morning",
+    ]);
+  });
+
+  it("excludes events that start the day after the week ends", () => {
+    // Sun May 3, 2026 is the first day of the next week.
+    const referenceDate = new Date(2026, 4, 2); // selectedDate sits on Saturday
+    const events: IEvent[] = [
+      createMockEvent({
```
BenSeymourODB: **Bug: `saturday.setHours()` mutates the shared `Date` object — event timestamps are not what they appear.**

`Date.prototype.setHours` mutates in-place and returns the new timestamp as a number. `new Date(saturday.setHours(9, 0, 0, 0))` sets `saturday` to 09:00, uses that value for `sat-morning`'s `startDate`, then immediately calls `saturday.setHours(10, 0, 0, 0)` — which mutates `saturday` again and returns the 10:00 timestamp used as `endDate`. Fine so far.

Then the next event calls `saturday.setHours(14, 0, 0, 0)` — but at this point `saturday` is already `10:00` (mutated by the previous `endDate` construction), so `sat-afternoon.startDate` is set to 14:00 and `saturday` is now at 14:00 going into the next call. Same pattern for `sat-late-night`.

Accidentally the `id`-based sort in the assertion (`sort()`) masks this: all three events do end up with some Saturday timestamp, so the test passes. But if you ever assert on the actual times (e.g. "morning event starts at 09:00"), the values will be wrong.

Fix: create a fresh `Date` for each event or use a helper:
```ts
const sat = (h: number, m = 0) => new Date(2026, 4, 2, h, m, 0, 0).toISOString();

createMockEvent({ id: "sat-morning",    startDate: sat(9),     endDate: sat(10) }),
createMockEvent({ id: "sat-afternoon",  startDate: sat(14),    endDate: sat(15) }),
createMockEvent({ id: "sat-late-night", startDate: sat(23, 30), endDate: sat(23, 59) }),
```
This is non-mutating and the intent is immediately readable.

---

src/lib/calendar-helpers.ts
```diff
@@ -354,9 +354,8 @@ export const getWeekDates = (date: Date): Date[] => {
 };
 
 export const getEventsForWeek = (events: IEvent[], date: Date): IEvent[] => {
-  const weekDates = getWeekDates(date);
-  const startOfWeekDate = weekDates[0];
-  const endOfWeekDate = weekDates[6];
+  const startOfWeekDate = startOfWeek(date, { weekStartsOn: WEEK_STARTS_ON });
```
BenSeymourODB, line 357: **Nit / future-proofing: fix is correct but still hard-codes `WEEK_STARTS_ON`.**

The fix is definitely right — `endOfWeek(date, { weekStartsOn: WEEK_STARTS_ON })` is the same pattern used by `rangeText` and `WeekCalendar.tsx` itself, so this is consistent with the existing codebase.

Worth noting for the tracker: once issue #205 / PR #229 threads `weekStartDay` through from `CalendarProvider`, `getEventsForWeek` will need a `weekStartsOn` parameter to match the user's preference — a Monday-first user hitting a Sunday boundary will have the same off-by-one for Sunday that a Sunday-first user was having for Saturday. The fix as-is is correct for the default and un-breaks the regression; it just isn't the final form once per-user week-start is live.

No action required on this PR — just flagging so the follow-up issue scope is clear.

---

BenSeymourODB (commented):
## Review summary

**Core fix** (`calendar-helpers.ts`) is correct and minimal. `endOfWeek(date, { weekStartsOn: WEEK_STARTS_ON })` produces `Sat 23:59:59.999`, which is exactly what `getEventsForMonth`/`getEventsForYear` do with their own boundaries. The three-line change is the right change.

**Unit tests** (`calendar-helpers.test.ts`) do exercise the real bug path and would fail on `main`. One reliability issue: `saturday.setHours()` mutates the shared Date object across all three event constructions (see inline comment) — the tests pass today because the assertion sorts by id, but the actual timestamps aren't what the variable names suggest.

**Component test** (`WeekCalendar.test.tsx`) is solid. It wires the full context + component path, correctly derives Saturday from `startOfWeek + addDays(6)` without mutation, and asserts both the count chip and the rendered title. This would fail on main and pass on this branch — good regression coverage.

**AgendaList CI unblock** is safe. `deleteEvent` was simply absent from the mock object after PR #197 added it to `ICalendarContext`. No existing test was asserting on `deleteEvent` calls, so adding `vi.fn().mockResolvedValue(undefined)` is the right pattern (mirrors `WeekCalendar.test.tsx`).

**E2E spec** has a structural problem (see inline comment): `mockEventSets` is evaluated at module scope on the server, so `page.clock.install` in the browser cannot affect those timestamps. The spec will not reliably exercise the Saturday boundary unless run on an actual Saturday, which makes it a false-confidence guard rather than a genuine regression test. Recommend either moving event construction into `TestCalendarContent` (client-side) or adding an `?anchor=` URL param before landing this.

**No CLAUDE.md violations**: no manual memoization, no `any` types, standard Tailwind palette only, TDD order (tests added alongside fix).

---

src/lib/calendar-helpers.ts
BenSeymourODB, line 357: **Fix correctness — looks good, but `weekStartsOn` coupling is incomplete (pre-existing, not new)**

The fix is correct for the reported bug: `endOfWeek(date, { weekStartsOn: WEEK_STARTS_ON })` returns `Saturday 23:59:59.999` (local time) which is consistent with how `getEventsForMonth` uses `endOfMonth` and `getEventsForYear` uses `endOfYear`. DST and timezone boundaries are handled correctly because `date-fns` `endOfWeek` always anchors to the local-time end of the final day regardless of UTC offset.

**Pre-existing coupling issue (not introduced here):** `getEventsForWeek` hardcodes `WEEK_STARTS_ON` (the constant `0`) instead of accepting a `weekStartsOn` parameter from the caller. `WeekCalendar.tsx` also hardcodes `WEEK_STARTS_ON` for display. When issue #86 eventually wires `weekStartDay` from `CalendarProvider` through to the view, this function will need a `weekStartsOn` parameter added. Worth tracking so the fix from this PR doesn't get overlooked during that work. Not a blocker for this PR.

---

src/lib/__tests__/calendar-helpers.test.ts
```diff
@@ -354,6 +354,54 @@ describe("getEventsForWeek", () => {
     const result = getEventsForWeek(events, new Date(2024, 2, 13));
     expect(result.length).toBe(2);
   });
+
+  // Regression: #234 — events that start on the last day of the week (Saturday
+  // when WEEK_STARTS_ON = 0) at any time after 00:00 were being filtered out
+  // because the upper bound was `weekDates[6]`, i.e. Saturday at midnight.
+  it("includes events that start on the last day of the week", () => {
+    // Sat May 2, 2026 is the last day of the Apr 26 – May 2 Sunday-first week.
+    const saturday = new Date(2026, 4, 2);
```
BenSeymourODB: **Nit: `saturday` is mutated in-place across all three events**

`saturday.setHours(...)` mutates the single `saturday` object rather than creating a fresh `Date` for each event. The chain is safe here because each `setHours` call stays on May 2 and returns a valid timestamp, but the pattern is fragile — a future maintainer adding a fourth event that crosses midnight (e.g. `setHours(24, ...)`) would silently roll to May 3 and produce a confusing test failure.

Consider:
```ts
const satAt = (h: number, m = 0) => new Date(2026, 4, 2, h, m, 0, 0);
createMockEvent({ id: "sat-morning",   startDate: satAt(9).toISOString(),  endDate: satAt(10).toISOString() }),
createMockEvent({ id: "sat-afternoon", startDate: satAt(14).toISOString(), endDate: satAt(15).toISOString() }),
createMockEvent({ id: "sat-late-night",startDate: satAt(23,30).toISOString(), endDate: satAt(23,59).toISOString() }),
```
Not a correctness bug in the current PR, just a readability/resilience nit.

---

e2e/week-day-views.spec.ts
```diff
@@ -58,6 +58,31 @@ test.describe("Week Calendar", () => {
...
+  // (Saturday in Sunday-first weeks), `getEventsForWeek` was filtering every
+  // boundary-day event out because the upper bound was Saturday at 00:00.
+  // Pin the system clock to a Saturday and verify the count and rendering.
+  test("renders events on the last day of the week when today is Saturday", async ({
+    page,
+  }) => {
+    // Sat May 2 2026 at 12:00 local. The mock event set anchors events to
+    // `new Date()` at module load, so installing the clock before navigation
+    // lands "today" on Saturday and produces a Saturday-boundary scenario.
+    await page.clock.install({ time: new Date("2026-05-02T12:00:00") });
```
BenSeymourODB: **`page.clock.install` may not reliably pin `mockEventSets` — SSR/client split**

`mockEventSets` is built at **module scope** in `src/app/test/calendar/page.tsx` (a `"use client"` component). In Next.js, `"use client"` modules are **still server-rendered** during the SSR pass. That means the server evaluates `getRelativeDate(0, ...)` → `new Date()` before `page.clock.install` has any effect. The SSR HTML will embed events anchored to the real server date (e.g. Sunday May 3, 2026 in today's environment).

When the browser receives that HTML:
- React hydrates against the SSR output.
- The JS bundle is evaluated client-side with the mocked clock, rebuilding `mockEventSets` anchored to Saturday May 2.
- This creates a **hydration mismatch** — dates in the server HTML differ from the client's `new Date()`. React 19 logs a hydration error and may fall back to a full client re-render, or it may silently keep the SSR data.

The practical risk is that the test may pass or fail depending on whether React ends up using the SSR-anchored events (no Saturday events in the week) or the client-anchored events (Saturday events present). This makes the test non-deterministic in a CI environment where the real date is not a Saturday.

**Suggested fix:** Add a dedicated URL param (e.g. `?date=2026-05-02`) to `TestCalendarPage` that sets `initialDate` on `MockCalendarProvider` and computes `mockEventSets` relative to that explicit date instead of `new Date()`. That way the test becomes `page.goto('/test/calendar?events=default&view=week&date=2026-05-02')` with no clock manipulation needed. Alternatively, move `mockEventSets` into the component body so it's computed each render (after hydration with the mocked clock).

---

BenSeymourODB (commented):
Overall: the core fix is correct and well-tested at the unit and component levels. Three comments below — one informational, one nit, one substantive concern about the E2E spec.

**What's solid:**
- `endOfWeek(date, { weekStartsOn })` is exactly the right fix. It is consistent with how `getEventsForMonth` and `getEventsForYear` handle their respective boundaries (using `endOfMonth`/`endOfYear`), and it is DST-safe because `date-fns` anchors to local-time end-of-day regardless of UTC offset.
- The two new unit tests in `calendar-helpers.test.ts` are not tautological — they would fail on `main` (old upper bound was Saturday 00:00) and pass after the fix.
- The component test in `WeekCalendar.test.tsx` correctly drives the fix through the full render path with a Saturday `selectedDate`.
- The `AgendaList.test.tsx` CI-unblock is safe: `AgendaList` does not call `deleteEvent`, so adding the mock just satisfies the type shape. No meaningful assertions were being silently masked. The pattern is identical to every other mock context file in the test suite.
- CLAUDE.md rules are respected: no manual memoization, standard Tailwind palette, no `any` types, TDD followed (tests added before/with implementation), separate commits for fix vs CI-unblock vs E2E.

**Substantive issue (E2E spec):** see the inline comment on `e2e/week-day-views.spec.ts` line 72 — `page.clock.install` may not reliably control `mockEventSets` due to the SSR/client split in the test page.

**Nit (unit test):** see `calendar-helpers.test.ts` line 363 — `saturday` object is mutated in-place across three events; safe today but fragile for future maintenance.

**Informational (pre-existing, not a blocker):** `getEventsForWeek` hardcodes `WEEK_STARTS_ON` rather than accepting a `weekStartsOn` parameter. This is pre-existing (#86) but worth noting so the issue doesn't get missed when `weekStartDay` from context is eventually threaded through — see the inline note on `calendar-helpers.ts` line 357.

---

e2e/week-day-views.spec.ts
BenSeymourODB: Good catch — the SSR/client split was a real hole. Fixed in 8afa852.

The test page now accepts `?anchor=YYYY-MM-DD`. `mockEventSets` is built by a `buildMockEventSets(anchor)` factory called inside `TestCalendarContent`, so server and client compute the same deterministic timestamps from the URL — no `page.clock.install`, no hydration mismatch. The same `anchor` is also threaded as `MockCalendarProvider`'s `initialDate` so the calendar lands on Saturday, not just the events.

The E2E now uses `?anchor=2026-05-02` and scopes the title assertions to the Sat 2026-05-02 day column (`week-calendar-day-col-2026-05-02`) instead of letting `getByText` match anywhere in the DOM, which was the second part of the false-confidence problem you flagged.

---

src/lib/__tests__/calendar-helpers.test.ts
BenSeymourODB: Fixed in 8afa852. Replaced the mutating `saturday.setHours(...)` chain with a `sat(h, m = 0)` helper that always allocates a fresh `Date` (per your suggestion). Also added explicit timestamp assertions (`getHours()` checks per event id) so the next time someone trips over the same pattern, the test will fail on the actual values, not just the membership set.

---

src/lib/calendar-helpers.ts
BenSeymourODB: Acknowledged — leaving this PR scoped to the regression fix as you suggested. The `weekStartsOn` parameterisation belongs with the broader #86/#205 work (and PR #229 is already in flight for the helpers). Once that lands, both `getEventsForWeek` and `WeekCalendar.tsx`'s direct `WEEK_STARTS_ON` references will need to thread through the user's `weekStartDay`. Tracking under #224 alongside the other deferred-comment work.

---

BenSeymourODB: Same response as the earlier thread on this line — see https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/240#discussion_r3177737429. Holding the `weekStartsOn` parameter change for the broader #86/#205 work; this PR stays scoped to the regression.

---

Issue #150, opened 2026-04-21
Make Agenda an optional toggle within Day and Week views (not a top-level view)


BenSeymourODB:
## Context

Today, Agenda is a **peer** of Day / Week / Month / Year in `TCalendarView` (`src/types/calendar.ts:5`) and `ViewSwitcher` (`src/components/calendar/ViewSwitcher.tsx`). Users pick a view and render either a time-grid or a separate agenda list.

The better model — as used by the Windows 11 / Microsoft Teams Calendar widget (see screenshots on the original task) — is to keep **Day** and **Week** as the two primary time-range views, and treat **Agenda** as an *alternate rendering mode* available inside each of them. When Agenda is on, empty time gaps between events are collapsed and events are laid out as a compact chronological list for the same date range the user is already viewing. When Agenda is off, the familiar hourly time-grid is shown.

### Reference UX (Windows / Teams Calendar widget)

- Single "Day ▾" dropdown that offers **Agenda** and **Day** as siblings.
- Switching to Agenda keeps the selected date and the prev/next controls — only the rendering changes.
- No empty 8 AM → 11 AM blank space; instead, events are listed with their start time, duration, and "Starts in X mins" affordances.
- Past events fade; the next event is visually highlighted.

## Goal

- Remove `"agenda"` as a top-level `TCalendarView` entry.
- Introduce an `agendaMode: boolean` (or `displayMode: "grid" | "agenda"`) that applies **only** when the active view is `day` or `week`.
- Expose the toggle from the same control that picks the view (matching the Teams dropdown), so the user sees e.g. `Day ▾ → { Agenda, Day }` and `Week ▾ → { Agenda, Week }`.
- In agenda mode, collapse time gaps and render events as a chronological list for the current date (Day) or the current week (Week), grouped by day with sticky date headers when helpful.

## Non-goals

- Changing Month or Year views.
- Re-designing the existing agenda item card — the current `AgendaCalendar` event card is a fine starting point and can be lifted into the new mode.
- Adding search/filter UI inside agenda mode (tracked by #91 / PR #144 — the two should compose, but are separate slices).

## Acceptance criteria

- [ ] `TCalendarView` no longer includes `"agenda"`. A new `agendaMode` (or equivalent) flag is added to `CalendarProvider` and persisted to localStorage alongside `view`.
[...]

PR #241, opened 2026-05-03
fix(calendar): honour ?view= URL param on production /calendar page (#238)

### LINKED ISSUE #238: /calendar?view=year (and other view= URL params) ignored on production page
**Discovered during:** [QA regression sweep 2026-05-02 (Phase C)](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/claude/qa-regression-plan-2026-05-02/docs/screenshots/qa-regression-2026-05-02/RESULTS-PHASE-C.md) — Finding 6.

## Reproduce

1. Sign in.
2. Visit `http://localhost:3000/calendar?view=year` (or `?view=week`, `?view=day`, etc.).

## Observed

Page loads with **Month** view active (highlighted in ViewSwitcher), not Year. The `view=year` URL param is silently dropped. Same for other `view=` values.

The test page (`/test/calendar?view=year`) correctly honors the URL param.

## Likely cause

`/calendar/page.tsx` initializes the view from `CalendarProvider`'s persisted state (probably localStorage / cookie) and never reads the `view` URL search param. The test page works because it explicitly reads `useSearchParams()` (`src/app/test/calendar/page.tsx` line ~447), but the production page does not.

## Severity

**Medium.** Bookmark/sharing of view-specific URLs on `/calendar` doesn't work as expected. Also blocks autonomous QA of view-specific features (e.g. #196 year view) since we can't deep-link in.

## Suggested next step

In `src/app/calendar/page.tsx` (or whatever wraps `CalendarProvider` for the live route), read the `view` search param and pass it through as the provider's initial view. Mirror the test page pattern.

## Evidence

`docs/screenshots/qa-regression-2026-05-02/C196.2-year-live.png` (URL is `/calendar?view=year` but Month is highlighted in ViewSwitcher).

BenSeymourODB:
Closes #238.

## Summary

The production `/calendar` page silently dropped `?view=year` (and every other `view=` URL param) — it always landed on whatever view localStorage held, never reading the URL. The test page at `/test/calendar` already honoured the param via `useSearchParams`, so this PR mirrors that pattern on the live route and gives `CalendarProvider` an explicit one-shot override that wins over the persisted value.

## Why localStorage was winning

`CalendarProvider`'s existing `view` prop only seeded `useLocalStorage`'s _default_ value. Once the user had ever interacted with the calendar, localStorage held a value and the prop was ignored. So passing `view={urlView}` to the existing prop wouldn't have overridden anything — a deep link to `/calendar?view=year` for a returning user would have stayed on Month.

The fix introduces a separate `initialView` prop that's explicitly an override:

- Reads from the prop for the initial `useState` value (wins over localStorage on first render — no FOUC of the wrong view).
- A mount-only effect writes the override through to localStorage, so the deep-linked choice persists across reloads.
- Companion `initialAgendaMode` handles the legacy `?view=agenda` deep link, which maps to `view=day` + `agendaMode=true` — same migration semantics as the provider's existing pre-#150 → post-#150 storage migration.
- When both props are `undefined`, behaviour is unchanged (regression-tested).

## Changes

- `src/components/providers/CalendarProvider.tsx` — add `initialView` and `initialAgendaMode` props; seed `useState` from them; one-shot mount effect persists the override to localStorage.
- `src/app/calendar/page.tsx` — wrap in `<Suspense>` (required for `useSearchParams`), read `view` / `agendaMode`, validate against the known set (`day`, `week`, `month`, `year`, `clock`, `agenda`), map legacy `agenda` → `day` + `agendaMode=true`, forward overrides to the provider.
- `src/components/providers/__tests__/CalendarProvider.test.tsx` — 5 new cases under "initialView / initialAgendaMode overrides": seeds default, overrides localStorage, writes through to localStorage, agenda companion, undefined-passthrough regression.
- `src/app/calendar/__tests__/page.test.tsx` — new "URL ?view= deep linking (#238)" describe with 10 cases (one per known view via `it.each`, plus legacy agenda, agendaMode pass-through, unknown-param fallthrough, no-param fallthrough). Mocks `useSearchParams` and asserts the props the page hands to `CalendarProvider`.
- `src/components/calendar/__tests__/AgendaList.test.tsx` — adds the missing `deleteEvent` field to the mock context. This has been failing `pnpm check-types` on `main` since #197 landed (already noted as a "CI unblock" commit in #240); included here so this branch's CI passes. If #240 lands first, this hunk merges trivially.

## Why no Playwright E2E

The production `/calendar` page is auth-gated (`useSession`) and there's no shared E2E auth fixture in `e2e/`. The component-level coverage validates the wiring shape end-to-end at the level the bug actually exists: the page reads search params and the provider applies them as initial state. The existing `/test/calendar?view=year` page already exercises the same `useSearchParams` pattern via E2E. If we add an auth fixture later, an E2E that visits `/calendar?view=year` and asserts the visible view becomes Year would slot straight in.

## Test plan

- [x] `pnpm test` — 1587 pass (was 1572 before; +15 new cases — 5 provider + 10 page)
- [x] `pnpm lint` — 0 errors (5 pre-existing warnings on unrelated files)
- [x] New tests fail on main (initialView prop doesn't exist; page never reads search params) and pass on this branch.
- [ ] Manual smoke: load `/calendar?view=year` while signed in → year view active and persists after dropping the query string.

https://claude.ai/code/session_01D7mQWzLez1FdrMFAV6nHtj

---
_Generated by [Claude Code](https://claude.ai/code/session_01D7mQWzLez1FdrMFAV6nHtj)_


BenSeymourODB (commented):
Reviewed the full diff against the actual source. The intent is correct and the wiring is solid. A few things worth addressing before merge, ranging from a real correctness gap down to a missing regression test.

---

## 1. Race condition: legacy-migration effect fires *after* the override effect on the same mount — wrong order (medium risk)

**File:** `src/components/providers/CalendarProvider.tsx`, lines 186–193 vs 218–236

The legacy-agenda migration effect (lines 186–193) and the new deep-link override effect (lines 218–236) both fire in the same React commit, in declaration order. That means if localStorage contains `view: "agenda"` **and** `initialView` is something else (e.g. `?view=year`), the sequence is:

1. `useState` seeds `currentView = initialView` ("year") — correct.
2. Deep-link effect fires: writes `view: "year"` to localStorage — correct.
3. Legacy-migration effect fires: `rawView` is `"agenda"` (the snapshot taken before any effect), so it fires `setSettings(migrateLegacySettings(...))` — this re-reads localStorage (which now contains "year" from step 2) and spreads it, **but** it also sets `agendaMode: true` because the functional updater receives the `prev` that was stale in `useLocalStorage.setValue` at the moment it was called (the hook reads localStorage directly, so it now sees "year"). The migration function sees `prev.view === "agenda"` is false, so it returns the spread as-is. On the actual stored value this is probably safe — but `rawView` was captured from the pre-effect render snapshot.

The **real** fragility: the two effects fire in the same micro-task batch, and `useLocalStorage.setValue` reads `window.localStorage.getItem` directly (not the React-state snapshot), so they compose correctly *in practice*. However if the order ever changes (e.g. StrictMode double-invoke, or these effects are refactored closer together), the migration effect could clobber the override. The safe fix is to add a guard in the migration effect so it no-ops when `initialView` is defined:

```ts
useEffect(() => {
  if (rawView === "agenda" && initialView === undefined) {
    setSettings(prev => migrateLegacySettings(prev) as unknown as LegacyCalendarSettings);
  }
}, [rawView, setSettings]);   // initialView is stable (mount prop), so adding it here costs nothing
```

---

## 2. `?agendaMode=true` with no `?view=` param sets `initialAgendaMode` but leaves `initialView` undefined — `useState` seeds the *wrong* agendaMode (low-medium)

**File:** `src/app/calendar/page.tsx`, lines 150–154

```ts
const initialAgendaMode = isLegacyAgenda
  ? true
  : searchParams.get("agendaMode") === "true"
    ? true
    : undefined;
```

If a user hits `/calendar?agendaMode=true` (no `view=`), `initialView` is `undefined` and `initialAgendaMode` is `true`. Inside the provider:

```ts
const [agendaMode, setAgendaModeState] = useState<boolean>(
  initialAgendaMode ?? settings.agendaMode ?? DEFAULT_SETTINGS.agendaMode
);
```

That seeds `agendaMode = true` immediately — correct. But there is no `initialView` override, so `currentView` is seeded from localStorage. If localStorage says `view: "month"`, the user lands on Month view with `agendaMode: true` in state, which is a meaningless combination (agendaMode has no effect on month). On a reload (after the write-through effect has fired) the URL no longer carries `?agendaMode=true`, so localStorage will hold `agendaMode: true` forever even though the user never explicitly requested it in a meaningful context.

This is a corner case (the URL is a strange one to deep-link), but it is a silent state corruption. The simplest guard: only forward `initialAgendaMode` when `initialView` is also present (or is already day/week from localStorage):

```ts
const initialAgendaMode =
  (isLegacyAgenda || searchParams.get("agendaMode") === "true") &&
  initialView !== undefined   // only meaningful when we also know the view
    ? true
    : undefined;
```

There is no test for this scenario — adding one would catch it.

---

## 3. Missing regression test: `?agendaMode=true` alone (no `view=`) should not corrupt localStorage agendaMode

**File:** `src/app/calendar/__tests__/page.test.tsx`

The existing test for `?agendaMode=true` (line 201) always pairs it with `view=week`. There is no test for `?agendaMode=true` without a `view=` param. Given finding #2 above, this is the exact scenario that would silently corrupt state and no test would catch a regression.

Suggested addition:

```ts
it("ignores ?agendaMode=true when no view param is present, leaving initialAgendaMode undefined", () => {
  setView("month");
  setSearchParams("agendaMode=true");

  render(<CalendarPage />);

  const props = calendarProviderProps.mock.calls.at(-1)?.[0] as {
    initialView?: TCalendarView;
    initialAgendaMode?: boolean;
  };
  expect(props.initialView).toBeUndefined();
  expect(props.initialAgendaMode).toBeUndefined();
});
```

---

## 4. `isViewParam` type-guard allows `"agenda"` through to `TCalendarView` cast — requires the `as TCalendarView` cast to remain correct (minor, correctness observation)

**File:** `src/app/calendar/page.tsx`, lines 26–40

`VIEW_PARAM_VALUES` includes `"agenda"`, which is **not** in `TCalendarView`. The guard `isViewParam(rawView)` returns `true` for `"agenda"`, and then the code correctly intercepts it with `isLegacyAgenda ? "day" : (rawView as TCalendarView)`. This works, but the `as TCalendarView` cast is only safe because the `isLegacyAgenda` branch catches the one non-`TCalendarView` member first.

If `VIEW_PARAM_VALUES` ever grows another non-`TCalendarView` entry (e.g. `"list"` for a future view that's added to the URL param set before it reaches the provider), the cast becomes unsound silently. A tighter approach that removes the cast entirely:

```ts
const NON_VIEW_PARAMS = new Set<string>(["agenda"]);

const initialView: TCalendarView | undefined =
  isViewParam(rawView) && !NON_VIEW_PARAMS.has(rawView)
    ? (rawView as TCalendarView)  // narrowed — only TCalendarView members reach here
    : isLegacyAgenda
      ? "day"
      : undefined;
```

Or, alternatively, split `VIEW_PARAM_VALUES` into two `as const` arrays — one that is exactly `TCalendarView` members, one for legacy aliases — and derive the type guard from the canonical set only. This is a low-priority refactor, but worth noting before the next param is added.

---

## 5. No test exercises the double-effect ordering (mount with legacy localStorage + initialView set simultaneously)

**File:** `src/components/providers/__tests__/CalendarProvider.test.tsx`

There is a test for "overrides a persisted localStorage view when initialView is provided" (line 325), but the persisted value is `"month"`, not `"agenda"`. The race condition in finding #1 only manifests when the persisted value is `"agenda"` AND `initialView` is defined. Worth adding:

```ts
it("does not let the legacy-agenda migration clobber an initialView override", async () => {
  window.localStorage.setItem(
    "calendar-settings",
    JSON.stringify({ badgeVariant: "colored", view: "agenda", use24HourFormat: true,
                     agendaModeGroupBy: "date", weekStartDay: 0 })
  );

  renderWithOverrides({ initialView: "year" });

  await waitFor(() => {
    expect(screen.getByTestId("view")).toHaveTextContent("year");
  });
  await waitFor(() => {
    const parsed = JSON.parse(window.localStorage.getItem("calendar-settings") ?? "{}");
    expect(parsed.view).toBe("year");
  });
});
```

---

## What looks good

- The `Suspense` boundary placement is correct — `useSearchParams()` requires it in the App Router and wrapping only `CalendarPageBody` (not the entire page) is the right granularity.
- The `useState` initialization from `initialView ?? settings.view ?? DEFAULT_SETTINGS.view` is correct: it eliminates the FOUC of the wrong view on first render without needing a layout effect.
- The mount-only `// eslint-disable-next-line react-hooks/exhaustive-deps` with a clear comment explaining the intentional pattern is acceptable here.
- The `migrateLegacySettings` call inside the write-through effect's functional updater is the right place to call it (ensures the write always lands a fully-migrated payload).
- The `as unknown as LegacyCalendarSettings` cast in the write-through effect is consistent with existing usage in the file.
- The `AgendaList.test.tsx` CI-unblock addition is correct and the mock shape matches `ICalendarContext.deleteEvent`.
- CLAUDE.md compliance: standard Tailwind classes only in the fallback div, no `useMemo`/`useCallback`/`React.memo` added, no `any` types, pnpm only.

---

BenSeymourODB (commented):
## Review — PR #241 `fix(calendar): honour ?view= URL param on production /calendar page`

Overall the approach is sound and the implementation is largely correct. CI is green (+15 tests all passing). A few issues below, ordered by severity.

---

### 1. Silent ordering dependency between the two `setSettings` effects — moderate risk

`CalendarProvider.tsx` lines 186–193 and 218–236

Both the legacy-migration effect (`rawView === "agenda"` guard) and the new override effect call `setSettings` on the same render cycle. React fires `useEffect` in declaration order within a single component, so migration runs first and the override runs second — the correct order.

The problem: this dependency is structural and silent. If either block is moved during a refactor (e.g. someone reorders them while tidying up, or extracts the migration to a custom hook), the scenario where `localStorage` holds `view: "agenda"` **and** `initialView` is provided would silently invert: migration now fires second, unconditionally writes `{ view: "day" }`, and clobbers the override.

The `useLocalStorage` setter reads directly from `localStorage` inside the updater, so both writes land synchronously in source order, but the _second_ write wins. Concretely: a bookmark to `?view=year` while the user's `localStorage` still holds the legacy `"agenda"` value would land on `day` instead of `year`.

**Suggestion**: add a comment directly above the override `useEffect` that names the dependency:

```ts
// NOTE: this effect must come after the rawView === "agenda" migration
// effect above (lines 186-193). Both call setSettings on mount; React
// fires useEffect in declaration order, so migration wins on the first
// write and this override wins on the second. Moving this block above
// the migration block would cause it to be clobbered when legacy
// localStorage is present.
useEffect(() => {
```

Alternatively — and more robustly — fold the override into the migration effect so there is only one `setSettings` call for the initial-mount writes:

```ts
useEffect(() => {
  const needsMigration = rawView === "agenda";
  const needsOverride = initialView !== undefined || initialAgendaMode !== undefined;
  if (!needsMigration && !needsOverride) return;
  setSettings((prev) => {
    const migrated = migrateLegacySettings(prev) as Partial<CalendarSettings>;
    return {
      ...migrated,
      ...(initialView !== undefined ? { view: initialView } : {}),
      ...(initialAgendaMode !== undefined ? { agendaMode: initialAgendaMode } : {}),
    } as unknown as LegacyCalendarSettings;
  });
}, []);
```

This is a single atomic write and removes the ordering coupling entirely.

---

### 2. `isViewParam` typeguard: `"agenda"` is in `VIEW_PARAM_VALUES` but not in `TCalendarView` — the cast is load-bearing and fragile

`src/app/calendar/page.tsx` lines 26–39, 145–148

`TCalendarView = "day" | "week" | "month" | "year" | "clock"` — `"agenda"` is intentionally absent (post-#150).

`VIEW_PARAM_VALUES` includes `"agenda"`, and `isViewParam` returns `value is TViewParam`. The subsequent cast `rawView as TCalendarView` at line 148 is only safe because the `isLegacyAgenda` branch above it has already excluded `"agenda"`. TypeScript accepts the cast because `TViewParam` is a supertype of `TCalendarView` and the narrowing is manual.

This is correct today, but it's one easy mistake away from a type hole: if someone adds a new legacy value to `VIEW_PARAM_VALUES` and forgets to add a corresponding exclusion branch, the cast lets a non-`TCalendarView` value flow into `initialView`. 

**Suggestion**: make the mapping exhaustive and explicit rather than relying on a cast. Either:

```ts
const LEGACY_TO_VIEW: Record<string, { view: TCalendarView; agendaMode?: true }> = {
  day:   { view: "day" },
  week:  { view: "week" },
  month: { view: "month" },
  year:  { view: "year" },
  clock: { view: "clock" },
  agenda: { view: "day", agendaMode: true },
};
```

Or at minimum, replace `rawView as TCalendarView` with a satisfies-guarded lookup that avoids the cast. This is a minor hardening concern, not a bug in the current code.

---

### 3. `initialAgendaMode=true` without a `view` param overrides localStorage's `agendaMode` but leaves `currentView` unchanged — intentional but untested

`src/components/providers/CalendarProvider.tsx` line 204 and `src/app/calendar/page.tsx` line 150–154

The page only forwards `initialAgendaMode` when either `isLegacyAgenda` is true (in which case `initialView` is also set to `"day"`) or `?agendaMode=true` is explicit in the URL. In the second case (`?agendaMode=true` without `?view=`), `initialView` will be `undefined`, so `currentView` is seeded from localStorage. The agenda-mode toggle will be overridden but the view stays wherever localStorage left it. That may or may not be the desired UX, but there is no test for this path.

The existing test "forwards ?agendaMode=true alongside a non-legacy view" only tests `?view=week&agendaMode=true` (both params present). There is no test for `?agendaMode=true` alone (no `view` param). If the intent is that `?agendaMode=true` on its own is a valid, intentional URL, add a test; if it's a don't-care case, a comment saying so would help.

---

### 4. The `as unknown as LegacyCalendarSettings` double-cast in the override effect — minor

`src/components/providers/CalendarProvider.tsx` line 230

The same pattern is used in the migration effect (line 190) and is pre-existing. The type cast is necessary because `useLocalStorage` is typed as `LegacyCalendarSettings` but the spread produces `Partial<CalendarSettings>`. This is fine; just noting it propagates the existing workaround rather than fixing it.

---

### 5. Tests: the `it.each` for known views omits `"clock"` in the parametrised row but the format is correct — minor

`src/app/calendar/__tests__/page.test.tsx` line 172

Wait — actually `"clock"` IS present: `[["day"], ["week"], ["month"], ["year"], ["clock"]]`. But `"year"` is also tested individually in the preceding `it(...)` block. The duplication is harmless but the individual `it` case for `"year"` is now redundant given the `it.each` covers it. Not a correctness issue.

---

### 6. CLAUDE.md compliance — clean

- No custom Tailwind palette — `bg-background` and `min-h-screen` are standard Tailwind/CSS-variable tokens used in the existing page, the fallback `<div>` matches the real content div.
- No manual memoization added.
- No `any` types introduced.
- No test artifacts committed.
- `agendaMode` field in `AgendaList.test.tsx` — the `mockResolvedValue(undefined)` is correct for `Promise<void>`.

---

### 7. Missing: no test covers the interaction where both a legacy localStorage `"agenda"` value AND `initialView` are provided simultaneously

This is the specific scenario from finding #1 above. The five provider tests cover the happy path (no localStorage, localStorage with different view, write-through, agenda companion, undefined passthrough) but not the "legacy localStorage + override" combination. This is the exact race described in finding #1 and the one most likely to regress if the effect ordering changes.

**Suggested test**:
```ts
it("override wins when localStorage holds the legacy agenda value AND initialView is provided", async () => {
  window.localStorage.setItem(
    "calendar-settings",
    JSON.stringify({ view: "agenda", badgeVariant: "colored", use24HourFormat: true, agendaModeGroupBy: "date", weekStartDay: 0 })
  );
  renderWithOverrides({ initialView: "year" });
  await waitFor(() => {
    expect(screen.getByTestId("view")).toHaveTextContent("year");
  });
  await waitFor(() => {
    const parsed = JSON.parse(window.localStorage.getItem("calendar-settings") ?? "{}");
    expect(parsed.view).toBe("year");
  });
});
```

---

### Summary

| # | Severity | Actionable? |
|---|----------|-------------|
| 1 | Moderate — silent ordering coupling between two effects | Yes — comment or merge into one effect |
| 7 | Moderate — no test for legacy-agenda-LS + initialView combo | Yes — add test |
| 2 | Low — fragile `as TCalendarView` cast | Hardening — addressable in follow-up |
| 3 | Low — `?agendaMode=true`-only path untested | Minor — comment or test |
| 4 | Trivial — double cast inherits existing pattern | No action needed |
| 5 | Trivial — duplicate `year` test case | No action needed |

The fix is correct for all currently-tested paths. Findings #1 and #7 are the only ones worth addressing before merge — they're related (same scenario) and a test + comment would close both.

---

BenSeymourODB:
Round-2 push (`ad751b2`) addressing both reviews. Mapping each finding to what changed:

| Review point | Resolution |
|---|---|
| **R1 #1 / R2 #1** — silent ordering between two `setSettings` effects | Merged into a single atomic `useEffect` (migration runs inside the functional updater first, override fields replace it second). The ordering coupling is gone — there are no longer two separate writes whose order matters. `CalendarProvider.tsx` lines ~183–217 in the new diff. |
| **R1 #2** — `?agendaMode=true` alone seeds state on top of a localStorage view | Page now only forwards `initialAgendaMode` when an `initialView` is also set, so the deep link is self-contained. |
| **R1 #3** — missing test for orphan `?agendaMode=true` | Added `it("ignores ?agendaMode=true when no view param is present", ...)` in `page.test.tsx`. |
| **R1 #4 / R2 #2** — `as TCalendarView` cast on a value that may be `"agenda"` | Replaced the single `VIEW_PARAM_VALUES` array with `CANONICAL_VIEW_PARAMS` (`as const satisfies readonly TCalendarView[]`) + `LEGACY_VIEW_ALIASES`. The canonical typeguard narrows to `TCalendarView` directly; the cast is gone. |
| **R1 #5 / R2 #7** — no test for legacy `localStorage("agenda")` + `initialView` | Added `it("override wins when localStorage holds the legacy agenda value AND initialView is provided", ...)` — asserts both in-state view and persisted view land on the override. |
| **R2 #3** — `?agendaMode=true` on its own is now blocked, see R1 #2 above | Same change handles both. |
| **R2 #4, #5, #6** — minor / no action | No change needed. |

`pnpm test` → 1589 pass (+2 over round 1), `pnpm check-types` and `pnpm lint` clean. Note: round-2 also legitimately preserves the user's pre-#150 `agendaMode: true` preference when migrating a legacy `"agenda"` localStorage value — the URL `?view=year` only overrides the view, not the agenda flag, since year doesn't use it. A comment in the merged effect documents this.

---

PR #242, opened 2026-05-03
test(calendar): cover day-overflow popover→modal keyboard + E2E (#209)

### LINKED ISSUE #209: feat(calendar): wire day-overflow popover event cards to open EventDetailModal
## Context

Deferred from PR #138 ("clickable day overflow popover"):

> Clicking an event inside the popover currently does nothing. Wiring each card to the event detail modal is tracked under #81 (out of scope here).
>
> Reviewer (#138): Issue #88 specifies each popover event card should be clickable... The `<li>` items in the popover are not interactive.
>
> — https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/138#issuecomment-4284229972

## Scope

- Make each event card inside the day-overflow popover an interactive button (or `role="button"` `<li>`)
- On click, close the popover and open `EventDetailModal` for the clicked event
- Keyboard support: Tab to event, Enter/Space to activate
- Match the same click-handler used by month-cell event pills so behaviour is consistent

## Acceptance criteria

- [ ] Click → modal opens with correct event
- [ ] Keyboard activation works
- [ ] Closing the modal restores focus to the trigger that opened the popover (or to the event card, if the modal returns focus mid-popover)
- [ ] Playwright E2E covering the popover→modal flow (see related testing issue)

## Dependencies

- Modal is now wired (PR #192) — this issue is unblocked.

BenSeymourODB:
Closes #209.

## Context

The production wiring for "click a card inside the day-overflow popover → open `EventDetailModal`" already lives in `SimpleCalendar.tsx:388-397` + `DayOverflowPopover.tsx:103-122` (delivered alongside PR #138 / #192). Two of #209's four acceptance criteria already had passing component tests:

- ✅ Click → modal opens with correct event (`SimpleCalendar.test.tsx:1118`)
- ✅ Closing modal restores focus to popover trigger (`SimpleCalendar.test.tsx:1147`)

The remaining acceptance criteria — **explicit keyboard activation** and a **Playwright E2E** — were the gap this PR closes.

## Phase 1 — keyboard activation component tests

`SimpleCalendar.test.tsx` gains two cases under the existing `Day overflow popover` describe block: focus a popover event card and press `Enter` / `Space`, then assert the modal heading reflects the chosen event. Native `<button>` semantics already make this work — the tests lock the contract so a future refactor that replaces the buttons can't silently regress keyboard users.

Side fix: `AgendaList.test.tsx` mock context was missing the now-required `deleteEvent` provider field, causing `pnpm check-types` to fail on `main`. Added a no-op `vi.fn().mockResolvedValue(undefined)` so the type check is clean. Pre-existing miss from PR #115's provider extension; tracked here because it blocks CI on this branch.

## Phase 2 — Playwright E2E with video capture

`e2e/day-overflow-popover-modal.spec.ts` against `/test/calendar?events=overflow&view=month&controls=false` (the overflow fixture seeds 10 events on today, so the `+N more` trigger is always present).

Three cases:
1. Click the 5th overflow event card → modal opens with that event's heading; the popover unmounts when the dialog grabs focus.
2. Keyboard path: focus a card and press Enter → modal opens.
3. Close the modal → focus returns to the `+N more` trigger.

Video capture (`test.use({ video: "on" })`) gives the wall-display use case a visual artefact for review.

## Acceptance criteria

- [x] Click → modal opens with correct event
- [x] Keyboard activation works
- [x] Closing the modal restores focus to the trigger that opened the popover
- [x] Playwright E2E covering the popover→modal flow

## Test plan

- [x] `pnpm test` — 1574/1574 ✅
- [x] `pnpm check-types` ✅
- [x] `pnpm lint:fix` (warnings only, no errors) ✅
- [x] `pnpm format:fix` ✅
- [x] CI green on second run (first run hit a transient Prisma-validation flake — unrelated to this PR's diff)
- [ ] Playwright not runnable in this sandbox (browser download network-restricted); spec validates against existing patterns in `e2e/event-detail-modal.spec.ts` + `e2e/calendar-filter-panel.spec.ts`.

https://claude.ai/code/session_01U2UyHqBw6bWkxjXj6e7Toh


BenSeymourODB (commented):
Test-only PR filling the last two acceptance criteria for #209. The component tests are real assertions backed by fixed dates — they would genuinely fail if the `onSelectEvent` wiring in `DayOverflowPopover` or `SimpleCalendar` were removed. The `deleteEvent` side-fix is the right call. One blocker around timezone flake in the E2E spec, one important note on the mock value, and two nits. Details inline.

---

BenSeymourODB:
## Detailed review findings

### `e2e/day-overflow-popover-modal.spec.ts` — line 17 — BLOCKER: timezone mismatch risk

```ts
const dayKey = new Date().toISOString().slice(0, 10);
```

This is evaluated **at module load time in Node.js**, so it uses UTC. The overflow fixture in `src/app/test/calendar/page.tsx` builds event `startDate` values via `getRelativeDate(0, ...)`, which calls:

```ts
const date = new Date();
date.setDate(date.getDate() + 0);
date.setHours(hours, minutes, 0, 0);
return date.toISOString();
```

`new Date()` + `setHours()` operates in the **local timezone of the browser/server process**. On a CI runner in UTC+0 this is fine, but on any runner or developer machine where local time is UTC-N and the test runs during the UTC-N window where local date < UTC date (i.e. after midnight UTC but before midnight local), `dayKey` will be tomorrow's date from the browser's perspective. The popover trigger `data-testid` is keyed off the **local** date via `date-fns/format` (`toDateKey` in `SimpleCalendar.tsx`), so the selector `day-overflow-trigger-${dayKey}` will find no element and all three tests fail with a timeout.

**Fix:** derive `dayKey` inside each test from the browser's own clock, not Node's:

```ts
// Instead of the module-level const, compute dayKey in the browser
const dayKey = await page.evaluate(() => {
  const d = new Date();
  return `${d.getFullYear()}-${String(d.getMonth() + 1).padStart(2, "0")}-${String(d.getDate()).padStart(2, "0")}`;
});
```

Or move it into `test.beforeEach` so each test gets its own browser-side value. This mirrors how the production code constructs the key and eliminates the cross-midnight flake entirely.

---

### `src/components/calendar/__tests__/AgendaList.test.tsx` — line 65 — IMPORTANT: mock return value matches signature but may mask assertion gaps

```ts
deleteEvent: vi.fn().mockResolvedValue(undefined),
```

The `ICalendarContext` signature is `deleteEvent: (eventId: string, calendarId: string) => Promise<void>`, so `mockResolvedValue(undefined)` is the correct resolved type — the fix is sound and unblocks `pnpm check-types`.

One note for future test authors: `AgendaList` itself does not call `deleteEvent` directly, so this field is provided purely to satisfy the context type. If `AgendaList` ever gains a delete affordance, the existing tests will need explicit call assertions. Not a blocker for this PR, just worth noting in the fixture comment.

---

### `e2e/day-overflow-popover-modal.spec.ts` — line 83 — NIT: `trigger.focus()` may not be reliable cross-browser for popover open

```ts
const trigger = page.getByTestId(`day-overflow-trigger-${dayKey}`);
await trigger.focus();
await page.keyboard.press("Enter");
```

Using `.focus()` + keyboard in Playwright is generally fine for Chromium, but on WebKit `.focus()` on a `<button>` may not produce the same focused state that a real Tab keypress would (WebKit historically requires `tabindex` or user interaction to make buttons keyboard-focusable via script). Since the config runs this spec on `webkit` too, consider replacing `.focus()` + `{Enter}` with `page.keyboard.press("Tab")` to reach the trigger naturally, or use `.click()` for the trigger open (matching the existing click test) and only use `.focus()` + `{Enter}` for the card activation where the element is already in a focusable popover context. Low probability of flake in practice since JSDOM/Radix handles this fine, but worth noting for the CI webkit project.

---

### `e2e/day-overflow-popover-modal.spec.ts` — line 93 — NIT: assertion gap on popover dismissal after keyboard activation

The keyboard test asserts the modal heading appears but does not assert `await expect(popover).not.toBeAttached()` after the modal opens (unlike the click test on line 58). Not a correctness issue since the heading assertion proves the modal opened, but parity with the click test would lock the Radix popover-dismissal contract for the keyboard path too — exactly the kind of regression the PR description says it wants to prevent.


---

BenSeymourODB:
## Round-2 follow-up

All four findings addressed in `3d26806`. Per-finding response:

### BLOCKER — timezone mismatch on `dayKey` ✅ fixed

You're right, this would have flaked any time the runner's local date trailed UTC. The fix follows your suggestion — `dayKey` is now derived per-test from the browser's own clock via `page.evaluate()`, mirroring the production `format(day, "yyyy-MM-dd")` in `SimpleCalendar.tsx`. Pulled the helper out as `getTodayDayKey(page)` so the three tests share one implementation. The module-level `const` is gone.

### IMPORTANT — `deleteEvent` mock note ✅ acknowledged, no code change

Agreed on the read: `mockResolvedValue(undefined)` is structurally correct (`Promise<void>`) and `AgendaList` doesn't currently call it, so it's there purely to satisfy the context type. I'm leaving the mock as-is — adding a JSDoc-style "this field is structural" comment in the fixture would prematurely document a contract that may change in either direction (AgendaList could grow a delete affordance, or the provider could split deleteEvent off into a separate hook). Future test authors hitting it will see the mock plus the call-site lookups easily; no action needed in this PR.

### NIT — `trigger.focus()` cross-browser reliability ✅ fixed

Replaced `trigger.focus()` + `{Enter}` with `trigger.click()` for the popover open in the keyboard test. The keyboard path under test was always the in-popover card activation; the trigger-open is incidental. Reserving scripted `.focus()` for the card (which is inside an open Radix popover and naturally focusable) avoids the WebKit edge case you flagged.

### NIT — parity assertion in keyboard test ✅ fixed

The keyboard test now asserts `expect(popover).not.toBeAttached()` after the modal opens, matching the click test. Locks the Radix focus-grab / popover-dismiss contract for both activation paths. Good catch — this is exactly the "regression that breaks Radix focus handoff" scenario the PR exists to prevent.

CI was green on the previous push (the first run was a transient Prisma-validation flake). Ready for round 2 review when you have a moment.

---

PR #245, opened 2026-05-03
test(calendar): unit-test AgendaCalendar helpers + unique mock-event ids (#222)

BenSeymourODB:
Closes a meaningful slice of #222 — items **1**, **2**, and **4** of the test backlog called out from PR review feedback.

### Phase 1 — items 1 + 2 — `AgendaCalendar` pure-helper coverage

`filterEventsBySearch`, `groupEventsByColor`, and `sortEventsByStartTime` were only exercised through `AgendaCalendar`'s rendering tests, leaving edge cases un-asserted. This phase:

- Exports the three helpers from `AgendaCalendar.tsx` (no runtime change — visibility only).
- Adds `src/components/calendar/__tests__/agenda-helpers.test.ts` with 22 isolated unit tests covering:
  - empty / whitespace-only queries
  - title vs description vs attendee-name match (case-insensitive, trimmed)
  - sparse `description` / `user`
  - input-mutation safety
  - sort stability and ascending order
  - **explicit assertion that `groupEventsByColor` keeps each color bucket sorted ascending by `startDate`** — item 2's missing guarantee.

### Phase 2 — item 4 — `createMockEvent` unique default ids

Seven near-duplicate `createMockEvent` helpers across the test suite all defaulted to the same literal id (`"test-event-1"` or `"evt"`), silently producing React-key collisions in any test that rendered multiple events without explicit ids. This phase:

- Adds a single shared fixture at `src/test/fixtures/calendar-event.ts` exporting `createMockEvent` (counter-driven unique default ids: `mock-event-1`, `mock-event-2`, …) and `resetMockEventIds()` for tests that want deterministic ids in `beforeEach`.
- Replaces the seven duplicates in `AgendaCalendar.test.tsx`, `SimpleCalendar.test.tsx`, `YearCalendar.test.tsx`, `WeekCalendar.test.tsx`, `DayCalendar.test.tsx`, `AnalogClockView.test.tsx`, and `calendar-helpers.test.ts` with imports from the shared fixture.
- `calendar-helpers.test.ts` still owns date-based defaults for its assertions, so it imports the shared helper as `createBaseMockEvent` and re-wraps it — picking up the unique-id behaviour for free.
- Adds `__tests__/calendar-event.test.ts` (7 tests) to lock in the contract: uniqueness, override precedence, counter reset.

## Co-fix

The pre-existing `AgendaList.test.tsx` typecheck failure on `main` (missing `deleteEvent` mock) was blocking this PR's CI. Mirrored the one-liner fix from open PR #244 so this branch is green. If #244 lands first, this hunk will trivially three-way merge.

## Deferred from #222

- Item 3 (MiniCalendarSidebar day-click E2E) — better paired with the broader sidebar work.
- Item 5 (popover→modal Playwright E2E) — blocked by #209 (now PR #242).
- Item 6 (scheduler edge-case tests) — owned by #221.
- Item 7 (Profile fixtures retype) — separable infrastructure change.

## Verification

```bash
pnpm test                                # 1601 / 1601 passing (was 1594 + 7 new fixture tests)
pnpm check-types                         # clean
pnpm lint:fix && pnpm format:fix         # clean (5 pre-existing warnings, none from this PR)
```

Closes #222 (partial — items 1, 2, 4)

https://claude.ai/code/session_01Hvf1tfQCMBpeJNavcM5ncP


src/test/fixtures/calendar-event.ts
```diff
@@ -0,0 +1,48 @@
...
+  mockEventCounter = 0;
+}
+
+/**
+ * Build an `IEvent` for tests. Each call without an explicit `id` gets a
+ * unique default (`mock-event-1`, `mock-event-2`, …) so multi-event renders
+ * never collide on React keys.
+ */
+export function createMockEvent(overrides: Partial<IEvent> = {}): IEvent {
+  mockEventCounter += 1;
```
BenSeymourODB, line 31: **Counter always increments, even for explicit-id calls — test name below is misleading**

`mockEventCounter += 1` runs unconditionally before the spread, so `createMockEvent({ id: "custom" })` silently bumps the counter to 1 and the next default-id call returns `"mock-event-2"` instead of `"mock-event-1"`. That's the right trade-off (simplicity over exactness), but the corresponding test in `calendar-event.test.ts` is named `"does not increment the counter for calls that pass an explicit id"` — which is the opposite of what happens. The comment inside the test acknowledges this, but a reader who only sees the test name will have the wrong mental model.

Suggestion: rename the test to `"an explicit id override does not break the uniqueness of subsequent default ids"` (or similar) so the name matches the behaviour that is actually being asserted.

---

src/test/fixtures/__tests__/calendar-event.test.ts
```diff
@@ -0,0 +1,82 @@
...
+    const event = createMockEvent();
+    expect(event.id).toMatch(/^mock-event-\d+$/);
+  });
+
+  it("honours an explicit id override", () => {
+    const event = createMockEvent({ id: "custom" });
+    expect(event.id).toBe("custom");
+  });
+
+  it("does not increment the counter for calls that pass an explicit id", () => {
```
BenSeymourODB: **Test name contradicts the implementation**

This test is named `"does not increment the counter for calls that pass an explicit id"` but the fixture unconditionally does `mockEventCounter += 1` before spreading overrides — the counter *always* increments. The comment inside the test body correctly says "even if the counter ticks for an overridden call" but the name says the opposite.

What the test actually guards is that *the next default-id call is still unique after an override call* — which is a fine thing to assert. The name just needs to match that intent. See the companion comment on `calendar-event.ts` line 31.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    ]);
+  });
+
+  it("does not mutate the input array", () => {
+    const snapshot = events.map((e) => e.id);
+    filterEventsBySearch(events, "soccer");
+    expect(events.map((e) => e.id)).toEqual(snapshot);
+  });
+
+  it("is stable in original order for partial matches", () => {
```
BenSeymourODB: **"Is stable in original order for partial matches" is tautological — query `"a"` matches all three events**

The test's intent is to show that `filterEventsBySearch` preserves insertion order for matched results. But the chosen query `"a"` matches all three events (`"Dentist Appointment"` / `"Annual cleaning"` / `"Emma"`, `"Soccer Practice"` / `"water"` / `"Dad"`, `"Birthday Party"` / `"Grandma"` / `"Grandma"`), so `filter` returns all elements unchanged. `Array.prototype.filter` is specified to preserve order, so the assertion reduces to "the array still equals itself" — it cannot fail regardless of any implementation change.

A meaningful stability test would use a query that matches a *proper subset* of events, interleaved in the array. For example, swap the `events` definition so a middle element does NOT match, and assert the remaining matched elements appear in their original relative positions:

```ts
it("is stable in original order for partial matches", () => {
  // e2 (Soccer Practice / water / Dad) does not contain "cleaning" or "grandma"
  const result = filterEventsBySearch(events, "cleaning");
  expect(result.map((e) => e.id)).toEqual(["e1"]);
  // Or use a query that hits e1 and e3 but skips e2:
  const result2 = filterEventsBySearch(events, "appointment");  // only e1
  // A richer fixture would interleave matching and non-matching events to verify order.
});
```

As written the test is harmless, but it gives false confidence.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    ]);
+  });
+
+  it("returns an empty array for empty input", () => {
+    expect(sortEventsByStartTime([])).toEqual([]);
+  });
+});
+
+describe("groupEventsByColor", () => {
+  it("buckets events under their color key", () => {
```
BenSeymourODB: **"Buckets events under their color key" does not distinguish grouping from sorting**

`b1` and `b2` both use the default `startDate` (`"2026-05-03T10:00:00.000Z"`), so after `sortEventsByStartTime` runs inside the bucket they remain in insertion order regardless. The assertion `["b1", "b2"]` would pass even if `groupEventsByColor` never called `sortEventsByStartTime` at all. This is fine as a grouping test, but it means it cannot catch a regression where the per-bucket sort is accidentally removed.

The dedicated regression test at line 214 covers the sort properly, so this is low-severity. But worth noting: the two tests overlap without the first one adding confidence about sort behaviour. Consider giving `b1` and `b2` distinct `startDate` values here too (b2 earlier than b1) so that this test independently verifies sort-within-bucket in addition to grouping.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    ];
+    const before = events.map((e) => e.id);
+
+    const result = sortEventsByStartTime(events);
+
+    expect(result).not.toBe(events);
+    expect(events.map((e) => e.id)).toEqual(before);
+  });
+
+  it("is stable for events that share a start time", () => {
```
BenSeymourODB: **Sort-stability test is spec-guaranteed and very small — low value as written**

All three events share the same `startDate`, so the comparator always returns `0`. `Array.prototype.sort` has been required to be stable since ES2019, so the order `["first","second","third"]` is mandated by the language spec, not by anything in `sortEventsByStartTime`'s implementation. The test cannot fail on any conformant JS engine.

This is not a blocking issue — the test is at least readable documentation. But if you want this to genuinely guard stability, you'd need an implementation that has a non-stable sort path (e.g., uses an unstable algorithm), which this implementation does not. Consider keeping it as a documentation test or converting it to show the semantics of mixed same-time/different-time ordering.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
```diff
@@ -0,0 +1,284 @@
...
+  it("trims whitespace around the query before matching", () => {
+    const result = filterEventsBySearch(events, "   DENTIST   ");
+    expect(result.map((e) => e.id)).toEqual(["e1"]);
+  });
+
+  it("returns an empty list when nothing matches", () => {
+    expect(filterEventsBySearch(events, "nothing-matches-this")).toEqual([]);
+  });
+
+  it("does not throw when an event has an undefined description", () => {
```
BenSeymourODB: **`undefined description` test only verifies the title-match path — description-match path with undefined is untested**

The test confirms the function does not throw when `description` is `undefined` and matches via title (`"walk"`). But it does not test the case where the user searches for a term that would only appear in the description field — which would exercise the `event.description ?? ""` branch in the filter haystack. If someone changed `??` to `?.` or accidentally dropped it, this test would still pass.

Suggestion: add a companion assertion (or a second event) where the match would come from the description value if it were present, and verify the event is NOT returned when description is undefined and the query matches only what the description would have contained. Or simply rename the test to `"does not throw when description is undefined (matches by title)"` to accurately scope what is being tested.

---

BenSeymourODB (commented):
## Review summary

The PR is structurally sound and the three issue-#222 items (helper exports, sort-order assertion, unique fixture ids) are all addressed correctly. No tests were broken, no banned patterns were introduced (no `any`, no manual memoisation, no committed artefacts), and the `deleteEvent` co-fix is an exact functional mirror of PR #244. The line-level comments above call out specific things to fix or watch; summarised here:

**Substantive (worth fixing before merge):**

1. **Misleading test name in `calendar-event.test.ts` line 32** — `"does not increment the counter for calls that pass an explicit id"`. The counter *always* increments; the name contradicts the implementation. The comment inside the test body correctly explains what is actually asserted. Rename the test to match what it checks.

2. **`filterEventsBySearch` stability test (agenda-helpers.test.ts line 119) is tautological** — the query `"a"` matches all three events, so `filter` returns the full array unchanged. This cannot fail regardless of implementation. It looks like a stability test but isn't; it should use a query that matches a strict subset interleaved in the array.

3. **`undefined description` test (agenda-helpers.test.ts line 87) only exercises the title-match branch** — the test confirms no throw, but the description-path (`event.description ?? ""`) is not actually stressed when description is `undefined`. Rename or extend.

**Low-severity (informational, no action required to merge):**

4. **`groupEventsByColor` "buckets" test (line 178)** — `b1` and `b2` share the same default `startDate`, so the assertion would pass even if per-bucket sorting were removed. The dedicated regression test at line 214 covers this correctly; this test is purely a grouping check. Consider giving `b2` an earlier `startDate` to make the test independently self-checking.

5. **Sort-stability test for `sortEventsByStartTime` (line 158)** — all three events have the same time, so the stable-sort assertion is guaranteed by the ES2019 spec, not tested by exercising any code path. Harmless documentation, but low guard value.

6. **`MiniCalendarSidebar.test.tsx` was not migrated** — it already had its own local sequential counter before this PR and was left behind. The PR description says "seven duplicates" but only six were migrated. Not blocking — the local implementation is functionally equivalent — but the description is slightly inaccurate.

Everything else (CLAUDE.md compliance, fixture location, self-tests, `AgendaList.test.tsx` co-fix, TypeScript strictness, no Tailwind violations, no React Compiler violations) is clean.

---

src/test/fixtures/calendar-event.ts
BenSeymourODB: Addressed in `e0df86a`. Renamed the test to `"an explicit id override does not break the uniqueness of subsequent default ids"` and updated the in-test comment to be explicit that the counter increments unconditionally — that's the documented trade-off, not an oversight.

Left the implementation as-is (counter ticks for every call, including overrides) because the alternative — only ticking for default-id calls — would let the caller smuggle in a colliding default by passing `id: "mock-event-2"` explicitly. The current behaviour keeps the default sequence monotonic and decoupled from caller-supplied values.

---

src/test/fixtures/__tests__/calendar-event.test.ts
BenSeymourODB: Addressed in `e0df86a` — see the companion reply on `calendar-event.ts` line 31. The test name now matches the actual contract being asserted.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
BenSeymourODB: Fair catch. Replaced in `e0df86a` with a fixture that interleaves matching and non-matching events:

```ts
it("preserves the original input order when matches are interleaved with non-matches", () => {
  const interleaved: IEvent[] = [
    makeEvent({ id: "match-first", title: "Birthday Bash", description: "" }),
    makeEvent({ id: "skip-middle", title: "Soccer Practice", description: "" }),
    makeEvent({ id: "match-last",  title: "Birthday Cake",  description: "" }),
  ];
  const result = filterEventsBySearch(interleaved, "birthday");
  expect(result.map((e) => e.id)).toEqual(["match-first", "match-last"]);
});
```

Now the assertion can only pass if matched events keep their relative input position — a regression that scrambled the output order would actually surface.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
BenSeymourODB: Addressed in `e0df86a`. `b1` is now `15:00`, `b2` is `09:00`, and the assertion expects `["b2", "b1"]` — so this test independently verifies sort-within-bucket and would catch a regression that dropped the `sortEventsByStartTime` call inside `groupEventsByColor`. The dedicated three-event-per-bucket regression at the bottom of the file stays as a richer back-stop.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
BenSeymourODB: Converted in `e0df86a` to mix same-time and different-time events:

```ts
it("interleaves same-time and different-time events deterministically", () => {
  const events: IEvent[] = [
    makeEvent({ id: "tie-a",    startDate: "...09:00:00.000Z" }),
    makeEvent({ id: "later",    startDate: "...15:00:00.000Z" }),
    makeEvent({ id: "tie-b",    startDate: "...09:00:00.000Z" }),
    makeEvent({ id: "earliest", startDate: "...08:00:00.000Z" }),
    makeEvent({ id: "tie-c",    startDate: "...09:00:00.000Z" }),
  ];
  expect(sortEventsByStartTime(events).map((e) => e.id)).toEqual([
    "earliest", "tie-a", "tie-b", "tie-c", "later",
  ]);
});
```

Now the assertion exercises the actual contract — different-time events fall into ascending order while same-time events keep their relative input position. ES2019 stable-sort is still the underlying mechanism, but the assertion is no longer a tautology against a single comparator value.

---

src/components/calendar/__tests__/agenda-helpers.test.ts
BenSeymourODB: Addressed in `e0df86a`. Reframed the test around the actual contract — "treats an undefined description as the empty string (does not match against the literal 'undefined')" — and added a companion assertion that asserts the description-only-match query does NOT match when description is undefined. Together they cover both directions of the `event.description ?? ""` branch:

```ts
expect(filterEventsBySearch(sparse, "undefined")).toEqual([]); // ?? branch is short-circuited, not stringified
expect(filterEventsBySearch(sparse, "annual")).toEqual([]);    // description-only match short-circuits
```

A regression that dropped `??` would surface as either the `"undefined"` substring leaking into the haystack or a runtime throw on `.toLowerCase()`.

---

PR #246, opened 2026-05-03
chore(infra): enforce Prisma migration naming convention (#219)

### LINKED ISSUE #219: chore(infra): standardize Prisma migration naming convention
## Context

Deferred from PR #157:

> Worth a standalone decision issue if the convention should change project-wide; out of scope for this PR.
>
> — @BenSeymourODB, https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/157#issuecomment-4293142503

## Problem

Recent migrations have inconsistent naming — some use Prisma's default timestamp prefix, others use sequential numeric prefixes. We need a single convention so future migrations don't drift further.

## Decision required

Choose one:

1. **Default Prisma** — `YYYYMMDDHHMMSS_descriptive_name/` (timestamp-prefixed; the framework default)
2. **Sequential** — `0001_descriptive_name/` (manually incremented; matches some shops' preference for tidy ordering)

## Acceptance criteria

- [ ] Decision documented in `docs/database.md` with rationale
- [ ] Existing migrations renamed to match (or a one-off ADR explaining the cutover point)
- [ ] CI check or `pnpm` script that warns on non-conforming new migration names
- [ ] Update onboarding docs (`CLAUDE.md` Quick Reference) to mention the convention

## Notes

Small but high-leverage — currently the only piece of friction every new migration touches.

BenSeymourODB:
Closes #219.

## Decision

Going with **option 2 — sequential `NNNN_descriptive_name`**, formalising what the codebase already does. All four existing migrations (`0001_initial`, `0002_add_scheduler_settings`, `0003_unique_account_user_provider`, `0004_add_calendar_settings`) follow this pattern, so no rename is needed.

Rationale:

- Tidier ordering for code review (4-digit prefixes line up; timestamps don't).
- Existing migrations already form a coherent sequence; switching to timestamps would require renumbering and bumping every dev's local `_prisma_migrations` row for zero functional benefit.
- The "next number" is unambiguous in PRs (timestamps can collide if two PRs are open at the same minute).

## Phase 1 — Enforcement (this commit)

- **`scripts/check-migration-naming.mjs`** scans `prisma/migrations/` and asserts that every subdirectory matches `^[0-9]{4}_[a-z0-9_]+$` *and* that prefixes are strictly sequential starting at 0001 (no gaps, no duplicates).
- **Vitest unit tests** (`scripts/__tests__/check-migration-naming.test.ts`) cover: canonical input, regex failures (uppercase, hyphens, short/long prefixes, timestamp-prefixed Prisma defaults), gaps, duplicates, not-starting-at-0001, mixed files/dirs, shuffled input order, empty directory.
- **`pnpm db:migrate:check-names`** new script.
- **CI integration** — wired into `.github/workflows/main_nextjs-template-build.yml` immediately after the existing migration-drift step so a non-conforming name fails CI on the same job.

### Drive-by

`src/components/calendar/__tests__/AgendaList.test.tsx` was missing the `deleteEvent` mock added to `ICalendarContext` in PR #197. The same one-line patch is carried by PRs #228, #229, and #244 — included here so this PR's typecheck step passes on its own.

## Phase 2 — Documentation (follow-up commit on this same PR)

- ADR-style entry in `docs/database.md` documenting the convention.
- Update `CLAUDE.md` Quick Reference to mention the rule and the new check script.

## Out of scope (deferred)

- Pre-commit hook to suggest the next `NNNN` prefix. CI catch is sufficient and the friction of the hook outweighs its value for an occasional task.
- Renaming any existing migration (none violate the rule).

## Test plan

- [x] `pnpm db:migrate:check-names` → ✓ 4 migrations follow the convention.
- [x] `pnpm test:run scripts/__tests__/check-migration-naming.test.ts` → 12/12 passing.
- [x] `pnpm test:run` → 1584/1584 passing.

---

---
_Generated by [Claude Code](https://claude.ai/code/session_01S3GHWBt1Gby2J1r5fo8kxB)_

### DIFF
--- a/.github/workflows/main_nextjs-template-build.yml
+++ b/.github/workflows/main_nextjs-template-build.yml
@@ -41,5 +41,8 @@ jobs:
             exit 1
           }
 
+      - name: Validate Prisma migration naming convention
+        run: pnpm db:migrate:check-names
+
       - name: Build
         run: pnpm build
--- a/docs/database.md
+++ b/docs/database.md
@@ -5,11 +5,50 @@ This project uses **PostgreSQL** with **Prisma ORM** for database management.
 ## Quick Reference
 
 ```bash
-pnpm db:migrate          # Create a new migration (development)
-pnpm db:migrate:deploy   # Apply pending migrations (production)
-pnpm db:migrate:reset    # Reset database and reapply all migrations
-pnpm db:migrate:status   # Check migration status
+pnpm db:migrate              # Create a new migration (development)
+pnpm db:migrate:deploy       # Apply pending migrations (production)
+pnpm db:migrate:reset        # Reset database and reapply all migrations
+pnpm db:migrate:status       # Check migration status
+pnpm db:migrate:check-names  # Validate the NNNN_snake_case naming convention
 ```
+
+## Migration Naming Convention
+
+This project uses **sequential numeric prefixes** for migration directory names:
+
+```
+NNNN_snake_case_description
+```
+
+- `NNNN` — strictly increasing 4-digit prefix starting at `0001` (no gaps, no duplicates).
+- `snake_case_description` — lowercase letters, digits, and underscores only.
+
+Examples:
+
+```
+0001_initial
+0002_add_scheduler_settings
+0003_unique_account_user_provider
+0004_add_calendar_settings
+0005_add_meal_planning
+```
+
+### Why sequential, not Prisma's default timestamp prefix?
+
+|                                         | Sequential `NNNN_`                             | Timestamp `YYYYMMDDHHMMSS_`              |
+| --------------------------------------- | ---------------------------------------------- | ---------------------------------------- |
+| Tidy ordering in code review            | ✓ aligned columns                              | ✗ wide column                            |
+| Collision-free with concurrent PRs      | ✓ obvious next number; conflicts surface in CI | ✗ two PRs at the same minute can collide |
+| Communicates ordering at a glance       | ✓ `0007` clearly follows `0006`                | ✗ requires reading the timestamp         |
+| Matches Prisma's out-of-the-box default | ✗ requires explicit naming                     | ✓                                        |
+
+Decision: trade the small friction of explicitly naming each new migration `NNNN_…` for the larger gains in review legibility and conflict avoidance.
+
+### Enforcement
+
+`pnpm db:migrate:check-names` (also wired into CI) validates that every entry under `prisma/migrations/` matches the convention and that prefixes are strictly sequential. Run it locally before pushing if you've added a migration.
+
+When `pnpm db:migrate` prompts for a name, answer with the snake-case description only — `add_scheduler_settings`, not `0005_add_scheduler_settings`. Prisma normally prepends a timestamp, so after the migration is created, **rename the directory** from `<timestamp>_add_scheduler_settings/` to the next sequential prefix (`0005_add_scheduler_settings/`) before committing. The check script will reject the timestamp form so a forgotten rename fails CI.
 
 ## Development Workflow
 
@@ -24,7 +63,9 @@ pnpm db:migrate:status   # Check migration status
    - Generates a SQL migration file in `prisma/migrations/`
    - Applies it to your local database
    - Regenerates the Prisma client
-3. **Commit** the migration file along with the schema change
+3. **Rename the migration directory** to the next sequential prefix (e.g. `0005_add_scheduler_settings/`) — see [Migration Naming Convention](#migration-naming-convention).
+4. **Verify** the name with `pnpm db:migrate:check-names`.
+5. **Commit** the migration file along with the schema change.
 
 ### Example: Adding a New Field
 
@@ -40,9 +81,13 @@ model UserSettings {
 pnpm db:migrate
 # Enter migration name: add_scheduler_interval
 # Migration created and applied!
-```
 
-This creates `prisma/migrations/<timestamp>_add_scheduler_interval/migration.sql`.
+# Rename to the next sequential prefix (see Migration Naming Convention):
+mv prisma/migrations/<timestamp>_add_scheduler_interval \
+   prisma/migrations/0005_add_scheduler_interval
+
+pnpm db:migrate:check-names  # Verify
+```
 
 ## Production Deployment
 
@@ -67,11 +112,13 @@ This tells Prisma that the `0001_initial` migration has already been applied (si
 
 ## CI Validation
 
-The CI pipeline validates that migrations are in sync with the schema. If you modify `prisma/schema.prisma` without creating a migration, CI will fail with:
+The CI pipeline validates two things on every push:
 
-```
-Prisma migrations are out of date. Run 'pnpm db:migrate' to create a new migration.
-```
+1. **Schema-vs-migrations drift** — if you modify `prisma/schema.prisma` without creating a migration, CI fails with:
+   ```
+   Prisma migrations are out of date. Run 'pnpm db:migrate' to create a new migration.
+   ```
+2. **Naming convention** — `pnpm db:migrate:check-names` runs in CI and fails if any migration directory does not match `^[0-9]{4}_[a-z0-9_]+$` or if prefixes are not strictly sequential. See [Migration Naming Convention](#migration-naming-convention).
 
 ## Migration Directory Structure
 
@@ -82,8 +129,8 @@ prisma/
 │   ├── migration_lock.toml          # Provider lock file (do not edit)
 │   ├── 0001_initial/
 │   │   └── migration.sql            # Baseline migration
-│   └── <timestamp>_<name>/
-│       └── migration.sql            # Subsequent migrations
+│   └── NNNN_snake_case_description/
+│       └── migration.sql            # Subsequent migrations (sequential prefix)
 ```
 
 ## Important Rules
--- a/docs/database.md
+++ b/docs/database.md
@@ -48,7 +48,35 @@ Decision: trade the small friction of explicitly naming each new migration `NNNN
 
 `pnpm db:migrate:check-names` (also wired into CI) validates that every entry under `prisma/migrations/` matches the convention and that prefixes are strictly sequential. Run it locally before pushing if you've added a migration.
 
-When `pnpm db:migrate` prompts for a name, answer with the snake-case description only — `add_scheduler_settings`, not `0005_add_scheduler_settings`. Prisma normally prepends a timestamp, so after the migration is created, **rename the directory** from `<timestamp>_add_scheduler_settings/` to the next sequential prefix (`0005_add_scheduler_settings/`) before committing. The check script will reject the timestamp form so a forgotten rename fails CI.
+### Renaming a freshly-created migration
+
+When `pnpm db:migrate` prompts for a name, answer with the snake-case description only — `add_scheduler_settings`, not `0005_add_scheduler_settings`. Prisma still prepends a timestamp to the directory it creates on disk, and **also writes a row into the local `_prisma_migrations` table under the same timestamp name**. Renaming the directory only — without touching the DB row — leaves the two out of sync, and the next `pnpm db:migrate` run will fail with a "migration not found" / drift error.
+
+The full rename ritual is:
+
+```bash
+# 1. Find the timestamp directory Prisma just created.
+TIMESTAMP_DIR=$(ls -1d prisma/migrations/[0-9]*_add_scheduler_settings | head -1)
+NEXT_PREFIX=$(printf "%04d" $(($(ls -1d prisma/migrations/[0-9]*/ | wc -l))))
+NEW_NAME="${NEXT_PREFIX}_add_scheduler_settings"
+
+# 2. Rename the directory.
+mv "$TIMESTAMP_DIR" "prisma/migrations/$NEW_NAME"
+
+# 3. Update the local _prisma_migrations row so Prisma still recognises it.
+#    Either (a) rewrite the row in place:
+psql "$DATABASE_URL" -c "UPDATE _prisma_migrations SET migration_name = '$NEW_NAME' \
+  WHERE migration_name = '$(basename "$TIMESTAMP_DIR")';"
+
+#    or (b) reset the local DB and re-apply from scratch — simplest in dev:
+pnpm db:migrate:reset
+
+# 4. Verify.
+pnpm db:migrate:check-names
+pnpm db:migrate:status
+```
+
+The check script rejects the timestamp form, so a forgotten rename fails CI. Production databases are not affected — `prisma migrate deploy` reads the renamed directory directly, and the `_prisma_migrations` row is created on first apply with the correct (renamed) value.
 
 ## Development Workflow
 
@@ -82,9 +110,12 @@ pnpm db:migrate
 # Enter migration name: add_scheduler_interval
 # Migration created and applied!
 
-# Rename to the next sequential prefix (see Migration Naming Convention):
+# Rename to the next sequential prefix (replace <NNNN> with the actual next
+# number — check prisma/migrations/ or run pnpm db:migrate:check-names).
+# Remember to also update the local _prisma_migrations row, or reset the
+# dev DB. See "Renaming a freshly-created migration" above.
 mv prisma/migrations/<timestamp>_add_scheduler_interval \
-   prisma/migrations/0005_add_scheduler_interval
+   prisma/migrations/<NNNN>_add_scheduler_interval
 
 pnpm db:migrate:check-names  # Verify
 ```
--- a/scripts/__tests__/check-migration-naming.test.ts
+++ b/scripts/__tests__/check-migration-naming.test.ts
@@ -27,6 +27,13 @@ describe("MIGRATION_NAME_PATTERN", () => {
     expect(MIGRATION_NAME_PATTERN.test("0001")).toBe(false);
     expect(MIGRATION_NAME_PATTERN.test("0001_")).toBe(false);
   });
+
+  it("rejects trailing or doubled underscores in the description", () => {
+    expect(MIGRATION_NAME_PATTERN.test("0001_add_")).toBe(false);
+    expect(MIGRATION_NAME_PATTERN.test("0001___")).toBe(false);
+    expect(MIGRATION_NAME_PATTERN.test("0001_add__settings")).toBe(false);
+    expect(MIGRATION_NAME_PATTERN.test("0001__initial")).toBe(false);
+  });
 });
 
 describe("validateMigrationNames", () => {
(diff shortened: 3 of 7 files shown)


scripts/check-migration-naming.mjs
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env node
+// Validates Prisma migration directory names match the project convention:
+// NNNN_snake_case_description, with NNNN a strictly increasing 4-digit prefix
+// starting at 0001. See docs/database.md for the rationale.
+import { readdirSync, statSync } from "node:fs";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+export const MIGRATION_NAME_PATTERN = /^[0-9]{4}_[a-z0-9_]+$/;
```
BenSeymourODB: **Minor: regex allows a trailing underscore in the description**

`/^[0-9]{4}_[a-z0-9_]+$/` — the character class `[a-z0-9_]` includes `_`, so `0001_add_` passes. In practice no developer will type that, but the test suite has no case for it and the error message says "snake_case_description", which implies no trailing separator.

Consider tightening to forbid a trailing underscore:
```js
export const MIGRATION_NAME_PATTERN = /^[0-9]{4}_[a-z0-9]+(?:_[a-z0-9]+)*$/;
```
This matches `0001_initial` and `0004_add_calendar_settings` but rejects `0001_add_` and `0001___`. If you keep the current regex, add a test case `expect(MIGRATION_NAME_PATTERN.test("0001_add_")).toBe(false)` so the intent is documented either way.

---

scripts/check-migration-naming.mjs
```diff
@@ -0,0 +1,96 @@
...
+ * @param {MigrationEntry[]} entries
+ * @returns {string[]} list of error messages (empty when everything is valid)
+ */
+export function validateMigrationNames(entries) {
+  const errors = [];
+
+  const dirNames = entries
+    .filter((e) => e.isDirectory)
+    .map((e) => e.name)
+    .sort();
```
BenSeymourODB, line 32: **Potential concern: sort order assumption**

`.sort()` on the `dirNames` array is a lexicographic (string) sort. Because NNNN is always 0-padded to exactly 4 digits, lexicographic sort happens to equal numeric sort for 0001–9999, so the sequential-prefix check works correctly in practice.

However, this is a subtle assumption. A brief comment would protect the next reader:
```js
// Lexicographic sort is equivalent to numeric order because prefixes are
// always zero-padded to exactly 4 digits (0001–9999).
const dirNames = entries
  .filter((e) => e.isDirectory)
  .map((e) => e.name)
  .sort();
```
Low priority, but it prevents a future well-intentioned change to `sort((a, b) => ...)` from introducing a subtle bug.

---

scripts/check-migration-naming.mjs
```diff
@@ -0,0 +1,96 @@
...
+  return errors;
+}
+
+/**
+ * Read entries (files + directories) from a real filesystem path.
+ *
+ * @param {string} dir
+ * @returns {MigrationEntry[]}
+ */
+export function readMigrationEntries(dir) {
```
BenSeymourODB, line 65: **`readMigrationEntries` uses `statSync` — symlinks are silently followed**

`statSync` follows symlinks and returns the target's stat. If a migration directory is replaced by a symlink to a directory (unlikely but possible), it will be treated as a real directory entry. `lstatSync` is more defensive here. Not a blocking issue, just worth knowing.

---

scripts/check-migration-naming.mjs
```diff
@@ -0,0 +1,96 @@
...
+    process.exit(1);
+  }
+
+  const count = entries.filter((e) => e.isDirectory).length;
+  console.log(
+    `✓ ${count} migration${count === 1 ? "" : "s"} follow the NNNN_snake_case convention`
+  );
+}
+
+const invokedDirectly =
```
BenSeymourODB: **`invokedDirectly` check: the `resolve()` branch may give a false positive under `pnpm exec`**

```js
const invokedDirectly =
  import.meta.url === `file://${process.argv[1]}` ||
  import.meta.url === `file://${resolve(process.argv[1] ?? "")}`;
```

The second branch resolves `process.argv[1]` relative to `process.cwd()`. When `pnpm db:migrate:check-names` is run from the project root, `process.argv[1]` is already an absolute path (`/home/user/.../scripts/check-migration-naming.mjs`), so `resolve()` is a no-op and both branches are equivalent. The guard works correctly in that case.

The edge case is when someone `import`s the module in another script from a different cwd — `resolve(process.argv[1])` could accidentally match `import.meta.url` of the importer's script if paths collide. This is extremely unlikely in practice, but the standard idiom is simply:
```js
if (process.argv[1] === fileURLToPath(import.meta.url)) {
  main();
}
```
That is unambiguous and doesn't depend on cwd.

---

scripts/__tests__/check-migration-naming.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    const entries = [
+      dir("0001_initial"),
+      dir("0002_first"),
+      dir("0002_second"),
+    ];
+
+    const errors = validateMigrationNames(entries);
+    // Both 0002s sort after 0001; the second one in sort order is at index 2,
+    // expects prefix 0003. Either way, at least one ordering error is reported.
+    expect(errors.length).toBeGreaterThan(0);
```
BenSeymourODB: **Duplicate-prefix test: assertion is weaker than necessary**

```ts
expect(errors.length).toBeGreaterThan(0);
expect(errors.some((e) => e.includes("0002"))).toBe(true);
```

The comment explains why — the second `0002_*` entry is treated as "prefix 0002 at position where 0003 is expected". That's correct behaviour. But `toBeGreaterThan(0)` means the test would still pass even if five spurious errors were emitted. Since the implementation produces exactly one error for this input (only `0002_second` is flagged), tightening to `toHaveLength(1)` would catch a future regression that double-reports or silently suppresses errors:

```ts
expect(errors).toHaveLength(1);
expect(errors[0]).toContain("0002");
```

---

scripts/__tests__/check-migration-naming.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    expect(errors.some((e) => e.includes("0002"))).toBe(true);
+  });
+
+  it("flags a sequence that does not start at 0001", () => {
+    const entries = [dir("0002_skipped_one"), dir("0003_second")];
+
+    const errors = validateMigrationNames(entries);
+    expect(errors).toHaveLength(2);
+    expect(errors[0]).toContain("expected 0001");
```
BenSeymourODB, line 118: **"does not start at 0001" test asserts `toHaveLength(2)` — worth a comment**

```ts
expect(errors).toHaveLength(2);
expect(errors[0]).toContain("expected 0001");
```

This passes because `0002_skipped_one` is at index 0 (expected 0001) and `0003_second` is at index 1 (expected 0002), so both produce errors. That's correct, but a reader skimming the test may expect only one error (the gap before `0002`). A one-line comment would help:

```ts
// Both entries are wrong: 0002 should be 0001, 0003 should be 0002.
expect(errors).toHaveLength(2);
```

---

docs/database.md
```diff
@@ -5,12 +5,51 @@ This project uses **PostgreSQL** with **Prisma ORM** for database management.
...
+| Communicates ordering at a glance       | ✓ `0007` clearly follows `0006`                | ✗ requires reading the timestamp         |
+| Matches Prisma's out-of-the-box default | ✗ requires explicit naming                     | ✓                                        |
+
+Decision: trade the small friction of explicitly naming each new migration `NNNN_…` for the larger gains in review legibility and conflict avoidance.
+
+### Enforcement
+
+`pnpm db:migrate:check-names` (also wired into CI) validates that every entry under `prisma/migrations/` matches the convention and that prefixes are strictly sequential. Run it locally before pushing if you've added a migration.
+
+When `pnpm db:migrate` prompts for a name, answer with the snake-case description only — `add_scheduler_settings`, not `0005_add_scheduler_settings`. Prisma normally prepends a timestamp, so after the migration is created, **rename the directory** from `<timestamp>_add_scheduler_settings/` to the next sequential prefix (`0005_add_scheduler_settings/`) before committing. The check script will reject the timestamp form so a forgotten rename fails CI.
```
BenSeymourODB: **Bug in documented workflow: renaming the directory breaks the local `_prisma_migrations` record**

The docs say:

> rename the directory from `<timestamp>_add_scheduler_settings/` to `0005_add_scheduler_settings/` before committing

This is the right thing to do before committing, but there is a hidden trap: `prisma migrate dev` already **applied** the migration to your local database and recorded it in `_prisma_migrations` under the original timestamp name. After the rename, the filesystem name (`0005_add_scheduler_settings`) no longer matches the database record (`<timestamp>_add_scheduler_settings`). On the next `pnpm db:migrate` run Prisma will detect the mismatch and report:

```
✗ 0005_add_scheduler_settings  Failed to apply cleanly to the shadow database
  The migration file at prisma/migrations/0005_add_scheduler_settings/migration.sql was not found
```
(exact wording varies by Prisma version, but the effect is a "migration not found" or "drift detected" error.)

**The fix is to update the `_prisma_migrations` row immediately after the rename:**

```bash
# After: mv prisma/migrations/<timestamp>_add_x prisma/migrations/0005_add_x
psql $DATABASE_URL -c "
  UPDATE _prisma_migrations
  SET migration_name = '0005_add_x'
  WHERE migration_name = '<timestamp>_add_x';
"
```

Or, equivalently, use `prisma migrate resolve`:

```bash
# Mark the timestamp version as rolled back, then mark the new name as applied.
npx prisma migrate resolve --rolled-back <timestamp>_add_x
npx prisma migrate resolve --applied 0005_add_x
```

This is a real developer-experience footgun. The docs should either describe the resolution step or warn explicitly that `pnpm db:migrate:reset` (drop and reapply from scratch using the now-renamed directory) is the simplest local fix.

---

docs/database.md
```diff
@@ -40,9 +81,13 @@ model UserSettings {
 pnpm db:migrate
 # Enter migration name: add_scheduler_interval
 # Migration created and applied!
-```
 
-This creates `prisma/migrations/<timestamp>_add_scheduler_interval/migration.sql`.
+# Rename to the next sequential prefix (see Migration Naming Convention):
+mv prisma/migrations/<timestamp>_add_scheduler_interval \
```
BenSeymourODB, line 117: **Example uses hardcoded `0005_` — will silently go stale**

```bash
mv prisma/migrations/<timestamp>_add_scheduler_interval \
   prisma/migrations/0005_add_scheduler_interval
```

The project currently has four migrations (`0001`–`0004`), so `0005` is correct today. But when a fifth migration lands, this example becomes misleading and a copy-paste hazard. Consider wording it generically:

```bash
# Determine the next sequential prefix (currently 0005, but check your migrations dir):
mv prisma/migrations/<timestamp>_add_scheduler_interval \
   prisma/migrations/<next_prefix>_add_scheduler_interval
```

Or add a note: _"Replace `0005` with the actual next number — check `prisma/migrations/` or run `pnpm db:migrate:check-names` to confirm."_

---

.github/workflows/main_nextjs-template-build.yml
```diff
@@ -41,5 +41,8 @@ jobs:
             exit 1
           }
 
+      - name: Validate Prisma migration naming convention
```
BenSeymourODB, line 44: **Pre-existing gap: CI has no `pnpm test` step**

The naming-check step is correctly placed (after drift check, before build), so the new CI wiring is fine for this PR's scope.

However, while reading the full workflow it is apparent that the CI job runs lint, typecheck, migration drift, naming check, and build — but **never runs the unit test suite**. The 12 new Vitest tests (and all 1584 existing ones) are never executed in CI. This means a test regression would only be caught locally or by the `test:run` call referenced in the PR description, not automatically on every push.

This is a pre-existing gap, not introduced here. Consider adding:
```yaml
      - name: Test
        run: pnpm test:run
```
before the Build step. Out of scope for this PR if the team wants to track it separately, but worth flagging while the workflow is being touched.

---

scripts/__tests__/check-migration-naming.test.ts
```diff
@@ -0,0 +1,128 @@
+import { describe, expect, it } from "vitest";
```
BenSeymourODB, line 1: **`vitest.config.ts` `include` pattern will pick this file up correctly — confirmed**

`vitest.config.ts` sets `include: ["**/*.{test,spec}.{ts,tsx}"]`, which matches `scripts/__tests__/check-migration-naming.test.ts`. The `exclude` list does not mention `scripts/__tests__/`. So `pnpm test` and `pnpm test:run` will exercise these 12 tests as expected. No action needed — just confirming the wiring is correct.

---

scripts/__tests__/check-migration-naming.test.ts
```diff
@@ -0,0 +1,128 @@
...
+
+  it("flags a sequence that does not start at 0001", () => {
+    const entries = [dir("0002_skipped_one"), dir("0003_second")];
+
+    const errors = validateMigrationNames(entries);
+    expect(errors).toHaveLength(2);
+    expect(errors[0]).toContain("expected 0001");
+  });
+
+  it("returns multiple errors when several names break the rules", () => {
```
BenSeymourODB, line 122: **"returns multiple errors" test: the expected count of 2 silently depends on implementation order**

```ts
const entries = [
  dir("not-a-migration"),
  dir("0001_initial"),
  dir("0003_skipped_two"),
];
const errors = validateMigrationNames(entries);
expect(errors).toHaveLength(2);
```

After sorting, `dirNames` = `["0001_initial", "0003_skipped_two", "not-a-migration"]`. The validator:
1. Emits 1 regex error for `not-a-migration`.
2. Emits 1 sequential error for `0003_skipped_two` (expected `0002`).
= 2 errors total. Correct.

But the assertion doesn't state *which* two errors were expected, so a future refactor that collapses both into one message or adds an extra one would only be caught by a count mismatch. Adding explicit content checks makes the intent clearer:

```ts
expect(errors).toHaveLength(2);
expect(errors.some((e) => e.includes('"not-a-migration"'))).toBe(true);
expect(errors.some((e) => e.includes('"0003_skipped_two"'))).toBe(true);
```

---

src/components/calendar/__tests__/AgendaList.test.tsx
```diff
@@ -62,6 +62,7 @@ function makeContext(
     addEvent: vi.fn(),
     updateEvent: vi.fn(),
     removeEvent: vi.fn(),
+    deleteEvent: vi.fn(),
```
BenSeymourODB: **Drive-by fix is correct and justified**

The one-liner `deleteEvent: vi.fn()` exactly matches what PRs #228, #229, and #244 carry. Including it here is the right call — without it the typecheck step on this branch would fail because `ICalendarContext` gained `deleteEvent` in #197 and any test that constructs a full mock object must include it. The inclusion is well-explained in the PR description and does not weaken any test. No issues here.

---

BenSeymourODB (commented):
First-pass advisory review of PR #246 — enforcing Prisma migration naming convention.

Overall this is a solid, well-scoped PR. The validator logic is correct for the existing migrations, the test coverage is good, and the CI wiring is in the right place. The comments below are a mix of one real developer-experience footgun in the docs (the `_prisma_migrations` rename trap), a handful of test-quality improvements, and minor nits.

**Summary by severity:**

- **Needs attention** — `docs/database.md` line 51: the rename-then-commit workflow does not mention that the local `_prisma_migrations` DB record also needs updating, which will cause `pnpm db:migrate` to fail on the next run.
- **Should fix** — `scripts/__tests__/check-migration-naming.test.ts` lines 95 and 107: two assertions are weaker than the actual behaviour warrants; tightening them would catch future regressions.
- **Nice to have** — trailing-underscore regex hole, sort-assumption comment, `invokedDirectly` idiom, hardcoded `0005_` example in docs.
- **Pre-existing / out of scope** — no `pnpm test:run` step in CI (flagged for awareness, not blocking this PR).

---

scripts/check-migration-naming.mjs
BenSeymourODB: Tightened the regex to your suggestion in `69a3e71` and added a "rejects trailing or doubled underscores" test case covering `0001_add_`, `0001___`, `0001_add__settings`, and `0001__initial`. Now matches `0001_initial` and `0004_add_calendar_settings` while rejecting both forms of underscore abuse.

---

scripts/check-migration-naming.mjs
BenSeymourODB: Added the comment in `69a3e71` so a future numeric-comparator change doesn't silently break the sequential check.

---

scripts/check-migration-naming.mjs
BenSeymourODB: Switched to `lstatSync` in `69a3e71`. A directory-typed symlink now reports `isDirectory === false` and is skipped from validation rather than silently treated as a real migration. Comment in the source explains the choice.

---

scripts/check-migration-naming.mjs
BenSeymourODB: Replaced with the standard idiom in `69a3e71`:

```js
if (process.argv[1] && fileURLToPath(import.meta.url) === process.argv[1]) {
  main();
}
```

No more cwd-dependence and no second branch to reason about.

---

scripts/__tests__/check-migration-naming.test.ts
BenSeymourODB: Tightened in `69a3e71` to `toHaveLength(1)` with explicit content assertions on the single error message:

```ts
expect(errors).toHaveLength(1);
expect(errors[0]).toContain('"0002_second"');
expect(errors[0]).toContain("expected 0003");
```

And kept the explanation comment so the next reader still sees why exactly one error is expected.

---

scripts/__tests__/check-migration-naming.test.ts
BenSeymourODB: Added the explanatory comment in `69a3e71` and an extra `expect(errors[1]).toContain("expected 0002")` so both error messages are checked, not just the count.

---

docs/database.md
BenSeymourODB: Good catch — fixed in `69a3e71`. Added a new "Renaming a freshly-created migration" subsection in `docs/database.md` that:

- Explains the `_prisma_migrations` rename trap explicitly (rename → DB row out of sync → next `pnpm db:migrate` fails with drift / "migration not found").
- Documents both fixes: rewriting the row in place via `UPDATE _prisma_migrations` and the simpler `pnpm db:migrate:reset` shortcut for dev.
- Notes that production deploys are unaffected — `prisma migrate deploy` reads the renamed directory directly and writes the row with the correct (renamed) value on first apply.

Also rewrote the example block in `### Example: Adding a New Field` to use `<NNNN>` and link to the new subsection.

---

docs/database.md
BenSeymourODB: Replaced `0005_` with `<NNNN>` in the example in `69a3e71` and pointed at `pnpm db:migrate:check-names` for confirming the actual next number.

---

.github/workflows/main_nextjs-template-build.yml
BenSeymourODB: Acknowledged — agree this is worth filing, but leaving it out of this PR per your "out of scope if the team wants to track it separately" steer. Will open a follow-up issue tracking the missing `pnpm test:run` step in CI as a separate slice so it gets its own discussion / PR rather than expanding the scope here.

---

scripts/__tests__/check-migration-naming.test.ts
BenSeymourODB: Confirmed — and the ticking-tests count went from 12 → 13 in `69a3e71` after adding the trailing-underscore case, then through to all 1585 tests in the full suite (`pnpm test:run` green locally). No changes needed.

---

scripts/__tests__/check-migration-naming.test.ts
BenSeymourODB: Tightened in `69a3e71` to assert on each error's content rather than just the count. Now a future refactor that merges or drops one of the messages would fail the test rather than silently passing on a count match:

```ts
expect(errors).toHaveLength(2);
expect(errors.some((e) => e.includes('"not-a-migration"'))).toBe(true);
expect(errors.some((e) => e.includes('"0003_skipped_two"'))).toBe(true);
```

---

src/components/calendar/__tests__/AgendaList.test.tsx
BenSeymourODB: Acknowledged — thanks for the explicit confirmation. No action needed here.

---

.github/workflows/main_nextjs-template-build.yml
BenSeymourODB: Filed as #247.

---

PR #248, opened 2026-05-03
chore(ci): run pnpm test:run before Build (#247)

BenSeymourODB:
Closes #247.

## Summary

Adds a `Test` step to `.github/workflows/main_nextjs-template-build.yml` so the Vitest suite runs on every push and PR. Without this, lint/typecheck/build can pass while a real test regression silently lands on `main`.

```yaml
      - name: Test
        run: pnpm test:run
```

Step is positioned **between** the Prisma migration validation and the production `Build` so a failing test fails the job before the slower build runs (fast-fail).

## TDD

Per CLAUDE.md, written test-first:

1. **Red** — `scripts/__tests__/ci-workflow.test.ts` parses the workflow file and asserts the `Test` step exists, runs `pnpm test:run`, and is correctly ordered (after install/lint/typecheck/migration validation, before `Build`). Initial run failed (no step).
2. **Green** — Added the workflow step. All 3 assertions pass.
3. **Verify** — `pnpm test:run` → 75 files / 1014 tests pass; `pnpm lint:fix && pnpm format:fix && pnpm check-types` clean.

The test guards against future edits that might accidentally drop or reorder the step.

## Acceptance criteria

- [x] `Test` step added to `.github/workflows/main_nextjs-template-build.yml` before the `Build` step
- [ ] CI green on this PR confirms the step runs (will be visible once CI runs)
- [ ] CI red on a PR that intentionally breaks one assertion confirms the step fails the job — verified locally; not landed on `main`'s history.

## Notes on issue scope

The issue body says "between the migration-naming check and the build step", referring to a check added by **#246** (still open). PR #246 hasn't merged yet, so the workflow on `main` doesn't yet have a migration-naming check; this PR places the `Test` step immediately after the existing migration-validation step. When #246 merges, the conflict is trivial (the migration-naming check slots in before `Test`).

## Out of scope

E2E (Playwright) — explicitly deferred per the issue body. It needs a running dev server + browser binaries and belongs in its own workflow.

https://claude.ai/code/session_01BhwNtrQoDvVFBghD5AUbZq

---
_Generated by [Claude Code](https://claude.ai/code/session_01BhwNtrQoDvVFBghD5AUbZq)_


BenSeymourODB (commented):
Overall: the change is correct and the TDD approach is sound. Three findings below — one blocking concern (no `DATABASE_URL` env var in the CI workflow), one moderate test-robustness issue, and one minor nit. Everything else is clean.

---

.github/workflows/main_nextjs-template-build.yml
```diff
@@ -41,5 +41,8 @@ jobs:
             exit 1
           }
 
+      - name: Test
```
BenSeymourODB, line 44: **Blocking: `DATABASE_URL` is not set in CI, and `pnpm test:run` may need it.**

`src/lib/db.ts` constructs a `PrismaPg` adapter at module-load time using `process.env.DATABASE_URL`. All 14 test files that import from `@/lib/db` use `vi.mock("@/lib/db", ...)`, which Vitest hoists above the real import — so in the happy path the real `PrismaPg` constructor is never called and no live database is needed.

However, this is an implicit dependency: if `vi.mock` hoisting ever fails (e.g. a test added without the mock, or a Vitest version bump that changes hoisting semantics), the suite will throw at import time with no DATABASE_URL rather than a helpful assertion failure.

More immediately: the existing `Build` step (`pnpm build`) also runs without `DATABASE_URL` and apparently succeeds today — so if the repo already works in CI without it, the same will be true for `pnpm test:run`. If CI is currently green, no action is needed here. **But if CI fails on the `Test` step with a Prisma/connection error, add a stub env var to the workflow:**

```yaml
      - name: Test
        env:
          DATABASE_URL: "postgresql://postgres:postgres@localhost:5432/calendar_db"
          NEXTAUTH_SECRET: "ci-stub-secret"
        run: pnpm test:run
```

This is a dummy value — no real DB is started, and all DB calls are mocked — but it satisfies Prisma's adapter constructor and any NextAuth module-load checks. Recommend verifying CI passes before merging.

---

scripts/__tests__/ci-workflow.test.ts
```diff
@@ -0,0 +1,49 @@
...
+import { describe, expect, it } from "vitest";
+
+const WORKFLOW_PATH = path.resolve(
+  __dirname,
+  "../../.github/workflows/main_nextjs-template-build.yml"
+);
+
+const workflow = readFileSync(WORKFLOW_PATH, "utf-8");
+
+function indexOfStep(name: string): number {
```
BenSeymourODB, line 19: **Moderate: `indexOfStep` will false-positive on prefix-matched step names.**

```ts
function indexOfStep(name: string): number {
  return workflow.indexOf(`- name: ${name}`);
}
```

`indexOfStep("Test")` matches `- name: Test` today, but would also match `- name: Testing` or `- name: Test coverage` if one were added in the future. Tighten the match to require either a newline or end-of-string immediately after the step name:

```ts
function indexOfStep(name: string): number {
  const match = workflow.match(new RegExp(`- name: ${name}\\s*\\n`));
  return match?.index ?? -1;
}
```

Or keep the current `indexOf` but include a trailing newline in the search string:

```ts
return workflow.indexOf(`- name: ${name}\n`);
```

This also makes the ordering assertions immune to a step whose name is a prefix of another (e.g. `Build` vs `Build and cache`).

---

scripts/__tests__/ci-workflow.test.ts
```diff
@@ -0,0 +1,49 @@
+import { readFileSync } from "node:fs";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+
+const WORKFLOW_PATH = path.resolve(
+  __dirname,
+  "../../.github/workflows/main_nextjs-template-build.yml"
+);
+
+const workflow = readFileSync(WORKFLOW_PATH, "utf-8");
```
BenSeymourODB: **Minor: `readFileSync` at module scope causes an opaque crash rather than a test failure if the path is wrong.**

```ts
const workflow = readFileSync(WORKFLOW_PATH, "utf-8");
```

If `WORKFLOW_PATH` is ever wrong (e.g. someone renames the workflow file), Vitest reports the entire test file as errored with `ENOENT`, not a failing assertion with a useful message. Moving the read inside the `describe` block (or even into a `beforeAll`) would give the same behaviour but with a cleaner error:

```ts
let workflow: string;
beforeAll(() => {
  workflow = readFileSync(WORKFLOW_PATH, "utf-8");
});
```

This is genuinely minor — the path is currently correct and `__dirname` is reliable in Vitest — but worth noting for long-term maintainability.

---

BenSeymourODB (commented):
Three findings (details in inline comments):

1. **Blocking (workflow line 44)** — No `DATABASE_URL` is set in the CI job. The test suite is almost certainly fine today (all DB-touching tests mock `@/lib/db` via `vi.mock`, and the existing `Build` step already runs without it), but this is an implicit assumption. Recommend letting CI run and confirming the `Test` step goes green; if it doesn't, add a stub `env:` block to the step.

2. **Moderate (test line 12)** — `indexOfStep` uses a bare `indexOf` prefix match, so `indexOfStep("Test")` would also match a future step named `Testing` or `Test coverage`. Tighten to match a trailing newline.

3. **Minor (test line 10)** — `readFileSync` at module scope makes the whole file crash with `ENOENT` (not a failing test) if the path is ever wrong. Moving it inside `beforeAll` is cleaner.

The workflow placement and ordering are correct. The assertions cover all three requirements from the issue (step exists, runs the right command, is ordered correctly). The TDD narrative in the PR description is accurate.

---

.github/workflows/main_nextjs-template-build.yml
BenSeymourODB: Holding off on the stub `env:` block for now per your "no action needed if CI passes" guidance.

Context for the previous CI red:
- The first CI run failed in the **Lint and type check** step (not the new `Test` step) on the pre-existing `ICalendarContext` regression that PR #244 was opened to fix. That regression was fixed in main by the merge commit `8ef8443` (PR #226), which added the missing `deleteEvent: vi.fn().mockResolvedValue(undefined)` to `AgendaList.test.tsx`'s `makeContext()`.
- This branch was based on a stale main (b79f22b). I've now merged current `origin/main` into the branch. Local `pnpm test:run` against the merged tree: **108 files / 1575 tests pass**, including the 14 files that import from `@/lib/db` (all of which mock it via `vi.mock`).

If the next CI run fails on the `Test` step with a Prisma/PG connection error, I'll add the stub `env:` block exactly as you've written it. Otherwise, leaving the workflow unchanged in this PR keeps the diff focused.

---

scripts/__tests__/ci-workflow.test.ts
BenSeymourODB: Fixed in `a206404`. Took the simpler `\n` suffix variant (`workflow.indexOf(`- name: ${name}\n`)`) over the regex — same protection against `Test` vs `Test coverage` / `Build` vs `Build and cache` false-positives, no regex escaping concerns, and a one-line diff. Added a comment above the function explaining why the trailing `\n` matters so a future cleanup doesn't accidentally drop it.

All 3 assertions still pass against the updated workflow.

---

scripts/__tests__/ci-workflow.test.ts
BenSeymourODB: Fixed in `a206404`. Moved the `readFileSync` into a `beforeAll` and made `workflow` a top-level `let`. If `WORKFLOW_PATH` is ever wrong now, Vitest reports it as a `beforeAll` failure attached to the suite (with a useful `ENOENT: ...` message and the offending path), and the individual `it` blocks show as skipped — much more navigable than a top-of-file crash.

---

PR #249, opened 2026-05-03
fix(tasks): structured 403 + requiresReauth when Google Tasks scope missing (#237)

BenSeymourODB:
Closes #237.

## Summary

`POST /api/tasks` (and the other tasks API routes) returned 403 for signed-in users when the stored Google OAuth grant predated the `https://www.googleapis.com/auth/tasks` scope being added to `auth.ts`. Google's Tasks API rejects calls with 403 in that case, the existing route forwarded the status verbatim with a generic `"Failed to create task"` body, and the client had no signal to prompt re-authentication. This PR makes the failure recoverable end-to-end.

### Phase 1 — auth-layer primitives (commit `a5ee6dd`)

- `GOOGLE_TASKS_SCOPE` constant — single source of truth, matches `auth.ts`.
- `accountHasScope(account, scope)` — splits the stored grant on whitespace or `+` (Google's URL-encoded form) and matches whole tokens, so `tasks.readonly` does not falsely satisfy a `tasks` check.
- `assertGoogleTasksScope()` — composes session + scope checks. Throws `AuthError(401)` for unauthenticated / `RefreshTokenError`, `AuthError(403)` with a "re-authentication required" message when the tasks scope is absent.
- 12 new vitest cases in `helpers.test.ts`.

### Phase 2 — pre-flight scope check + structured 403 (commit `e135ae4`)

Wired `assertGoogleTasksScope()` into every tasks route immediately after the existing session checks: `GET /api/tasks`, `POST /api/tasks`, `GET /api/tasks/lists`, `PATCH /api/tasks/[taskId]`, `POST /api/tasks/[taskId]/complete`. Users with stale grants now get a structured 403 *without* burning an upstream call. Also widened the `instanceof AuthError` branch in each route's handler to mark 403 (not just 401) as `requiresReauth: true` — both states are recoverable by re-authenticating, so the client gets a consistent signal. The existing `GoogleTasksApiError` 403 forwarding stays as a defensive second line for grants revoked or downgraded server-side. One new vitest case per route asserts: status 403, `requiresReauth: true`, error mentions "Google Tasks", upstream fetch never called.

### Phase 3 — re-auth CTA in `NewTaskModal` (commit `4a7f1d1`)

- `useCreateTask` now rejects with a typed `TaskApiError` carrying `status` + `requiresReauth` instead of a plain `Error`. Existing tests that match `rejects.toThrow(/regex/)` still pass since `TaskApiError extends Error`.
- `NewTaskModal` reads the typed error: when `requiresReauth` is true it renders a "Sign in again" outline button beneath the inline error message. Clicking calls `signIn("google", { callbackUrl })`, which forces fresh consent and refreshes the OAuth grant.
- Component test asserts the CTA renders + invokes `signIn` for `TaskApiError(403, requiresReauth=true)`; a second asserts the CTA is absent on a generic `Error` so unrelated failures don't show a misleading button.

## Out of scope / deferred

- Other tasks consumers (`useTasks`, `TaskList`, `useUpdateTask`) keep the existing generic-Error path. `NewTaskModal` is the user-visible flow #237 calls out as broken; expanding the typed error to the read paths is straightforward follow-up.
- E2E coverage for the CTA is limited by the OAuth redirect target — the only meaningful Playwright assertion would be the same one the component test already makes (CTA exists, click invokes `signIn`). Skipped to keep the slice tight.

## Test plan

- [x] `pnpm test:run` — 107 files / 1592 tests pass (16 new across the three phases)
- [x] `pnpm lint && pnpm check-types` — clean (only pre-existing warnings)
- [x] All four tasks routes covered (32 existing + 4 new pre-flight cases)
- [x] `useCreateTask` covered (10 existing + 1 new TaskApiError case)
- [x] `NewTaskModal` covered (22 existing + 2 new re-auth-CTA cases)

https://claude.ai/code/session_01DGVsVJNQNXULZezpQBeMPQ


src/lib/auth/helpers.ts
```diff
@@ -21,6 +21,30 @@ function decryptAccountTokens<T extends NonNullable<AccountRow>>(
...
+ * NextAuth Prisma adapter has historically stored either form.
+ */
+export function accountHasScope(
+  account: { scope?: string | null } | null | undefined,
+  scope: string
+): boolean {
+  if (!account?.scope) return false;
+  // Whole-token match — split on whitespace or `+` so substring-only matches
+  // (e.g. `tasks.readonly` when checking for `tasks`) do not falsely succeed.
+  const granted = account.scope.split(/[\s+]+/).filter(Boolean);
```
BenSeymourODB: The regex `/[\s+]+/` does not cover comma-separated scopes. Google's token-exchange response uses spaces, but the OAuth redirect URL uses `+`-encoding, and some versions of the Prisma adapter or manual DB seeds may have stored comma-delimited strings. A stored value of `"openid,email,https://www.googleapis.com/auth/tasks"` will never match because commas are not in the character class.

Suggest broadening to `/[\s,+]+/` (or at minimum adding a test that exercises a comma-delimited fixture) until you can confirm from existing production data that commas never appear.

---

src/lib/auth/helpers.ts
```diff
@@ -156,6 +180,34 @@ export async function withAuth<T>(
...
+export async function assertGoogleTasksScope(): Promise<void> {
+  const session = await getSession();
+  if (!session?.user?.id) {
+    throw new AuthError("Not authenticated", 401);
+  }
+  if (session.error === "RefreshTokenError") {
+    throw new AuthError("Session expired. Please sign in again.", 401);
+  }
+
+  const account = await getGoogleAccount();
```
BenSeymourODB, line 205: Every happy-path request to `/api/tasks` (GET or POST) now hits the database **three times** via this call chain:

1. `getSession()` — line 32/100 in the route
2. `assertGoogleTasksScope()` → `getSession()` → `getGoogleAccount()` → `getSession()` + `prisma.account.findMany`
3. `getAccessToken()` — `getSession()` + `prisma.account.findMany` (again)

`assertGoogleTasksScope` already fetches the full decrypted account but throws away the `access_token`. If it returned the token (or the `account` row) instead of `void`, the route could skip the `getAccessToken()` call entirely, halving the DB reads per request.

---

src/lib/auth/helpers.ts
```diff
@@ -156,6 +180,34 @@ export async function withAuth<T>(
...
+  if (!account) {
+    throw new AuthError("No Google account linked.", 401);
+  }
+
+  if (!accountHasScope(account, GOOGLE_TASKS_SCOPE)) {
```
BenSeymourODB, line 210: There is a gap between what this check detects and what actually causes a 403 from Google. `accountHasScope` reads the `scope` string that was persisted at the time of the original consent grant — it does **not** verify that the current `access_token` was issued with those scopes. If Google silently issued a narrower token on refresh (e.g. after a scope downgrade by the user in their Google account), the pre-flight passes but the upstream call still 403s. The downstream `GoogleTasksApiError` handler in the routes does catch that case, but the comment on `assertGoogleTasksScope` implies it's authoritative — worth narrowing the claim to "checks the stored grant" so future readers don't assume a network-level guarantee.

---

src/components/tasks/new-task-modal.tsx
```diff
@@ -254,13 +255,31 @@ export function NewTaskModal({
...
+                  <Button
+                    type="button"
+                    size="sm"
+                    variant="outline"
+                    onClick={() =>
+                      signIn("google", {
+                        callbackUrl:
+                          typeof window === "undefined"
+                            ? "/"
+                            : window.location.pathname,
```
BenSeymourODB: `window.location.pathname` drops query params and hash. If the user landed on `/calendar?date=2026-05-04&view=week`, after re-auth they'll be returned to `/calendar` and lose their selected view. Use `window.location.href` (or `window.location.pathname + window.location.search`) to preserve the full URL.

There is no open-redirect risk here because NextAuth validates `callbackUrl` against the configured `NEXTAUTH_URL` origin.

---

BenSeymourODB (commented):
## Review summary

The direction is correct — structured 403 + `requiresReauth` is a clean pattern, the mock setup is consistent across all five route test files, and the `TaskApiError` class gives callers a typed way to branch without parsing message strings. Three issues need addressing before merge.

### Blocking

**Scope delimiter coverage** (`src/lib/auth/helpers.ts` line 44) — the regex `/[\s+]+/` does not handle comma-separated scope strings. If any existing row in the DB has scopes delimited by commas (possible via manual seeding, the Prisma adapter's historic behaviour, or a URL-decoded `%20`-encoded string that was stored literally), `accountHasScope` returns `false` and every Tasks request 403s for that user. The fix is one character (`/[\s,+]+/`), but first add a test fixture with a comma-delimited scope string to confirm the scope of the risk.

**`callbackUrl` drops query params** (`src/components/tasks/new-task-modal.tsx` line 275) — `window.location.pathname` discards query params and the hash. A user on `/calendar?date=2026-05-04&view=week` is returned to `/calendar` after re-auth, losing their view. Change to `window.location.pathname + window.location.search` (or `window.location.href`). There is no open-redirect risk here because NextAuth validates `callbackUrl` against `NEXTAUTH_URL`.

### Non-blocking (follow-up)

**Redundant DB reads** (`src/lib/auth/helpers.ts` line 198) — each happy-path tasks request makes 2 `prisma.account.findMany` calls and 3+ `auth()` calls: one explicit `getSession()` in the route, then `assertGoogleTasksScope` calls `getSession()` + `getGoogleAccount()` (which calls `getSession()` again + a DB query), then `getAccessToken()` calls `getSession()` + another DB query. `assertGoogleTasksScope` already has the decrypted account in hand; returning the token instead of `void` would let the route skip `getAccessToken()` entirely. Not a correctness issue at this scale, but worth a follow-up issue.

**`useTasks` / `updateTask` swallow `requiresReauth`** — `use-tasks.ts` (unchanged in this PR) throws a plain `Error` on non-OK responses, so the "Sign in again" CTA only surfaces inside `NewTaskModal`. Fetch failures and task-completion failures hit the same 403 routes but the signal is lost. A follow-up issue to promote `useTasks` errors to `TaskApiError`-equivalents would close the gap cleanly.

---

src/lib/auth/helpers.ts
BenSeymourODB: Fixed in c9c2e11. Broadened the split to `/[\s,+]+/` and added a comma-delimited fixture test (`accountHasScope > returns true when the scope is present with comma separators`) so any future regression to the narrower form fails loudly.

---

src/lib/auth/helpers.ts
BenSeymourODB: Good catch. Deferred to follow-up issue #260 — the refactor crosses all five `/api/tasks*` routes (each currently calls both `assertGoogleTasksScope()` and `getAccessToken()`) and changes the helper signature, which felt too large to land in this PR alongside the bug-fix slice. Tracked it with the option-1 vs option-2 trade-off you implied (return the token from the assert, vs introduce a new combined `requireGoogleTasksAccess` helper) so the next pass can pick.

---

src/lib/auth/helpers.ts
BenSeymourODB: Fixed in c9c2e11. Tightened the doc comment on `assertGoogleTasksScope` to spell out that it reads the persisted grant string only, and called out the silent-narrower-token-on-refresh case explicitly so future readers know the `GoogleTasksApiError(403)` fallback in routes is still load-bearing rather than belt-and-braces.

---

src/components/tasks/new-task-modal.tsx
BenSeymourODB: Fixed in c9c2e11. `callbackUrl` now uses `pathname + search` so the user lands back on their full view (including `?date=…&view=week`). Added a test that drives jsdom's location to `/calendar?date=2026-05-04&view=week`, clicks the CTA, and asserts the exact preserved value. Confirmed there's no open-redirect risk per your note — NextAuth validates `callbackUrl` against `NEXTAUTH_URL`.

---

Issue #220, opened 2026-05-02
a11y: extend keyboard navigation, ARIA, and focus management to remaining surfaces + WCAG audit


BenSeymourODB:
## Context

PR #153 (ARIA grid + keyboard nav for month view) explicitly scoped only to month-view; multiple companion items were deferred. Combined here so the a11y backlog is one tracked unit.

### Deferred items collected from review threads:

- **PR #153** — focus management for modal dialogs (largely covered by Radix/shadcn); screen-reader audit across non-calendar components; WCAG 2.1 AA contrast audit + font-scaling controls; keyboard nav for `AgendaCalendar`, `MiniCalendarSidebar`, week/day/year views
- **PR #140** — ARIA grid semantics + full-name day-of-week labels for `MiniCalendarSidebar` (https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/140#discussion_r3114733065)
- **PR #149** — `role="button"` + keyboard handling for `WeekCalendar` event pills (post click-to-edit landing) (https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/149#discussion_r3119922640)
- **PR #145** — month-name `<button>` aria-labels and dot-container `aria-hidden` in `YearCalendar` (https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/145#discussion_r3118199620)

## Scope

This is best handled as a track of small focused PRs:

1. ARIA grid semantics for `MiniCalendarSidebar` + full-name day labels
2. ARIA labels and `aria-hidden` cleanup in `YearCalendar`
3. `role="button"` + keyboard handling for `WeekCalendar` / `DayCalendar` event pills
4. Keyboard nav (arrow-key) for `AgendaCalendar` + `MiniCalendarSidebar` + week/day/year views (matching #73's month pattern)
5. Focus-management audit for all modal dialogs (verify Radix/shadcn defaults are sufficient; document gaps)
6. Screen-reader audit across non-calendar surfaces (settings, profile selector, scheduler controls)
7. WCAG 2.1 AA contrast audit; add font-scaling control to settings

## Acceptance criteria

- [ ] Each item has its own short PR; this issue stays open until all are merged
- [ ] Axe + Lighthouse a11y scores ≥ 95 on `/calendar`, `/settings`, `/profiles`
- [ ] Keyboard-only smoke test on each calendar view recorded as a checklist
- [ ] WCAG audit report saved under `docs/a11y/`

## Dependencies

[...]

PR #251, opened 2026-05-03
fix(calendar): YearCalendar correctness bugs — timezone, multi-year, perf, nav (#203)

### LINKED ISSUE #203: fix(calendar): YearCalendar correctness bugs — timezone, multi-year overlap, perf, navigation
## Context

Pending review on PR #145 surfaced four distinct YearCalendar bugs that landed unaddressed (review #4148502467 was filed but never submitted):

1. **All-day timezone bug** — `new Date(event.startDate)` parses bare `YYYY-MM-DD` strings as UTC, so an all-day event on June 15 in negative-offset zones renders the dot on June 14. https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/145#discussion_r3118195229
2. **Multi-year overlap missed** — `yearEventCount` filters only by `startDate`, so an event running Dec 2025 → Jan 2026 is invisible in the 2026 count. Use `getEventsForYear` overlap logic instead. https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/145#discussion_r3118196192
3. **Per-day perf** — `getUniqueColorsForDay` iterates the full event array per day cell (up to 366 cells × N events). Pre-bucket into `Map<string, Set<TEventColor>>` keyed on `formatDayKey`. https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/145#discussion_r3118197605
4. **Year-nav semantics** — Replace the current arithmetic with `new Date(year - 1, 0, 1)` (January 1) for prev/next-year navigation; add a test asserting the resulting year. https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/145#discussion_r3118198829

## Acceptance criteria

- [ ] All-day events render their dot on the correct day across timezones (regression test with `TZ=America/Los_Angeles`)
- [ ] `yearEventCount` reuses `getEventsForYear` (or equivalent overlap logic)
- [ ] Pre-bucketing reduces per-render iterations to O(events + cells) instead of O(events × cells)
- [ ] Prev/next year navigation lands on Jan 1 of the target year; tested by year value
- [ ] Accessibility improvements (separate issue) tracked elsewhere

## Notes

Independent. Should land before the year view goes into heavy production use, but does not block it from being wired (#117).

BenSeymourODB:
Closes #203.

## Summary

Fixes the four YearCalendar correctness bugs from #145's deferred review (#203). Each bug is delivered in its own commit with a regression test that exercises the broken path before the fix lands.

Plan: [`.claude/plans/yearcalendar-correctness-203.md`](./.claude/plans/yearcalendar-correctness-203.md).

## Phases

- [x] **Bug 1** — All-day events render their dot on the correct day across timezones (`parseEventStartLocal` treats bare `YYYY-MM-DD` as local time)
- [x] **Bug 2** — `yearEventCount` reuses `getEventsForYear` overlap logic so Dec → Jan events are counted in both years
- [x] **Bug 3** — Pre-bucketing reduces per-render iterations from O(events × cells) to O(events + cells) (`bucketEventColorsByDayKey`)
- [x] **Bug 4** — Prev/next year navigation lands on Jan 1 of the target year (drops the `selectedDate.getMonth()` carry-over)

## Test plan

- [x] `YearCalendar.test.tsx` — regression tests for all four bugs
- [x] `pnpm test` — 1894/1894 passing
- [x] `pnpm lint:fix` — 0 errors (5 pre-existing warnings unrelated)

## Out of scope

- The `today`-highlight midnight-tick treatment (#204 / PR #225 territory).
- The full-year fetch path (#117 / PR #196 territory).
- Accessibility improvements (tracked in #220).

Tracking: #224.

---
🤖 Generated by [Claude Code](https://claude.ai/code)


BenSeymourODB (commented):
## Review summary

Reviewed this PR end-to-end: read the full diff, the plan (`.claude/plans/yearcalendar-correctness-203.md`), and the related helpers in `src/lib/calendar-helpers.ts`. Ran `pnpm install`, `pnpm test` (1894/1894 ✅), `pnpm check-types` (clean ✅), and `pnpm lint` (only pre-existing warnings; none in changed files).

### Verdict

**Functionally correct, merge-ready.** Each of the four bugs is addressed exactly as scoped, with one commit per bug and a regression-style test for each. CLAUDE.md compliant (no manual memoization, no `any`, standard Tailwind colors). CI is green and the merge state is clean.

### Findings

**🟡 CONCERN — `src/components/calendar/__tests__/YearCalendar.test.tsx:573-603` and `:510-527` — Bug-1 regression tests are UTC-transparent**

Both the component-level test (`renders the dot on the calendar-local day for an all-day bare-date startDate`) and the lower-level bucketer test (`treats bare YYYY-MM-DD startDates as the local calendar day`) only catch the bug under a **negative-offset timezone**. Verified locally:

```text
new Date("2026-06-15")  in UTC → 2026-06-15 (no slip)
new Date("2026-06-15")  in PT  → 2026-06-14 (slip — the bug)
```

The PR runs CI on `ubuntu-latest` with no `TZ` env override (`.github/workflows/main_nextjs-template-build.yml:48`) and `vitest.setup.ts` doesn't set `process.env.TZ` either. Under UTC, the bare-date input produces key `"2026-06-15"` from **both** the buggy code and the fixed code — the test passes either way. A future revert to `new Date(event.startDate)` would not be caught.

The plan acknowledged this trade-off ("Pinning the harness timezone via `process.env.TZ` is unreliable in vitest's worker model") — but on Node 18+, setting `TZ` in `vitest.setup.ts` does work reliably. Suggested follow-up: add `process.env.TZ = "America/Los_Angeles"` to `vitest.setup.ts` so all timezone-sensitive paths are exercised by every test run. Worth a dedicated tracking issue rather than rolling into this PR (risk of fallout across the 1894 tests).

**🟢 NIT — `src/components/calendar/YearCalendar.tsx:224-226` — React Compiler memoization comment is slightly overstated**

```tsx
// Bucket once per render: each day cell then resolves its dots via an
// O(1) Map.get instead of scanning the full event array (was up to
// 366 cells × N events). React Compiler memoises identity automatically.
const colorsByDayKey = bucketEventColorsByDayKey(events);
```

The "React Compiler memoises identity automatically" claim is conditional — it requires `events` reference to be stable across renders. If `CalendarProvider` returns a new array on every render (common with filter/map chains in providers), the Compiler will rebuild the Map each render. The function is O(N) so this is **not** a perf concern (the original O(N×cells) bug is still solved), but the comment overpromises. Suggest dropping the last sentence or rephrasing as "Cheap enough (O(events)) that per-render is fine; React Compiler may memoise if `events` identity is stable."

**🟢 NIT — `src/components/calendar/YearCalendar.tsx:51-64` — `parseEventStartLocal` is local to YearCalendar.tsx but the bug is broader**

Same naive `new Date(event.startDate)` pattern exists in:
- `src/components/calendar/AgendaList.tsx:44, 50, 57, 126`
- `src/components/calendar/AgendaCalendar.tsx:23, 55-56, 78, 184`
- `src/components/calendar/SimpleCalendar.tsx:135, 145`
- `src/components/calendar/EventDetailModal.tsx:66, 72`
- `src/components/calendar/DayOverflowPopover.tsx:34`
- `src/components/calendar/analog-clock/clock-utils.ts:116, 139`

All of these will exhibit the same TZ slip for bare-date all-day events from non-canonical sources. Out of scope for #203 (which is explicitly scoped to YearCalendar), but worth a follow-up issue + a shared helper exported from `calendar-helpers.ts`. Note: `getEventsForYear` in `calendar-helpers.ts` is **already** TZ-safe because it uses `date-fns` `parseISO`, which parses bare dates as local — so the bug-2 fix in this PR is correct.

**🟢 NIT — `src/components/calendar/YearCalendar.tsx:71-85` — Bucket contract is "start day only"; not a regression but worth documenting**

`bucketEventColorsByDayKey` keys only on `event.startDate`, so a multi-day event (e.g. a Mon–Wed vacation) shows a dot on Mon only — never on Tue/Wed. This matches the pre-fix behavior of `getUniqueColorsForDay` (which also only matched start-day via `isSameDay`), so **no regression**. But the comment "the map shape is the contract that lets MonthPanel render dots without ever touching the raw event list" implies completeness. Suggest adding a one-line note: `// NOTE: only the start day is bucketed — multi-day events are not dotted on intermediate days. Matches pre-fix behavior; tracked separately.`

### Recommendation

**Approve and merge as-is.** The four bugs are correctly fixed. The findings above are either (a) a known test-tightness gap better addressed by a separate infrastructure PR (the TZ override in `vitest.setup.ts`), or (b) micro-comment polish. None block merging this slice of work.

Suggested follow-ups, none blocking:
- One issue to set `TZ=America/Los_Angeles` in `vitest.setup.ts` (and audit the resulting test fallout).
- One issue to extract a shared `parseEventStartLocal` (or `parseIEventStartLocal`) helper into `calendar-helpers.ts` and migrate the seven other call-sites listed above.

---
🤖 Reviewed by `/implement-issue` via Claude Code (Opus 4.7)


---

BenSeymourODB (commented):
## In-depth review

Verified locally on `7147703` cleanly merged with current `main` (`041ef63`):
- `pnpm install --frozen-lockfile` — clean
- `pnpm check-types` — clean
- `pnpm lint` — 0 errors, 6 pre-existing warnings (unrelated)
- `pnpm test` — **136 files / 2133 tests green** (no regressions from the four fixes)

### Rebase / merge state

Branch base was 30 days stale (last update 2026-05-03, current `main` at `041ef63`). Verified that a `git merge origin/main` into the branch results in **zero conflicts** — the four commits touch only `YearCalendar.tsx`, `YearCalendar.test.tsx`, and the plan file, none of which received concurrent edits on `main`. The test-fixture additions on `main` (`workingHoursStart`, `transitionDurationMs` fields on `ICalendarContext`) merge cleanly with the PR's test-file growth. Pushing the merge commit after this review to get fresh CI on the up-to-date base.

### Bug-by-bug verification

**Bug 1 — All-day timezone (`c75f039`)**
The `BARE_DATE_RE` + `parseEventStartLocal` shape is the right fix. The component's only direct `event.startDate` parse paths now all go through it (the `bucketEventColorsByDayKey` map keys + the indirect MonthPanel lookup). I confirmed by greping the file: no remaining `new Date(event.startDate)` calls in the dot-rendering path. Test fixture (`startDate: "2026-06-15"`, no `T00:00:00`) is exactly the input shape that exposes the bug — the symmetry test on `2026-06-14` is the right negative assertion.

**Bug 2 — Multi-year `yearEventCount` overlap (`3c07cae`)**
The fix delegates to `getEventsForYear` in `src/lib/calendar-helpers.ts:419`, which uses the canonical `eventStart <= endOfYear && eventEnd >= startOfYear` overlap formula. Both directions are tested (Dec→Jan event counts in both 2025 and 2026 views). One subtle asymmetry between this and Bug 1's fix — see inline comment.

**Bug 3 — Pre-bucketing (`c0b4695`)**
The `Map<dayKey, Set<TEventColor>>` shape drops the inner loop from O(events × cells) to O(events + cells). The `bucketEventColorsByDayKey` export gives the unit test a clean seam without bleeding implementation into the public surface. React Compiler will memoise the map identity since `events` is the sole dependency.

**Bug 4 — Year-nav semantics (`7147703`)**
Trivial substitution. Test asserts month=0 + date=1 for both directions, which is exactly what was missing from the prior "year change only" check.

### Architectural observations (non-blocking)

- **Asymmetry between dot path and count path for bare-date events** — see inline comment on `YearCalendar.tsx:232`. The fix is consistent within each commit but the two paths now use different date-parsing strategies, which is observable at Jan-1/Dec-31 boundaries in negative-offset zones.
- **Multi-day event dots (pre-existing, NOT introduced here)** — both the old `getUniqueColorsForDay` and the new `bucketEventColorsByDayKey` key only off `event.startDate`. A Jun 14→Jun 16 event still renders a dot only on Jun 14, not Jun 15/16. The plan explicitly stays out of this (#117 / #224 territory) and the PR's scope is correct.

### Verdict

Ready to merge after the merge commit pushes and CI re-runs. The four fixes are correct, well-scoped, individually testable, and the regression suite is appropriately targeted. The two architectural observations above are follow-up material, not blockers.

---
_Generated by [Claude Code](https://claude.ai/code) — `/implement-issue` PR-review fallback_

---

src/components/calendar/YearCalendar.tsx
```diff
@@ -183,16 +221,26 @@ export function YearCalendar() {
...
-  ).length;
+  // Bucket once per render: each day cell then resolves its dots via an
+  // O(1) Map.get instead of scanning the full event array (was up to
+  // 366 cells × N events). React Compiler memoises identity automatically.
+  const colorsByDayKey = bucketEventColorsByDayKey(events);
+
+  // Use overlap logic so an event spanning Dec → Jan is counted in both
+  // years. Filtering only by `startDate` would drop the Dec 30 → Jan 2
+  // case from the Jan-side year's count.
+  const yearEventCount = getEventsForYear(events, selectedDate).length;
```
BenSeymourODB, line 232: **Observation (non-blocking, follow-up worthy):** `getEventsForYear` internally uses `parseISO(event.startDate)` (calendar-helpers.ts:426), which parses bare `YYYY-MM-DD` as UTC midnight — the exact pitfall Bug 1 fixed in the dot-rendering path via `parseEventStartLocal`. The two paths now disagree at year boundaries in negative-offset zones:

- Bare-date all-day event `startDate: "2026-01-01"` in `America/Los_Angeles` (UTC-8):
  - **Dot path** (this PR, Bug 1): local-parses to `2026-01-01` → dot renders on Jan 1 ✓
  - **Count path** (this PR, Bug 2): `parseISO("2026-01-01")` → `2025-12-31T16:00 local` → falls outside `startOfYear(2026)` local bound → **not counted**

The count would silently undercount bare-date all-day events at year boundaries while their dots render correctly. The existing Bug 2 tests use full ISO timestamps (`new Date(2026, 0, 5, 9).toISOString()`) so they don't expose this.

Two ways to close the loop:
1. Push `parseEventStartLocal` into `calendar-helpers` as a shared `parseEventStart()` and have `getEventsForYear` / `getEventsForMonth` / `getEventsForDay` use it. Fixes the asymmetry everywhere.
2. Inside `YearCalendar`, normalize bare-date events before passing into `getEventsForYear`.

Out of scope for this PR — flag for a separate follow-up (could be subsumed into #117's full-year fetch work). Not blocking; the asymmetry is at most a one-event undercount at the Jan-1/Dec-31 boundary for non-Google sources.

---

src/components/calendar/YearCalendar.tsx
```diff
@@ -47,28 +48,65 @@ function formatDayKey(date: Date): string {
...
+  }
+  return new Date(event.startDate);
+}
+
+// Pre-bucket events into a Map<dayKey, Set<TEventColor>> so each day cell
+// reduces to an O(1) lookup instead of scanning the whole event array.
+// Drops the year-grid render from O(events × cells) to O(events + cells).
+// Exported for unit testing — the map shape is the contract that lets
+// MonthPanel render dots without ever touching the raw event list.
+export function bucketEventColorsByDayKey(
```
BenSeymourODB, line 71: **Nit (style):** `bucketEventColorsByDayKey` is exported only "for unit testing" per the comment. The unit test on the export already passes via the indirect public-API tests (`renders a dot for a day that has an event`, etc. — they cover the bucket → MonthPanel pipeline end-to-end). The two direct tests at `YearCalendar.test.tsx:478-528` add a marginal amount of coverage beyond what the indirect tests already provide.

If you wanted to keep this fully internal, the two direct tests collapse cleanly into the existing indirect ones (the bare-date case is already covered by the `All-day timezone correctness` block). Not a blocker — the explicit unit on the bucket function is the cheapest way to lock in the Map contract that PR's Bug 3 introduces, and that explicit contract has documentation value even if it duplicates coverage. Tradeoff call; either choice is fine.

---

BenSeymourODB (commented):
Two inline observations. Both non-blocking — flagged for the maintainer's awareness.

---

src/components/calendar/YearCalendar.tsx
BenSeymourODB: **Disposition:** filed as #375 — `fix(calendar): unify bare-date parsing between dot-rendering and getEventsForX helpers`. The followup widens the fix to all four `getEventsForX` helpers (Year/Month/Week/Day) since the same `parseISO` vs local-parse asymmetry lives in each.

Should land **after** this PR merges to avoid file-level conflict on `calendar-helpers.ts`. Not blocking this PR.

---

src/components/calendar/YearCalendar.tsx
BenSeymourODB: **Disposition:** no change requested — the comment explicitly noted "either choice is fine". The explicit unit on the bucket export has documentation value that justifies the minor duplication with the indirect public-API tests. Not filing a follow-up issue for this; pure style tradeoff.

---

PR #252, opened 2026-05-03
feat(auth): singleflight Google OAuth token-refresh queue (#216)

BenSeymourODB:
Closes #216 (refresh-queue slice — idle-session-timeout half tracked separately).

## Why

Two concurrent `session` callbacks for the same user with an expired access token both race to hit `oauth2.googleapis.com/token` and both write to `prisma.account`. The second `update()` clobbers the first, and the loser's rejected `refresh_token` can surface as a bogus `RefreshTokenError` even though the winner just succeeded.

This PR adds a per-userId singleflight cache so concurrent callers `await` the same in-flight refresh promise: exactly one OAuth round-trip and exactly one DB write per stale token, regardless of how many concurrent API requests trigger it.

### `src/lib/auth/refresh-session-tokens.ts`

Adds a thin singleflight wrapper alongside the existing `refreshGoogleSessionTokensIfNeeded`:

```ts
const inflight = new Map<string, Promise<RefreshOutcome>>();

export function getOrStartSessionTokenRefresh(
  userId: string,
  account: GoogleAccountForRefresh,
  deps: RefreshSessionDeps
): Promise<RefreshOutcome> {
  const existing = inflight.get(userId);
  if (existing) return existing;
  const pending = refreshGoogleSessionTokensIfNeeded(userId, account, deps)
    .finally(() => inflight.delete(userId));
  inflight.set(userId, pending);
  return pending;
}

export function __resetSessionTokenSingleflightCache(): void {
  if (process.env.NODE_ENV === "production") return;
  inflight.clear();
}
```

Cache slot is purged in `.finally()` so transient failures don't pin the slot — the next caller can retry. `__resetSessionTokenSingleflightCache` is NODE_ENV-gated so an accidental production call is a no-op.

### `src/lib/auth/__tests__/refresh-session-tokens-singleflight.test.ts`

7 unit tests pinning the cache contract:

- ✅ 5 concurrent same-userId calls → exactly **1** refresh + **1** DB write (acceptance criterion).
- ✅ 3 distinct userIds → 3 refreshes (different keys don't collapse).
- ✅ Slot released after success → next sequential call refreshes again.
- ✅ Slot released after transient failure → first call returns `transient-error`, second returns `refreshed`.
- ✅ All concurrent awaiters see the same outcome on rejection (5 → all `transient-error`).
- ✅ `null` refresh_token resolves with `terminal-error`; cache still purges so a follow-up valid call refreshes.
- ✅ `not-expired` short-circuit — fresh account skips refresh + DB entirely.

### `src/lib/auth/auth.ts`

Single-line swap: `refreshGoogleSessionTokensIfNeeded(...)` → `getOrStartSessionTokenRefresh(...)`. All other behaviour preserved (decrypt → fetch → re-encrypt → DB write → classify on error, terminal outcome sets `session.error = "RefreshTokenError"`).

### `.claude/plans/auth-token-refresh-singleflight-216.md`

Updated to capture the final design + the rejected alternative (a standalone module that would have duplicated main's HTTP/DB layers without encryption support).

## History — the architecture pivot

The original 2026-05-03 design extracted a self-contained `src/lib/auth/token-refresh.ts` that owned the HTTP round-trip, the DB update, and the singleflight cache in one module. That module was reviewed and approved on 2026-05-04 (PR feedback addressed in `4657bfd`).

Between then and the 2026-06-08 rebase, `main` landed a substantial auth refactor:

- **#215** — refresh tokens encrypted at rest (`encryptToken` / `decryptToken` via `crypto/token-cipher.ts`).
- **#315 / #348** — refresh errors classified `terminal` vs `transient` (`refresh-error-classifier.ts`).
- **#217** — `fetchWithRetry` wrapping the OAuth round-trip (`refresh-google-token.ts`).
- Session-callback path factored into three modules.

Replaying the original `token-refresh.ts` on current `main` would have **silently regressed two production-critical fixes** — plaintext token writes (#215) and terminal-classification of every error (#315). So the singleflight idea is preserved, but the implementation now wraps `refreshGoogleSessionTokensIfNeeded` instead of duplicating its layers.

## What does NOT land here

- **Idle-session timeout** — sliding `maxAge`, user-settings toggle, 24/7 wall-display preserve mode. Separate follow-up; #216 stays open after this merges to track that work.
- **Unifying `/api/auth/refresh-token`** with the singleflight queue. Tracked in #285.
- **Cross-process refresh dedupe** for multi-instance deployments. Tracked in #286.

## Test plan

- [x] `pnpm install --frozen-lockfile` clean
- [x] `pnpm lint:fix` — 0 errors (6 pre-existing warnings on unrelated files)
- [x] `pnpm format:fix` clean
- [x] `pnpm test:run` — **2337 / 2337 passing across 147 files**
- [x] `pnpm test:run src/lib/auth/__tests__/` — **113 / 113 passing** (includes the 7 new singleflight tests)

## Followups

- New issue / re-use #216 for the idle-session-timeout half once this lands.
- #285 and #286 (already filed) for the route-unification and cross-process dedupe.

---


src/lib/auth/auth.ts
```diff
@@ -38,58 +42,17 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
...
-            },
-            where: {
-              provider_providerAccountId: {
-                provider: "google",
-                providerAccountId: googleAccount.providerAccountId,
-              },
-            },
-          });
-
+          await getOrStartTokenRefresh(
```
BenSeymourODB: **Bug (hold-for-revision): `createDefaultTokenRefreshDeps()` is inside the `try` block — missing env vars produce a silent `RefreshTokenError` instead of a boot-time crash.**

`createDefaultTokenRefreshDeps()` throws when `GOOGLE_CLIENT_ID` or `GOOGLE_CLIENT_SECRET` are absent. Because the call is inside the `try/catch` at line 51, that `Error` is caught, logged as `TokenRefreshFailed`, and the session gets `error: "RefreshTokenError"`. The operator sees a runtime auth failure for every user rather than a clear startup or config-check failure.

The original inline code had the same issue (`process.env.GOOGLE_CLIENT_ID!` with a non-null assertion that could be `undefined`), but the new helper was specifically designed to throw on bad config, so the caller should honour that intent:

```ts
// Outside the try block (or at module level):
const deps = createDefaultTokenRefreshDeps();

// Inside the try block:
await getOrStartTokenRefresh(googleAccount, deps);
```

Alternatively, call it once at module load time and export the result, so the error surfaces at server startup rather than on the first session refresh.

---

src/lib/auth/token-refresh.ts
```diff
@@ -0,0 +1,161 @@
...
+  };
+};
+
+type PrismaLike = {
+  account: {
+    update: (input: AccountUpdateInput) => Promise<unknown>;
+  };
+};
+
+export type TokenRefreshDeps = {
```
BenSeymourODB: **Deviation from plan: `logger` was dropped from `TokenRefreshDeps`.**

The plan spec lists `logger: Logger` as an injected dependency in `TokenRefreshDeps`. It was removed from the implementation, so `refreshGoogleAccessToken` has no structured observability of its own. Every event (e.g. the Google round-trip latency, which is a natural `logger.dependency()` call) must now be emitted by the call-site (`auth.ts`), which means any other future caller of `getOrStartTokenRefresh` will silently miss telemetry.

Suggest adding `logger` back to `TokenRefreshDeps` (optional, so existing tests need no change) and emitting at least:
- `logger.dependency("GoogleOAuthRefresh", url, success, durationMs)` around the `fetch` call inside `refreshGoogleAccessToken`.
- The plan's `TokenRefreshed` event could move here too, keeping `auth.ts` thin.

This is a forward-compatibility concern rather than a correctness bug, so it could be addressed in a follow-up if Application Insights is still far off — but the plan deviation should be acknowledged.

---

src/lib/auth/token-refresh.ts
```diff
@@ -0,0 +1,161 @@
...
+export function getOrStartTokenRefresh(
+  account: GoogleTokenAccount,
+  deps: TokenRefreshDeps
+): Promise<RefreshedTokens> {
+  const key = cacheKey(account);
+  const existing = inflight.get(key);
+  if (existing) {
+    return existing;
+  }
+  const pending = refreshGoogleAccessToken(account, deps).finally(() => {
```
BenSeymourODB: **Microtask ordering: the cache slot is visible to a third caller in the window between `.finally()` clearing the map and the awaiting callers resuming.**

This is a very minor point — it does not affect correctness for the typical case — but worth noting for the record.

When the in-flight promise settles, `.finally(() => inflight.delete(key))` runs as a microtask. All awaiters that were already `await`ing the promise also resume in microtask order. Because `.finally()` is chained on `pending` before it is stored in the map, the delete fires *before* those awaiters continue. A third caller that happens to call `getOrStartTokenRefresh` synchronously from inside one of those awaiter continuations will therefore see an empty map slot and trigger a new fetch — which is the correct "cache cleared after settle" behaviour.

No code change needed; this is confirming the ordering is safe as written. Just leaving this note so future maintainers don't second-guess the chain order.

---

src/lib/auth/token-refresh.ts
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/**
+ * Singleflight wrapper around `refreshGoogleAccessToken`. Multiple
+ * concurrent callers for the same `(provider, providerAccountId)` await
+ * the same in-flight promise so only one OAuth round-trip is issued. The
+ * cache slot is purged after the promise settles (success or failure) so
+ * subsequent refreshes start fresh.
+ */
+export function getOrStartTokenRefresh(
```
BenSeymourODB: **Missing-refresh-token edge case does not pollute the cache — confirm this is tested.**

When `account.refresh_token` is `null`, `refreshGoogleAccessToken` throws synchronously (before the `async` function body returns a pending promise). `getOrStartTokenRefresh` calls `refreshGoogleAccessToken(account, deps)` and immediately gets a rejected promise back; it then attaches `.finally()` and stores that rejected+finally promise in the map.

The `.finally()` delete will fire on the next microtask tick, but between the `inflight.set(key, pending)` line and that microtask there is a window where a concurrent synchronous caller would see the (already-rejected) slot. In practice this is not a problem — the concurrent caller would simply receive the same rejection — but the plan listed "Missing refresh_token — cache is not polluted" as test case #7. The current test (`refreshGoogleAccessToken` suite, "rejects synchronously when the account has no refresh_token") only tests the inner function directly and does not call `getOrStartTokenRefresh`. Consider adding a `getOrStartTokenRefresh` variant of this test to fully satisfy acceptance criterion #7 from the plan.

---

src/lib/auth/token-refresh.ts
```diff
@@ -0,0 +1,161 @@
...
+    inflight.delete(key);
+  });
+  inflight.set(key, pending);
+  return pending;
+}
+
+/**
+ * Reset the in-flight cache. Test-only helper; production code never calls this.
+ */
+export function __resetTokenRefreshCache(): void {
```
BenSeymourODB: **Nit: `__resetTokenRefreshCache` is a public export — consider gating it.**

Exporting a double-underscore helper from the production module works, but any consumer can call it and wipe the in-flight map. A common pattern is to guard with `process.env.NODE_ENV`:

```ts
export function __resetTokenRefreshCache(): void {
  if (process.env.NODE_ENV === "production") return;
  inflight.clear();
}
```

This is a nit — the function name convention (`__`) is a reasonable signal on its own — but the guard makes the intent machine-checkable.

---

src/lib/auth/token-refresh.ts
```diff
@@ -0,0 +1,161 @@
...
+  const clientId = process.env.GOOGLE_CLIENT_ID;
+  const clientSecret = process.env.GOOGLE_CLIENT_SECRET;
+  if (!clientId || !clientSecret) {
+    throw new Error(
+      "GOOGLE_CLIENT_ID and GOOGLE_CLIENT_SECRET must be set to refresh tokens"
+    );
+  }
+  return {
+    fetch,
+    prisma: defaultPrisma as unknown as PrismaLike,
```
BenSeymourODB: **Nit: `prisma as unknown as PrismaLike` — the double cast is necessary but could be narrowed.**

The `PrismaLike` structural type is defined in this file and only exposes `account.update`. The `as unknown as PrismaLike` cast is the correct escape hatch here since Prisma's generated client type doesn't directly extend a hand-rolled structural interface. No change required, just confirming this is the right approach (not laziness), and noting that if `PrismaLike` ever grows more methods, the cast will silently pass even if the runtime client doesn't have them — keep `PrismaLike` as minimal as it is now.

---

BenSeymourODB (commented):
Overall this is solid work — the singleflight contract is correct, `.finally()` purges on both success and failure, all awaiters share the same promise reference so rejection propagates uniformly, and `response.json()` on a non-OK status is wrapped in a try/catch. CLAUDE.md compliance is good: TDD order followed, strict TS throughout, no manual memoization, no emojis added, no feature creep. The `/api/auth/refresh-token/route.ts` endpoint is confirmed to be completely separate — it takes a token from the request body, never touches the `inflight` Map, and has its own inline fetch. That scope boundary is correct and matches the plan.

Two changes are required before merge: (1) `createDefaultTokenRefreshDeps()` must be moved outside the `try` block in `auth.ts` — as written, a missing-env-var startup error is silently swallowed and every user session gets `RefreshTokenError`; (2) a `getOrStartTokenRefresh`-level test for the `null` refresh_token path is missing (plan acceptance criterion #7 is only half-covered). The `logger` omission from `TokenRefreshDeps` and the `__resetTokenRefreshCache` guard are flagged as lower-priority follow-ups. See inline comments for details.

**Verdict: hold for revision** (items 1 and 2 above).

---

src/lib/auth/auth.ts
BenSeymourODB: Fixed in 4657bfd. `createDefaultTokenRefreshDeps()` is now resolved via a module-level lazy memo (`getRefreshDeps()`) that's called *outside* the `try` block in the session callback. A missing `GOOGLE_CLIENT_ID`/`GOOGLE_CLIENT_SECRET` now propagates up through NextAuth as a real server error rather than getting silently turned into a per-user `RefreshTokenError`. The lazy memo also avoids re-reading env vars on every session call.

---

src/lib/auth/token-refresh.ts
BenSeymourODB: Addressed in 4657bfd. `logger?: LoggerLike` is now back on `TokenRefreshDeps` (optional, so existing tests didn't have to change) with a minimal structural type that only exposes `dependency(name, target, duration, success, resultCode?, type?)`. `refreshGoogleAccessToken` now emits `logger.dependency("GoogleOAuthRefresh", "https://oauth2.googleapis.com/token", durationMs, success, statusCode)` on every round-trip — including the abort/network-error path. Two new unit tests verify the wiring: ok-status (`success=true`, `resultCode=200`) and non-OK (`success=false`, `resultCode=400`).

I left the `TokenRefreshed` event in `auth.ts` for now since it's session-scoped (carries `userId`) and `token-refresh.ts` doesn't know about NextAuth's user — happy to revisit if you'd prefer it move down a layer in a follow-up.

---

src/lib/auth/token-refresh.ts
BenSeymourODB: Confirmed — no code change. The chain order (`.finally()` attached to `pending` *before* `inflight.set(key, pending)`) is what guarantees the delete fires before any awaiter resumes, so a third caller in an awaiter continuation correctly sees an empty slot and starts a fresh fetch. Leaving the note here for future maintainers as suggested.

---

src/lib/auth/token-refresh.ts
BenSeymourODB: Fixed in 4657bfd. Added `getOrStartTokenRefresh (singleflight) > does not pollute the cache when the account has no refresh_token`. The test:

1. Calls `getOrStartTokenRefresh` with `refresh_token: null` — expects the rejection (`/refresh token/i`).
2. Awaits the rejection so the `.finally()` purge has a chance to run on the next microtask.
3. Calls `getOrStartTokenRefresh` again on the same provider/providerAccountId but with a *valid* refresh_token, and asserts `fetch` was called exactly once and the new token is returned.

Plan acceptance criterion #7 ("Missing refresh_token — cache is not polluted") is now fully covered at both layers (`refreshGoogleAccessToken` direct + the singleflight wrapper).

---

src/lib/auth/token-refresh.ts
BenSeymourODB: Done in 4657bfd. `__resetTokenRefreshCache` now early-returns when `process.env.NODE_ENV === "production"`, so an accidental production call is a no-op rather than wiping the in-flight map.

```ts
export function __resetTokenRefreshCache(): void {
  if (process.env.NODE_ENV === "production") return;
  inflight.clear();
}
```

Tests run under `NODE_ENV === "test"` so the gate doesn't affect them.

---

src/lib/auth/token-refresh.ts
BenSeymourODB: Confirmed — no code change. Agreed: `PrismaLike` is the minimal hand-rolled structural surface (`account.update` only) and the `as unknown as PrismaLike` is the boundary cast for Prisma's generated client type. I'll keep it tight as new methods come up rather than letting it grow speculatively, so the cast stays a deliberate, narrow escape hatch rather than a silent compatibility shim.

---

BenSeymourODB (commented):
## In-depth review — recommendation: **close as architecturally superseded** (singleflight concept still wanted; reimplement on current `main`)

PR-review fallback pass on PR #252. Reviewed against the current PR head (`4657bfd`, unchanged since 2026-05-04) and `main` at `4b387e0`. Findings below are independent of the prior 2026-05-04 review thread — that review's two blockers were both addressed in `4657bfd` (the lazy `getRefreshDeps()` helper hoists `createDefaultTokenRefreshDeps()` outside the `try`, and the "does not pollute the cache when the account has no refresh_token" test pins the synchronous-reject path). What changed is `main`.

### Setup

- Local checkout, unshallowed, `git merge origin/main` into the PR branch:
  - **`src/lib/auth/auth.ts`** — content conflict, 5 hunks, every change site overlaps.
  - All other files in the PR (`token-refresh.ts`, the test file, the plan) merge clean.
- Merge aborted — pushing the conflict resolution would either revert the larger main-side refactor or strand half of the singleflight code.

### What landed on `main` after this PR's branch base (`8ef8443`, 30 days ago, 114 commits behind)

PR #348 (Issue #315 Part B) rewrote the same code path this PR touches into four new modules:

| File | Role |
|---|---|
| `src/lib/auth/refresh-google-token.ts` | Pure `refreshGoogleAccessToken(refreshToken, clientId, clientSecret)` helper. Throws `GoogleTokenRefreshError(status, body)` on non-OK. Uses `fetchWithRetry` for transient 5xx / 429 / network failures (#217). |
| `src/lib/auth/refresh-session-tokens.ts` | `refreshGoogleSessionTokensIfNeeded(userId, account, deps): Promise<RefreshOutcome>`. Encrypts/decrypts the refresh token (#315 Part A's `token-cipher` work), routes errors through the classifier, returns a discriminated union: `not-expired` / `refreshed` / `terminal-error` / `transient-error`. |
| `src/lib/auth/refresh-error-classifier.ts` | `classifyTokenRefreshError(err) → "terminal" | "transient"` — only `invalid_grant`, `unsupported_grant_type`, `MissingRefreshTokenError`, `RefreshTokenDecryptError` force re-auth; everything else (network, 5xx, 429) is transient and the user keeps their session. |
| `src/lib/auth/__tests__/session-callback.test.ts` | Full unit-test surface for the orchestrator. |

`src/lib/auth/auth.ts` lines 119–169 on `main` are now a 5-line delegation to that orchestrator. The inline 60-line `if (expires_at < now)` block this PR was written against **no longer exists**.

### Why a merge can't land this PR

The two architectures answer different questions:

- **This PR** treats refresh as a per-`(provider, providerAccountId)` singleflight wrapping a single OAuth round-trip + DB update. `getOrStartTokenRefresh(account, deps)` returns `RefreshedTokens`. Errors are unclassified — every failure manifests as `session.error = "RefreshTokenError"`.
- **`main`** treats refresh as an outcome-classified pipeline that also encrypts ciphertext at rest, decrypts it on read, and only flags `session.error = "RefreshTokenError"` for genuine terminal causes. Transient failures leave the session intact so the next callback retries against the IndexedDB-cache fallback.

Resolving the merge in either direction loses something:

- **Accept `main`'s session callback** ⇒ delete every line of this PR that touches `auth.ts`, lose the singleflight wiring, leaving only the `token-refresh.ts` module and tests sitting unused.
- **Accept this PR's session callback** ⇒ revert PR #348 entirely: drop refresh-token encryption (the production wire format on `main`), drop the transient/terminal split (user-visible regression where a 503 from Google forces a re-sign-in), drop `fetchWithRetry` integration, drop the `refresh-session-tokens` outcome union the rest of the codebase now branches on.

There is no \"take both\" resolution because the dependency surfaces are different: `getOrStartTokenRefresh` takes a `TokenRefreshDeps` and returns tokens; `refreshGoogleSessionTokensIfNeeded` takes a `RefreshSessionDeps` and returns an outcome. The `auth.ts` call site shapes are mutually exclusive.

### The underlying problem is still real on `main`

Spot-checked `refresh-session-tokens.ts` and `session-callback.test.ts` on `4b387e0`: **no concurrent-call de-duplication exists**. Two simultaneous `auth()` invocations on an expired token still race — both decrypt the same refresh token, both POST to `oauth2.googleapis.com/token`, both `prisma.account.update` (the second clobbering the first), and if Google rotates the refresh token only one of the two new ciphertexts wins the write. Issue #216's first acceptance criterion (\"5 concurrent expired-token requests → exactly one OAuth refresh hits the network\") is **not** satisfied by what's currently on `main`.

So the **singleflight concept this PR introduced is still wanted** — it just needs to be re-implemented as a thin wrapper around `refreshGoogleSessionTokensIfNeeded` on current `main`, rather than as a replacement for it.

### Recommended path forward

**Close this PR.** Open a small follow-up (carry the same #216 link) that adds singleflight on top of the current main:

```ts
// src/lib/auth/token-refresh-singleflight.ts
const inflight = new Map<string, Promise<RefreshOutcome>>();

export function getOrStartSessionRefresh(
  userId: string,
  account: GoogleAccountForRefresh,
  deps: RefreshSessionDeps
): Promise<RefreshOutcome> {
  const key = `${account.providerAccountId}`;
  const existing = inflight.get(key);
  if (existing) return existing;
  const pending = refreshGoogleSessionTokensIfNeeded(userId, account, deps).finally(() => {
    inflight.delete(key);
  });
  inflight.set(key, pending);
  return pending;
}
```

`auth.ts` swaps `await refreshGoogleSessionTokensIfNeeded(...)` → `await getOrStartSessionRefresh(...)`. The 8 unit tests in this PR (concurrent-collapse, different-keys-don't-collapse, same-rejection-to-all-awaiters, slot-released-on-success, slot-released-on-failure, no-pollution-on-missing-refresh-token, the two telemetry tests) port directly — only the inner `refreshGoogleAccessToken` mock changes to a `refreshGoogleSessionTokensIfNeeded` mock returning a `RefreshOutcome`. Estimate: ~60 lines of new code, ~150 lines of test, no rebase work, no architectural conflict.

The plan file `.claude/plans/auth-token-refresh-singleflight-216.md` in this PR is still ~85% accurate — only the \"Wire into the session callback\" Phase 2 section needs the substitution above. Worth carrying forward.

### Disposition note for #316 / #382 parallel

This is the same shape as PR #316 vs #304 (close-as-superseded recommended in [my 2026-06-06 review of #316](https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/316#pullrequestreview-4443864144)). The pattern — long-open feature PR overtaken by a parallel refactor on `main` — is now common enough that catching it earlier may justify a tracking issue (\"PR triage: branches whose base SHA is no longer reachable from main\"). I won't file that without your call.

### CLAUDE.md / quality gates

Not re-run on this PR because there's no merge-clean tree to run them against. The PR's own pre-merge `pnpm test:run` (1021/1021 in the PR body) and `pnpm check-types` were green at branch time; that's not the question — the question is whether the architecture is still compatible, and it isn't.

---

🤖 In-depth review by `/implement-issue` PR-review fallback path. No code changes pushed (rebase would silently regress #348's transient/terminal classification).

---

PR #254, opened 2026-05-03
fix(calendar): split-button ViewSwitcher + dropdown modal=false (#235)

### LINKED ISSUE #235: ViewSwitcher Day/Week split-button affordance + popover-pointer interception
**Discovered during:** [QA regression sweep 2026-05-02 (Phase A + B)](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/claude/qa-regression-plan-2026-05-02/docs/screenshots/qa-regression-2026-05-02/RESULTS.md) — Findings 2 / 5.

## Reproduce

1. Navigate to `/test/calendar?view=month&events=default` (or `/calendar` when signed in).
2. Click the **Week** button in the ViewSwitcher.

## Observed

- Clicking "Week" opens a popover with **Grid / Agenda** options instead of switching to week view.
- This is the #195 agenda-toggle-within-day-and-week feature, BUT the discovery model isn't obvious — there's no visible affordance signaling that Week is now a split-button. Day and Week buttons render with chevrons but the entire button surface opens the dropdown.
- After clicking, the popover doesn't release pointer cleanly — subsequent clicks on Day/Year/etc. are blocked by `<html>` element intercepting (pointer-events bug).
- Side effect: existing E2E specs and naive test interactions that click `getByRole("button", { name: "Week" })` fail or open the wrong UI.

## Severity

**Medium.** Functional but discovery-poor; breaks existing test selectors; pointer-interception causes downstream interaction failures.

## Suggested next step

Either:
- (a) Split into a primary "switch view" button + a separate caret for the agenda toggle (clearer affordance + preserves test selectors), or
- (b) Add a tooltip ("Toggle agenda mode") to the chevron and ensure the popover dismisses fully on outside click before next pointer event.

## Evidence

`test-results/qa-regression-2026-05-02-Q-a6fee-day-→-year-→-agenda-→-clock-chromium/test-failed-1.png` (visible Grid/Agenda popover open, blocking the next click).

BenSeymourODB:
## Summary

Replaces the Day/Week dropdown-as-button UI in `ViewSwitcher` with a proper split button and stops the Radix popover from leaking pointer events to sibling buttons after dismissal.

Closes #235.

### What changed

- **Split-button affordance.** Day and Week each render as a primary button + a caret:
  - Primary `view-switcher-{day,week}` is a plain `Button` that calls `setView(view)` and preserves the user's global `agendaMode`.
  - Caret `view-switcher-{day,week}-mode` is the `DropdownMenuTrigger` with `aria-haspopup="menu"` and an `aria-label` like *"Choose day display mode"*.
  - The pre-existing sub-mode label (`Day · Agenda`) stays on the primary button for sighted users.
- **Pointer-events leak.** `DropdownMenu modal={false}` on the caret so Radix does not install the body-level overlay that was eating subsequent clicks on Year/Month/Clock after dismissal.
- **E2E specs updated** to follow the new affordance — they were locking in the broken behaviour that the issue calls out:
  - `e2e/agenda-toggle.spec.ts`
  - `e2e/week-day-views.spec.ts`
  - `e2e/calendar.spec.ts`

### Tests

Component (Vitest + RTL):
- 24 specs covering primary-button click switches view without opening menu, primary preserves `agendaMode`, primary has no `aria-haspopup`, caret has `aria-haspopup="menu"`, caret has accessible label, Grid/Agenda radio state and selection, label reflects current sub-mode, Month/Year/Clock unchanged.

E2E (Playwright, `video: "on"` on the agenda spec):
- Updated round-trip Day → Day+Agenda → Week+Agenda → Week → Month.
- New: primary Day click switches view directly with no menu rendered.
- New: primary Week click preserves `agendaMode` (Day+Agenda → Week stays in agenda).
- New: Year click after Week-caret menu dismissal is not intercepted (regression for `modal={false}`).

### Verification

- `pnpm test:run` — 1584/1584 passing.
- `pnpm lint:fix` — clean (only pre-existing warnings unrelated to this PR).
- `pnpm format:fix` — clean.
- `pnpm check-types` — one pre-existing error in `src/components/calendar/__tests__/AgendaList.test.tsx` (verified present on `main` before this branch; out of scope).
- Playwright browser binaries are blocked from downloading in this sandbox; CI will run the E2E suite.

## Test plan

- [ ] CI Vitest suite green
- [ ] CI Playwright suite green (regression + new specs in `agenda-toggle.spec.ts`)
- [ ] Manual: load `/test/calendar?events=default&view=month`, click Week — lands in Week grid, no popover. Click Year — focus moves immediately, no pointer trap.
- [ ] Manual: open the Week caret, pick Agenda — lands in Week+Agenda. Click Day primary — stays in agenda mode (Day+Agenda).

https://claude.ai/code/session_01NqYRLtzdTuUifECHUkRg1R

---
_Generated by [Claude Code](https://claude.ai/code/session_01NqYRLtzdTuUifECHUkRg1R)_

### DIFF
--- a/.claude/plans/viewswitcher-split-button.md
+++ b/.claude/plans/viewswitcher-split-button.md
@@ -1,0 +1,85 @@
+# ViewSwitcher Split Button + Popover Pointer Fix — Implementation Plan
+
+> Addresses GitHub Issue #235.
+
+## Goal
+
+Replace the Day/Week single-button-as-dropdown UI in `ViewSwitcher` with a
+proper split-button UX (primary "switch view" button + caret that opens the
+Grid/Agenda menu). Eliminate the Radix-popover pointer-events leak that was
+blocking subsequent clicks after the popover closed.
+
+## Problem recap
+
+`ViewSwitcher` (post-#150) makes the entire Day/Week button a dropdown trigger.
+That:
+
+1. Hides the primary "switch view" affordance — sighted users have no signal
+   that there is a sub-menu, even though the button has a chevron.
+2. Breaks any test or naïve user interaction that clicks
+   `view-switcher-day` expecting to switch views, since clicking opens a
+   popover with Grid/Agenda options.
+3. After the popover closes, pointer-events on the body remain trapped under
+   the Radix modal overlay, so the next click on (e.g.) Year/Month is
+   intercepted.
+
+## Design — Option (a) from the issue
+
+Split each Day/Week control into two adjacent buttons sharing visual borders:
+
+```
+┌─────────────┬───┐
+│ ☐ Day · …   │ ▾ │
+└─────────────┴───┘
+   primary    caret
+```
+
+- **Primary button** (`data-testid="view-switcher-day"` / `…-week"`) — plain
+  `Button`, switches the view via `setView(view)`. Preserves the user's
+  current `agendaMode` setting (it is global state, set deliberately).
+- **Caret button** (`data-testid="view-switcher-day-mode"` / `…-week-mode"`) —
+  `DropdownMenu` trigger that opens a `Grid`/`Agenda` `RadioGroup`. Picking an
+  option both `setView(view)` and `setAgendaMode(mode === "agenda")`.
+- **`aria-label`** on the caret describes its purpose for screen readers
+  (e.g. `"Choose day display mode"`), since its visible content is just an
+  icon.
+- **`modal={false}`** on the `DropdownMenu` to stop Radix from installing the
+  body-level pointer-blocking overlay. This is the root fix for the
+  "popover doesn't release pointer cleanly" complaint.
+
+The primary button keeps the existing visible label that reflects the current
+mode (`Day` vs `Day · Agenda`) so sighted users still see the sub-mode at a
+glance.
+
+## Phases
+
+1. **Component tests + implementation.** Update
+   `src/components/calendar/__tests__/ViewSwitcher.test.tsx` to assert the
+   new behaviour (primary click switches view, caret click opens menu, primary
+   button has no `aria-haspopup`, caret has `aria-haspopup="menu"` and an
+   accessible label). Implement the split-button structure in
+   `ViewSwitcher.tsx`.
+2. **E2E spec updates.** Update existing Playwright specs that exercised the
+   broken affordance to click the caret instead of the primary button:
+   - `e2e/agenda-toggle.spec.ts`
+   - `e2e/week-day-views.spec.ts`
+   - `e2e/calendar.spec.ts`
+     Add a focused regression spec proving:
+   - Primary `view-switcher-week` click switches to week view (no menu).
+   - Primary click followed by an immediate Year click does not get
+     intercepted (regression for the pointer-events leak).
+3. **Verification + PR.** `pnpm test`, `pnpm lint:fix`, `pnpm format:fix`,
+   `pnpm check-types`. Push, open draft PR, mark ready when green.
+
+## Acceptance criteria (from issue + plan)
+
+- [ ] Day and Week render as split buttons (primary + caret).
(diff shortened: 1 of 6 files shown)


BenSeymourODB (commented):
Deep review of the split-button implementation and test rewrite. Logic is sound and the modal=false fix is correct. Four issues to address before merging, ordered by severity.

**1 — Stale selectors in two un-updated E2E specs (will fail CI)**

`e2e/calendar-transitions.spec.ts` (lines 31, 99) and `e2e/mini-calendar-sidebar-layout.spec.ts` (line 58) both still call `page.getByTestId("view-switcher-day").click()` and then immediately expect a `menuitemradio` to appear — which is the old broken behaviour. After this PR, that click switches the view instead of opening a menu, so those assertions will silently either time-out or be vacuously true. They need the same treatment as the other three E2E files: change `.getByTestId("view-switcher-day")` to `.getByTestId("view-switcher-day-mode")` at each line that is followed by a `getByRole("menuitemradio")` pick.

**2 — `aria-pressed` on the caret button is misleading**

`src/components/calendar/ViewSwitcher.tsx` line 185 sets `aria-pressed={active}` on the caret `<Button>`. The caret is a menu trigger — its pressed/expanded state is communicated through `aria-expanded` (injected automatically by Radix when the menu is open), not `aria-pressed`. Having `aria-pressed="true"` on a button that also has `aria-haspopup="menu"` is contradictory: AT will announce it as a toggle button *and* a menu button simultaneously. The fix is to remove the `aria-pressed` prop from the caret entirely. The primary button (line 168) correctly carries `aria-pressed={active}`; the caret does not need it.

**3 — Divider border colour uses raw `black/10` / `white/15` instead of the project's semantic `--border` token**

`src/components/calendar/ViewSwitcher.tsx` line 188:
```
className="rounded-l-none border-l border-l-black/10 px-1.5 dark:border-l-white/15"
```
`globals.css` maps `--border` to an oklch value for both light and dark themes (and registers it as `color-border`). The rest of the codebase uses `border-border` / `divide-border` for separator lines. Using hardcoded `black/10` will diverge visually if the theme is customised and won't read correctly in all light-mode colour-scheme variants (e.g. high-contrast). Replace with:
```diff
-className="rounded-l-none border-l border-l-black/10 px-1.5 dark:border-l-white/15"
+className="rounded-l-none border-l border-border px-1.5"
```
The `dark:` override becomes unnecessary because `--border` already remaps in `.dark`.

**4 — No component-level keyboard-navigation tests (Tab / Enter / Escape / arrow keys)**

The new unit suite covers mouse interaction comprehensively but has no assertions for keyboard flow: Tab should move focus from the primary button to the caret, Enter/Space on the caret should open the menu, arrow keys should navigate the radio items, and Escape should dismiss. With `modal={false}` Radix still handles Escape-to-dismiss and arrow-key navigation within the `DropdownMenuContent` (those behaviours are unaffected by the modal prop), but they are unverified by any test. Consider adding at minimum: a test that presses `Escape` after opening the caret menu and confirms the menu closes, and a test that presses `ArrowDown` inside the open menu and moves focus to the next radio item. This would also serve as a regression guard if a future upgrade to Radix changes keyboard handling.

---
**What looks good:**
- `modal={false}` is the correct and well-established fix for the Radix body-pointer-events leak; it does not affect Escape-to-dismiss or arrow-key navigation inside the menu.
- No existing assertions were *weakened* — the tests that previously clicked `view-switcher-day` to open a menu now correctly target `view-switcher-day-mode` for that action, and new assertions cover the primary-button contract.
- The nested `role="group"` with an `aria-label` on `SplitViewControl` is correct semantics; it wraps both halves of the split button into a logical unit without conflicting with the outer group.
- `modal={false}` does not break focus-trap behaviour because a `DropdownMenu` (unlike `Dialog`) does not use a focus trap — it returns focus to the trigger on close regardless of the modal setting.

---

PR #195, opened 2026-05-02
feat(calendar): agenda toggle within Day and Week views (#150)

BenSeymourODB:
## Summary

Closes #150.

Reshape Agenda from a peer of Day / Week / Month / Year into an `agendaMode` sub-toggle that applies inside Day and Week, matching the Windows 11 / Microsoft Teams Calendar widget UX. When agenda mode is on, the time-grid is replaced by a chronological list of events for the same date range — no empty 8 AM → 11 AM whitespace, just the events the user actually has.

## What changed

- **`TCalendarView`** no longer includes `"agenda"`; `CalendarProvider` adds `agendaMode: boolean` + `setAgendaMode`, persisted alongside `view` in localStorage. Old `view: "agenda"` payloads migrate to `view: "day", agendaMode: true` (and the migrated payload is flushed back on first render so the legacy value is removed).
- **`AgendaList`** — new reusable component extracted from the existing AgendaCalendar. Takes a date range + event list, renders a chronological, gap-collapsed list, and reads `agendaModeGroupBy` / `use24HourFormat` from CalendarContext.
- **`DayCalendar` / `WeekCalendar`** branch on `agendaMode`. When on, the hourly time-grid is swapped for `AgendaList` scoped to the current day or week. Header + prev/next navigation stay visible in both modes so users keep control of the date range.
- **`ViewSwitcher`** — flat tab row replaced by a row of buttons. Day and Week are dropdown triggers (Day ▾ / Week ▾) that surface "Grid" and "Agenda" as the two display modes — matches the Teams widget. Month / Year / Clock are plain toggle buttons. The dropdown commits both `setView()` and `setAgendaMode()` in a single user action so jumping from e.g. Month → Day+Agenda lands directly without an intermediate grid flash.
- **Animations** — the `/test/calendar` swap key composes `view + ":agenda"` when agenda mode is on for day/week, so the existing `AnimatedSwap` fade (#87) covers the grid ↔ agenda transition.
- **Sidebar** — visibility rule already keys on `view ∈ {day, week, year-or-clock}`; removing `"agenda"` from the enum + having Day/Week host agenda mode means the sidebar still shows whenever agenda mode is on.
- **Legacy URL & E2E compat** — the `/test/calendar` route still accepts `?view=agenda` and renders the search-driven `AgendaCalendar` in that case so existing search E2E specs keep working unchanged. New agenda mode is reachable via `?view=day&agendaMode=true` or via the dropdown.

## Phase commits

[...]

PR #255, opened 2026-05-03
chore(calendar): UX polish — scroll-to-7am, search match count, sidebar visibility (#214)

### LINKED ISSUE #214: chore(calendar): UX polish — scroll grid to 7am, visible search count, scope mini-cal visibility
## Context

Three small UX polish items deferred across recent calendar PRs. Grouped to keep the issue queue manageable.

### 1. Scroll Day/Week time grid to ~7am on mount (PR #176)

> Scroll-to-7am on grid mount — UX polish; tracked as a follow-up. The grid currently scrolls to top, which is fine when most events fall in business hours.

### 2. Visible match count next to AgendaCalendar search input (PR #144)

> The result count is only announced via the `sr-only` `aria-live` region... Sighted users get no on-screen feedback while filtering. Fix: render a visible count next to the search input.

### 3. Restrict MiniCalendarSidebar visibility to day/week views per design (PR #140)

> `week` view placement: the sidebar currently renders beside both `month` and `agenda` views in the production page; a future pass can decide whether to show it only in day/week views per the reference designs.

## Acceptance criteria

- [ ] Day/Week views auto-scroll the time grid to ~7am on mount (configurable working-hours start, defaulting to 7)
- [ ] AgendaCalendar shows a visible "N matches" badge next to the search input, in addition to the existing `aria-live` announcement
- [ ] Mini-calendar sidebar visibility is gated by view: visible in day/week, hidden in month/agenda/year (verify against reference designs in `docs/`)
- [ ] Component tests for each

## Dependencies

- Item 1 lands well alongside #206 (hourly time-grid) but doesn't strictly block it.

Tracking: #224.

BenSeymourODB:
Closes #214.

## Summary

Three small UX polish items deferred across recent calendar PRs, grouped to keep the issue queue manageable.

- **Phase 1 — Auto-scroll Day/Week time grid to 7am on mount** (`bc69661`/`70a5aba`)
  Working-hours events are immediately visible without a manual scroll. Uses `useLayoutEffect` to avoid a visible 00:00 → 07:00 flash. Re-runs when `agendaMode` toggles back to false (re-entering the grid view).
- **Phase 2 — Visible "N matches" badge in `AgendaCalendar`** (`e020e04`)
  Sighted users see a live match count while filtering. The existing `aria-live` `sr-only` announcement stays for screen readers. Singular form is "1 match", plural "N matches" / "0 matches".
- **Phase 3 — Restrict `MiniCalendarSidebar` visibility to day/week, non-agenda** (`b2ca72b`)
  Sidebar visibility uses an explicit allow-list: visible only when view is `day` or `week` AND `agendaMode` is off. Previously remained visible in agenda mode.
- **Bonus — `loadEventsForYear` test mock fix** (`1d1af61`)
  CI's strict tsc surfaced a pre-existing missing field on `midnight-rollover.test.tsx` after rebase onto main. One-line fix.

Plan: [`.claude/plans/calendar-ux-polish-214.md`](./.claude/plans/calendar-ux-polish-214.md).

## Test plan

- [x] `pnpm test:run` — 1750/1750 passing
- [x] `pnpm lint && pnpm check-types` — clean (warnings only, all pre-existing on main)
- [x] `npx prisma migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma --exit-code` — in sync
- [ ] Manual / Playwright (deferred to local QA): open `/calendar` on Day or Week view → grid shows 7am region first
- [ ] Manual: type in agenda search input → "N matches" badge appears next to the input
- [ ] Manual: cycle Day/Week/Month/Year/Clock + agenda toggle → sidebar visible only in day/week with agenda off

## Out of scope

- Reconciling `test/calendar`'s standalone agenda view with the production agenda-mode model (#195 follow-up).
- Wiring `WORKING_HOURS_START_HOUR` to a user-configurable preference; currently a constant.

https://claude.ai/code/session_01JcD8BYdas9esYxiu1tVG4b

### DIFF
--- a/.claude/plans/calendar-ux-polish-214.md
+++ b/.claude/plans/calendar-ux-polish-214.md
@@ -1,0 +1,81 @@
+# Calendar UX polish — issue #214
+
+Three small UX items deferred across recent calendar PRs. Each is independent and ships as its own phase / commit.
+
+## Phase 1 — Scroll Day/Week time grid to ~7am on mount
+
+**Goal.** When the user first lands on the Day or Week view, the time grid is positioned so working hours (~7am onwards) are immediately visible, without a manual scroll.
+
+**Approach.**
+
+- Add a shared `WORKING_HOURS_START_HOUR` constant (default `7`) — the hour to surface on mount.
+- `DayCalendar` (`HOUR_HEIGHT_PX = 48`): scroll target is `7 * 48 = 336px`.
+- `WeekCalendar` (`HOUR_HEIGHT_PX = 40`): scroll target is `7 * 40 = 280px`.
+- Use `useLayoutEffect` on a `useRef` attached to the scroll container. `useLayoutEffect` runs synchronously after DOM mutations but before paint, so the grid never visibly flashes at 0:00.
+- The effect runs only on mount — subsequent user scrolls are not overridden.
+- Each grid's existing `overflow-y-auto` div is the scroll container; capture it via `ref`.
+
+**Tests.**
+
+- Unit test: `getInitialScrollTop(7, 48) === 336` and `(7, 40) === 280`. Pure helper.
+- Component test (DayCalendar / WeekCalendar):
+  - Mount the component with mocked `scrollTop` setter.
+  - Assert the scroll container's `scrollTop` is set to the expected pixel value once.
+- E2E (Playwright): mount Day view, assert the 7am hour label is in the visible viewport.
+
+**Acceptance.** Day and Week views auto-scroll to 7am on mount; user scroll behaviour unaffected.
+
+## Phase 2 — Visible match count next to AgendaCalendar search input
+
+**Goal.** Sighted users see a live match count while filtering. The existing `aria-live` `sr-only` announcement stays for screen readers.
+
+**Approach.**
+
+- Render a small badge inline with the search input, only when `searchActive` (i.e. trimmed query length > 0).
+- Text: `${resultCount} matches` / `1 match` / `0 matches`. Reuse the existing `resultCount` constant.
+- Place it inside the input's right padding band (left of the clear button) or as a sibling beneath, depending on what looks cleanest. Defer to inline-on-the-right since the input already has right padding for the clear icon.
+- Use `text-muted-foreground text-xs` to match other secondary copy in the file.
+- Keep the existing `sr-only` `aria-live` region untouched.
+
+**Tests.**
+
+- Component test:
+  - With no query: badge is not rendered (assert `queryByTestId("agenda-search-match-count")` returns `null`).
+  - With query "Standup" against a fixture with 1 matching event: badge text reads `1 match`.
+  - With query "no-such-event": badge text reads `0 matches`.
+
+**Acceptance.** Visible "N matches" badge renders next to the search input while a query is active. `sr-only` aria-live region preserved.
+
+## Phase 3 — Restrict MiniCalendarSidebar visibility to day/week
+
+**Goal.** Sidebar visible only when the user is in Day or Week view _and_ not in agenda mode. Hidden in Month, Year, Clock, and any agenda-mode display.
+
+**Approach.**
+
+- `src/app/calendar/page.tsx` line 54 currently uses an exclusion list:
+  ```ts
+  const showSidebar = view !== "month" && view !== "clock" && view !== "year";
+  ```
+- Replace with an explicit allow-list and an agenda gate:
+  ```ts
+  const showSidebar = (view === "day" || view === "week") && !agendaMode;
+  ```
+- Pull `agendaMode` from `useCalendar()`.
+- The `test/calendar` page is left unchanged because it explicitly renders an `AgendaCalendar` view that no longer corresponds to the production agenda-mode model. (#195 deferred reconciling this.)
+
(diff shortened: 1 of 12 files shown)


BenSeymourODB:
## Code Review — PR #255 (calendar UX polish, issue #214)

Overall the three phases are implemented cleanly and the tests are meaningful. Three issues worth addressing before merge:

---

### 1. `WORKING_HOURS_START_HOUR` is duplicated, not shared (plan deviation)

The plan says: *"Add a **shared** `WORKING_HOURS_START_HOUR` constant (default `7`)"*.

The implementation copies the constant and the derived `INITIAL_SCROLL_TOP_PX` into both `DayCalendar.tsx` and `WeekCalendar.tsx`. If someone wants to change the default working-hours start (the plan even calls out wiring it to a preference later), they have to update two files in sync.

Suggestion: export `WORKING_HOURS_START_HOUR` from `src/lib/calendar-helpers.ts` (or a small `src/lib/calendar-constants.ts`) and import it in both calendars. `INITIAL_SCROLL_TOP_PX` stays local since the two files have different `HOUR_HEIGHT_PX` values.

---

### 2. `WeekCalendar` `useLayoutEffect([agendaMode])` resets user scroll on every agenda-off toggle

`DayCalendar` correctly uses `useLayoutEffect(fn, [])` — runs once on mount, never again. The user can freely scroll and it won't jump back.

`WeekCalendar` uses `useLayoutEffect(fn, [agendaMode])`. This fires on mount (good) **and** every time `agendaMode` changes from `true` → `false`. That means: user scrolls to midnight → user briefly opens agenda mode → user closes agenda mode → scroll position jumps back to 7am. The position the user was looking at is silently lost.

The comment says "Re-runs when agendaMode toggles back to false so re-entering the grid view also lands on working hours." — but this trades user agency for a convenience reset. Consider whether the correct behaviour is:

- **Only reset on mount** (match `DayCalendar`, dep array `[]`). The scroll container is unmounted/remounted when `agendaMode` toggles (it's inside the `agendaMode ? <AgendaList> : <div>` branch), so switching from agenda → grid already creates a fresh mount and the `[]`-dep effect fires. You'd get the 7am snap for free without the "erase user position" side-effect.

If the intent really is to reset on every re-entry, that's a deliberate product choice — but it should be covered by a test that exercises the toggle path, which is currently missing (see point 3).

---

### 3. Missing test: WeekCalendar re-scroll on agendaMode toggle

The `WeekCalendar` test suite adds two cases: initial scroll on mount, and grid absent in agenda mode. Neither test exercises the property that makes `WeekCalendar`'s effect different from `DayCalendar`'s — the re-scroll when toggling agenda off.

If the toggle-reset behaviour is intentional, add a test that:
1. Renders with `agendaMode: false` (scrolled to 280).
2. Re-renders with `agendaMode: true`.
3. Re-renders with `agendaMode: false`.
4. Asserts `scrollTop` is 280 again.

If it turns out the scroll container is unmounted/remounted on the toggle (see point 2 above), that test would pass with the `[]` dep array too, and the dependency on `agendaMode` can be removed.

---

### Minor: plan called for a `getInitialScrollTop` unit test, none provided

The plan spec says: *"Unit test: `getInitialScrollTop(7, 48) === 336` and `(7, 40) === 280`. Pure helper."* No such helper function was extracted, and no unit test was written for the calculation itself. The math is trivially correct, but the plan deviation means future maintainers reading the plan will look for a test that doesn't exist. Either extract the helper + add the unit test, or update the plan to reflect that inline constants were used instead.

---

**Summary:** Phase 2 (match count badge) and Phase 3 (sidebar visibility) are solid. Phase 1 has a behavioural question around the `WeekCalendar` dep array that should be resolved, and the shared-constant deviation from the plan should be addressed. Accessibility is fine — `aria-hidden` on the visual badge + the existing `sr-only` `aria-live` region is the correct pattern.


---

BenSeymourODB:
Thanks for the review — addressed all three actionable items + the minor plan note in `ce37e56`:

1. **Shared constant** — `WORKING_HOURS_START_HOUR` and a `getInitialScrollTop(hour, hourHeightPx)` helper now live in `src/lib/calendar-helpers.ts`. Both `DayCalendar` and `WeekCalendar` import them, so a future preference-wiring change touches one file.
2. **`WeekCalendar` dep array** — extracted `WeekGridView` sub-component (mirrors `DayGridView`). The `useLayoutEffect` lives where the ref does, with `[]` deps. The grid is already unmounted/remounted by the `agendaMode` ternary, so re-entry from agenda mode triggers the scroll-to-7am via the sub-component's mount, while normal re-renders no longer reset the user's scroll position.
3. **Toggle test** — added a regression test that drives three rerenders (grid → agenda → grid), simulates a user manual scroll, and asserts the grid lands back at 7am after re-entry. Locks in the intended behaviour.
4. **Minor (`getInitialScrollTop` helper)** — extracted the helper as the plan called for, with unit tests covering `(7, 48) === 336`, `(7, 40) === 280`, and `WORKING_HOURS_START_HOUR === 7`.

Full suite green: 1753/1753 vitest, lint clean (warnings only — all pre-existing on main), `pnpm check-types` clean, `npx prisma migrate diff` shows no drift.

---

Issue #292, opened 2026-05-04
tracking: deferred-PR-comment cleanup (round 2) — dependency map and phasing


BenSeymourODB:
## Context

Second pass of the deferred-PR-comment audit (first pass: #224, covering PRs #19 → #192). This round scanned **34 feature/fix PRs** opened or merged since #224 was filed (PRs #193 → #259, dependency-bump PRs skipped) — both **merged and currently-open** PRs were inspected.

After de-duplicating against:
- The original 23 issues #201-#223
- Other already-filed issues from this period (#230, #231, #235-#239, #260, #261)

**30 new issues** were filed: #262-#291.

Source data: 4 parallel scanner agents covered PR descriptions, general comments, and review threads.

### Calendar bug (1)
- #267 — all-day event wire format wrong on positive-UTC-offset clients (NZST etc.)

### Calendar features (10)
- #264 — fold legacy `AgendaCalendar` search into `AgendaList` day/week agenda mode
- #265 — event detail edit (PATCH) — completes #115
- #266 — hide `EventDetailModal` delete on read-only calendars (accessRole gating)
- #268 — multi-calendar selector UI in `EventCreateDialog`
- #269 — local PostgreSQL write path for calendar events
- #274 — salvage useful bits from stale RadialClock POC (PR #227) into live impl
- #277 — extend Zod validation to events.insert/patch/get + client gapi + 502 tests
- #283 — user-configurable transition speed (Off/Fast/Normal/Slow)
- #287 — reconcile `/test/calendar` agenda mode with production agenda-mode model
- #288 — user-configurable working-hours start hour

### Auth / HTTP / Tasks (5)
- #275 — drop `NEXT_PUBLIC_` prefix from server-side Google OAuth client-id
- #276 — cover retry-exhausted network-error path + tighten same-URL retry assertion
- #282 — re-auth CTA + typed `TaskApiError` across all task consumers
- #285 — route `/api/auth/refresh-token` through the singleflight queue (#216 follow-up)
- #286 — cross-process token-refresh dedupe for multi-instance deployments

### Storage / cache (2)
- #270 — explicit IndexedDB invalidate/write on event create + delete
- #290 — event-cache hardening — LRU on `QuotaExceededError`, visibility-sweep, settings-driven TTL (decomposes remaining work in #218)

### Code quality / refactor (3)
- #271 — repo-wide audit + removal of `useCallback`/`useMemo`/`React.memo` violations
[...]

PR #296, opened 2026-05-04
wave 1 cleanup: #267 all-day TZ, #287 test agenda, #289 tasks persist, #291 db docs

BDSeymour:
## Summary

Wave 1 cleanup from the round-2 deferred-PR-comment audit (tracking issue #292). One PR bundles four non-overlapping fixes:

- **fix(calendar): all-day event wire format uses YYYY-MM-DD strings (#267)** — `EventCreateDialog` now sends `YYYY-MM-DD` for `start`/`end` on all-day events instead of a JS `Date`. Server route validates the new shape via Zod and uses calendar-day arithmetic (`addOneDay`) for the exclusive end. Fixes the positive-UTC-offset bug (e.g. NZST UTC+12 picking Apr-20 local was being recorded as Apr-19). Tests cover UTC+12, UTC+0, UTC-7 plus multi-day and ISO-rejection cases.
- **chore(test): mount production agenda model on /test/calendar (#287)** — `/test/calendar?view=agenda` now renders the same `DayCalendar` + `agendaMode=true` → `AgendaList` surface as production, instead of the legacy standalone `AgendaCalendar`. `AgendaCalendar.tsx` itself is unchanged (its removal belongs to #264).
- **feat(test/tasks): persist list-picker selection to localStorage (#289)** — Live-mode list `<select>` selection survives reload via `useLocalStorage('test-tasks.live-list')`. Stale ids fall back to the placeholder without throwing. 5 new component tests.
- **docs(database): document CREATE INDEX CONCURRENTLY convention (#291)** — New section in `docs/database.md` covering when to use `CONCURRENTLY` (populated/hot tables in production), when the blocking variant is fine (greenfield/pre-launch), naming reminder, and concrete examples. References `0003_unique_account_user_provider` as a pre-launch precedent that should not be copied onto a hot table.

## Wave 1 issues NOT in this PR

- **#273** (gate `__resetUseDateNowForTests` from production) and **#284** (refactor `SimpleCalendar` to consume `useSlideDirection`) are blocked by upstream PRs **#225** and **#250** respectively — both still open. The hooks they touch don't exist on `main` yet.

## Test plan

- [x] `pnpm test` → 1644 / 1644 passing (added all-day TZ matrix, ISO rejection, list-picker persist/rehydrate/stale-fallback, EventCreateDialog string emission)
- [x] `pnpm lint:fix` → 0 errors (6 pre-existing warnings unchanged)
- [ ] Manual: create an all-day event from a non-UTC client and confirm the upstream Google Calendar entry lands on the correct local date
- [ ] Manual: load `/test/calendar?view=agenda` and confirm it renders the production agenda surface
- [ ] Manual: change list on `/test/tasks` live mode, reload, confirm selection persists; clear localStorage and reload, confirm graceful default

Closes #267, #287, #289, #291.

### DIFF
--- a/docs/database.md
+++ b/docs/database.md
@@ -173,6 +173,61 @@ prisma/
 5. **Use `prisma migrate deploy`** for production (applies only, never creates)
 6. **Don't use `prisma db push`** anymore — it bypasses the migration system
 
+## Index Creation: CONCURRENTLY vs Blocking
+
+PostgreSQL's default `CREATE INDEX` (and `CREATE UNIQUE INDEX`) takes an **AccessShareLock** on the table for the duration of the build. On a small or empty table this is instantaneous and harmless; on a large, hot production table it can block every read and write until the index is finished — minutes or longer.
+
+### When to use `CREATE INDEX CONCURRENTLY`
+
+Use the `CONCURRENTLY` keyword whenever the migration will run against a table that:
+
+- already has rows in production (any non-greenfield table), or
+- receives live traffic at the time the migration is applied.
+
+The concurrent form builds the index in the background without holding a table lock. It takes longer overall but allows normal reads and writes throughout.
+
+```sql
+-- Safe on a live table: no table lock, reads and writes continue normally.
+CREATE UNIQUE INDEX CONCURRENTLY "PointTransaction_profileId_taskId_key"
+    ON "PointTransaction"("profileId", "taskId");
+```
+
+> **Prisma note:** Prisma does not emit `CONCURRENTLY` in auto-generated migrations. Always hand-edit the migration SQL for indexes added to populated tables.
+
+### When the blocking variant is acceptable
+
+The non-concurrent form is acceptable when:
+
+- the table is **greenfield** (created in the same migration, so it has zero rows at migration time), or
+- the migration runs only in **development** or pre-launch environments where no live traffic exists, or
+- the table is provably tiny and the lock duration is negligible.
+
+```sql
+-- Fine in the same migration that creates the table (zero rows at run time).
+CREATE UNIQUE INDEX "PointTransaction_profileId_taskId_key"
+    ON "PointTransaction"("profileId", "taskId");
+```
+
+### Existing pre-launch migrations
+
+The migration `0003_unique_account_user_provider` (`CREATE UNIQUE INDEX "Account_userId_provider_key"`) and the `0001_initial` baseline both use the blocking form. This was intentional — both ran pre-launch against empty tables. **Do not copy this pattern** when adding indexes to tables that may contain data in any target environment.
+
+If a future migration introduces an index on an existing populated table, write:
+
+```sql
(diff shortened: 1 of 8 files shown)


BDSeymour:
Deferred Issues / Suggestions

  1. Minor — PR description says "Zod" but the validator is hand-rolled

  The PR body says "Server route validates the new shape via Zod", but validateCreateBody in src/app/api/calendar/events/route.ts:298-413 is a hand-rolled
  validator with manual type-narrowing. Not a code issue — just inaccurate description. (Several other routes in the codebase do use Zod, so this is worth noting
  if consistency matters to you.)

  2. Minor — DATE_ONLY_RE accepts impossible dates like "2026-13-40"

  /^\d{4}-\d{2}-\d{2}$/ only validates shape, not range. A request with startDate: "2026-13-40" would pass the regex, then addOneDay would silently roll over via
  new Date(2026, 12, 40) and emit something surprising. Low practical risk (the dialog produces the strings) but it's the kind of thing that bites later. A range
  check (month 1-12, day 1-31) or Date.UTC validity round-trip would close it.

  3. Minor — LiveListPicker stale-id behavior leaves localStorage with the stale value

  In src/app/test/tasks/page.tsx:91-95, when savedListId doesn't match any fetched list, validatedId falls back to "" for rendering — but the original stale id
  stays in localStorage (we never call setSavedListId("")). Consequence:

  - Visual state shows "— select a list —"
  - Reload again → same flash to the stale id, then same fallback

  The test at line 110-117 (falls back to default when persisted id is not in the fetched lists (stale)) only asserts the rendered value is not the stale id — it
  does not assert anything about localStorage. If you want a clean fallback, call setSavedListId("") from an effect when validation fails. If you want to keep it
  (the deleted list might reappear), worth a one-line comment noting the intent.

  4. Minor — Initial render flash in LiveListPicker

  validatedId requires lists to be populated before the saved selection can render. On first paint, lists = [], so even with a valid savedListId, the <select>
  shows the placeholder until the fetch resolves. Acceptable for a test page; would be visible noise on a production surface. No action needed here.

  5. Nit — fetch in LiveListPicker has no abort controller

  useEffect fetch with no cleanup. Test page only — won't cause real issues — but a reviewer in a more cautious area would flag it.

  Risk Assessment

  - #267 (calendar) — Wire-format change is technically a breaking contract for any external client of POST /api/calendar/events. The dialog is the only caller in
  this repo, so safe in practice; just be aware if/when an external client appears.
  - #287 (test page) — Pure test infra; no production impact.
  - #289 (test page) — Pure test infra; localStorage key is namespaced.
  - #291 (docs) — No code impact.

  Verdict

  Ready to merge after optionally addressing #1 (PR description tweak) and #3 (decide whether to also clear localStorage on stale fallback). The core fix in #267
  is the load-bearing piece, and it is well-modeled, well-tested, and the comments explain why — exactly the way a future reader will need to encounter it.

PR #300, opened 2026-05-05
chore(auth): drop NEXT_PUBLIC_ prefix from server-side OAuth client id (#275)

BenSeymourODB:
Closes #275.

## Summary

`/api/auth/refresh-token` was reading the OAuth client id from `process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID`. The `NEXT_PUBLIC_` prefix tells Next.js to inline the value into every client bundle that transitively imports the file, so the route was leaking the client id into the browser even though it only runs server-side.

This switches the route to the unprefixed `GOOGLE_CLIENT_ID` (matching `auth.ts`, the only other server-side consumer), and clarifies in the env examples and setup docs which variables are server-only vs. browser-only.

## Scope

Change is intentionally narrow:

- `src/app/api/auth/refresh-token/route.ts` — read `GOOGLE_CLIENT_ID` instead of `NEXT_PUBLIC_GOOGLE_CLIENT_ID`. Comment cites why the unprefixed name matters.
- `src/lib/google-calendar.ts` — **left as-is.** It's the legacy browser-side gapi/GIS path; `process.env.NEXT_PUBLIC_GOOGLE_*` reads there are correct because the value is consumed in the browser. (And in current main, this file is only `import type`d outside its own tests, so the runtime side is effectively dead — see verification below.)
- `.env.example`, `.env.local.example`, `docs/google-calendar-setup.md` — split the GOOGLE OAuth section into "server-only" (`GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET`) and "browser-only legacy" (`NEXT_PUBLIC_GOOGLE_CLIENT_ID` / `NEXT_PUBLIC_GOOGLE_API_KEY`), with a note that new server-side code MUST use the unprefixed pair.

## TDD

New tests in `src/app/api/auth/refresh-token/__tests__/route.test.ts` (9 cases). The two regression-prevention cases are the load-bearing ones:

- `returns 500 when GOOGLE_CLIENT_ID is missing (does NOT fall back to NEXT_PUBLIC_GOOGLE_CLIENT_ID)` — sets only `NEXT_PUBLIC_GOOGLE_CLIENT_ID="public-client-id-LEAKED"`, expects 500. Catches any future re-introduction of the public-prefixed read.
- `passes server-only GOOGLE_CLIENT_ID (not NEXT_PUBLIC_GOOGLE_CLIENT_ID) to refreshGoogleAccessToken` — sets both vars to distinct values, asserts the unprefixed value is what reaches the helper.

Both fail on the pre-fix code (verified locally) and pass on the post-fix code.

## Verification

`pnpm lint:fix && pnpm format:fix && pnpm check-types && pnpm test:run` — all green (1844/1844).

Standalone production build sanity check (`pnpm build` with distinct marker values for both env vars, then grep `.next/standalone`):

```
let r=process.env.GOOGLE_CLIENT_ID,n=process.env.GOOGLE_CLIENT_SECRET;
```

is what the compiled refresh-token chunk now contains. `NEXT_PUBLIC_GOOGLE_CLIENT_ID` does not appear in the route's chunk, and neither marker value leaks into `.next/static/*`. (Re: the public marker, it doesn't appear because every non-test import of `google-calendar.ts` is `import type` only — the runtime side that reads `NEXT_PUBLIC_GOOGLE_*` is currently dead code.)

## Test plan

- [x] Unit: 9 new route cases pass; full suite green
- [x] Type check / lint / format clean
- [x] Production build succeeds; grep `.next` confirms no server-only client id leakage and that `NEXT_PUBLIC_GOOGLE_CLIENT_ID` is no longer read in the refresh-token chunk
- [ ] Smoke: in a real deployment, sign in → wait for token expiry → confirm the refresh-token endpoint still issues a fresh access token (only `GOOGLE_CLIENT_ID` is required to be set; previously `NEXT_PUBLIC_GOOGLE_CLIENT_ID` would have sufficed)

---
_Generated by [Claude Code](https://claude.ai/code/session_01DdoDbbYMRZnvgK6oTVofVz)_

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -64,8 +64,15 @@ TOKEN_ENCRYPTION_KEY=your-32-byte-base64-encryption-key
 #    - Google Calendar API
 #    - Google Tasks API
 
+# Server-only — read by NextAuth and the /api/auth/refresh-token route.
+# The unprefixed name keeps the value out of the browser bundle.
 GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=your-client-secret
 
-# Legacy: Used by old client-side auth (deprecated, will be removed)
+# Legacy: only the deprecated browser-side gapi/GIS path in
+# `src/lib/google-calendar.ts` reads this. The `NEXT_PUBLIC_` prefix forces
+# Next.js to inline the value into the browser bundle, which is required for
+# the legacy client-side auth flow but unnecessary once that flow retires.
+# New server-side code MUST read `GOOGLE_CLIENT_ID` instead. See
+# docs/google-calendar-setup.md.
 NEXT_PUBLIC_GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
--- a/.env.example
+++ b/.env.example
@@ -70,9 +70,10 @@ GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
 GOOGLE_CLIENT_SECRET=your-client-secret
 
 # Legacy: only the deprecated browser-side gapi/GIS path in
-# `src/lib/google-calendar.ts` reads this. The `NEXT_PUBLIC_` prefix forces
+# `src/lib/google-calendar.ts` reads these. The `NEXT_PUBLIC_` prefix forces
 # Next.js to inline the value into the browser bundle, which is required for
 # the legacy client-side auth flow but unnecessary once that flow retires.
 # New server-side code MUST read `GOOGLE_CLIENT_ID` instead. See
 # docs/google-calendar-setup.md.
 NEXT_PUBLIC_GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
+NEXT_PUBLIC_GOOGLE_API_KEY=your-google-api-key
--- a/.env.local.example
+++ b/.env.local.example
@@ -32,21 +32,35 @@ NEXT_PUBLIC_APPLICATIONINSIGHTS_CONNECTION_STRING=your-connection-string-here
 # Google Calendar API Configuration
 # =============================================================================
 
-# Google Calendar API credentials for client-side authentication
-# Follow setup instructions in docs/google-calendar-setup.md
+# Google Calendar API credentials. Follow the setup steps in
+# docs/google-calendar-setup.md.
+#
+# Two separate OAuth identities are referenced below:
+#   1. Server-only:  GOOGLE_CLIENT_ID + GOOGLE_CLIENT_SECRET
+#      Read by NextAuth and the /api/auth/refresh-token route. Unprefixed
+#      so Next.js does NOT inline these into the browser bundle.
+#   2. Browser-only: NEXT_PUBLIC_GOOGLE_CLIENT_ID + NEXT_PUBLIC_GOOGLE_API_KEY
+#      Read by the legacy client-side gapi/GIS flow in
+#      `src/lib/google-calendar.ts`. The `NEXT_PUBLIC_` prefix is required
+#      because the value is consumed in the browser. New server-side code
+#      MUST NOT read these — use the unprefixed pair instead.
+#
+# In practice the same OAuth 2.0 Client ID can be used for both pairs while
+# the legacy client-side flow exists.
 
-# Get this from Google Cloud Console > APIs & Services > Credentials
-# Create OAuth 2.0 Client ID for Web application
+# Server-side OAuth Client ID (used by NextAuth + token refresh)
+GOOGLE_CLIENT_ID=your-google-client-id.apps.googleusercontent.com
+
+# Server-side OAuth Client Secret (REQUIRED for token refresh functionality)
+# IMPORTANT: Keep this secret! Never commit to source control.
+GOOGLE_CLIENT_SECRET=your-google-client-secret
+
+# Browser-side OAuth Client ID (legacy gapi/GIS flow only)
 NEXT_PUBLIC_GOOGLE_CLIENT_ID=your-google-client-id.apps.googleusercontent.com
 
-# Create API Key in Google Cloud Console > APIs & Services > Credentials
+# Browser-side API Key (legacy gapi/GIS flow only)
+# Create in Google Cloud Console > APIs & Services > Credentials
 NEXT_PUBLIC_GOOGLE_API_KEY=your-google-api-key
-
-# OAuth Client Secret (REQUIRED for token refresh functionality)
-# IMPORTANT: Keep this secret! Never commit to source control.
-# Get this from the same OAuth 2.0 Client ID credentials page
-# This is used server-side to refresh expired access tokens
-GOOGLE_CLIENT_SECRET=your-google-client-secret
 
 # =============================================================================
 # Other Configuration (Optional)
--- a/docs/google-calendar-setup.md
+++ b/docs/google-calendar-setup.md
@@ -89,12 +89,23 @@ This guide will walk you through setting up Google Calendar API credentials for 
 2. Open `.env.local` and add your credentials:
 
    ```
+   # Server-only — used by NextAuth and the token-refresh API route.
+   # Unprefixed so Next.js does NOT inline these into the browser bundle.
+   GOOGLE_CLIENT_ID=your-client-id.apps.googleusercontent.com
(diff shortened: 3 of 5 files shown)


BenSeymourODB (commented):
Overall this is a clean, well-scoped fix. The core change is correct, the documentation improvements are genuinely useful, and the test coverage for the regression risk is good. A few things to address before merging, ranging from a real correctness concern in the tests to a docs inconsistency.

---

**`google-calendar.ts` — `import type` claim verified ✓**

All four non-test imports of `google-calendar.ts` in the codebase are `import type`:
- `src/components/providers/CalendarProvider.tsx` — `import type { GoogleCalendarEvent }`
- `src/lib/calendar-transform.ts` — `import type { GoogleCalendarEvent }`
- `src/lib/event-cache.ts` — `import type { GoogleCalendarEvent }`
- `src/lib/calendar-storage.ts` — `import type { GoogleCalendarAccount }`

The PR's reasoning for leaving `google-calendar.ts` untouched is correct: the `NEXT_PUBLIC_` reads there are genuinely browser-side, and the runtime module is dead code in the current build.

**`route.ts` — `requiresReauth` contract verified ✓**

`src/lib/google-calendar.ts:refreshAccessToken` reads `errorData.requiresReauth` from the route's error body (line 333). The route returns `requiresReauth: true` on `invalid_grant` / `invalid_request`, and the test at line 162 / 183 exercises both. The contract is intact.

**Issue scope audit — complete ✓**

No `NEXT_PUBLIC_GOOGLE_CLIENT_ID` references remain outside of `google-calendar.ts` (the intentional browser-side owner) and its tests. The issue's "at least one or two other server-side helpers" turned out to be only the one route; the audit is complete.

---

BenSeymourODB (commented):
Two items need attention before merging — one correctness concern in the test env-restore pattern, one docs inconsistency. Details in inline comments below.

---

BenSeymourODB:
## Code review findings

### Finding 1 — `process.env = { ...ORIGINAL_ENV }` in `afterEach` is fragile (medium risk)

**File:** `src/app/api/auth/refresh-token/__tests__/route.test.ts`, line 69

```ts
afterEach(() => {
  process.env = { ...ORIGINAL_ENV };
});
```

Replacing the entire `process.env` reference with a new plain object is unreliable in Vitest for two reasons:

1. **Module-level captures.** `GOOGLE_CALENDAR_CONFIG` in `src/lib/google-calendar.ts` (line 47–56) reads `process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID` at module load time, not inside a function. If that module is ever imported non-type in a future test in this file, the shallow spread creates a new object but the module has already captured the old `process.env` reference. This is harmless today (all imports of `google-calendar.ts` are `import type`) but is a trap for the next person adding a test here.

2. **Vitest's own env injection.** Vitest workers inject env vars by writing to the *original* `process.env` object. After `afterEach` swaps it for a new plain object, Vitest's internal reference still points at the original, while `process.env` in the test module points at the spread copy. This means env vars injected by Vitest config (`.env.test`, `define`, `loadEnv`) may silently disappear after the first test runs.

The idiomatic safe pattern is to save/restore individual keys, or use Vitest's built-in `vi.stubEnv`:

```ts
// Option A — vi.stubEnv (preferred, Vitest-native)
beforeEach(() => {
  vi.stubEnv("GOOGLE_CLIENT_ID", "server-client-id");
  vi.stubEnv("GOOGLE_CLIENT_SECRET", "server-client-secret");
  vi.stubEnv("NEXT_PUBLIC_GOOGLE_CLIENT_ID", "public-client-id-LEAKED");
});

afterEach(() => {
  vi.unstubAllEnvs();
});
```

```ts
// Option B — manual key-level save/restore
const saved: Partial<NodeJS.ProcessEnv> = {};
const WATCHED = ["GOOGLE_CLIENT_ID", "GOOGLE_CLIENT_SECRET", "NEXT_PUBLIC_GOOGLE_CLIENT_ID"] as const;

beforeEach(() => {
  for (const k of WATCHED) saved[k] = process.env[k];
  process.env.GOOGLE_CLIENT_ID = "server-client-id";
  process.env.GOOGLE_CLIENT_SECRET = "server-client-secret";
  process.env.NEXT_PUBLIC_GOOGLE_CLIENT_ID = "public-client-id-LEAKED";
});

afterEach(() => {
  for (const k of WATCHED) {
    if (saved[k] === undefined) delete process.env[k];
    else process.env[k] = saved[k];
  }
});
```

`vi.stubEnv` is the cleanest option — it does exactly the key-level save/restore internally and integrates with `vi.restoreAllMocks()`.

---

### Finding 2 — `.env.example` is missing `NEXT_PUBLIC_GOOGLE_API_KEY` (minor inconsistency)

**File:** `.env.example`, after line 78 (end of file)

The updated `.env.example` now documents the "browser-only legacy" section with a comment that mentions `src/lib/google-calendar.ts` reads both `NEXT_PUBLIC_GOOGLE_CLIENT_ID` and `NEXT_PUBLIC_GOOGLE_API_KEY` — but only the former is actually listed. `.env.local.example` and `docs/google-calendar-setup.md` both include `NEXT_PUBLIC_GOOGLE_API_KEY`; `.env.example` alone omits it.

This was already missing before this PR, but the PR expands the Google OAuth section substantially and is the right moment to close the gap. Suggested addition at the end of the file:

```env
# Browser-side API Key (legacy gapi/GIS flow only)
# Create in Google Cloud Console > APIs & Services > Credentials
NEXT_PUBLIC_GOOGLE_API_KEY=your-api-key
```

---

### Non-issues / confirmed correct

- **`process.env` reads in `route.ts` are lazy** — both reads happen inside the `POST` function body (lines 27–28), not at module level. The `beforeEach`/`afterEach` pattern is therefore sufficient to control what the route sees during each test, even with the fragile object-swap approach. The regression tests *will* catch a revert to `NEXT_PUBLIC_GOOGLE_CLIENT_ID`.
- **Regression tests are genuinely load-bearing.** The "missing GOOGLE_CLIENT_ID" test (line 86) deletes the correct key while keeping the public one set, and asserts 500. On pre-fix code that read `NEXT_PUBLIC_GOOGLE_CLIENT_ID`, `clientId` would be non-empty and the test would get 200/mock-success instead of 500 — so it truly fails on the old code and passes on the new code. ✓
- **`requiresReauth` contract is intact.** `google-calendar.ts:refreshAccessToken` reads `errorData.requiresReauth` (line 333); the route returns it on `invalid_grant`/`invalid_request`; the tests cover both. ✓
- **`google-calendar.ts` correctly left unchanged.** All non-test imports are `import type`; the runtime browser code that reads `NEXT_PUBLIC_GOOGLE_CLIENT_ID` is legitimately browser-side. ✓
- **Docs are self-consistent** (modulo Finding 2). The server-only vs browser-only split is clearly explained across `.env.local.example` and `docs/google-calendar-setup.md`. The contributor note in the doc is a good addition.

---

BenSeymourODB:
Round-2 follow-up — both review findings addressed in e522953.

**Finding 1 — `process.env = { ...ORIGINAL_ENV }` in afterEach.** Agreed; reassigning the env object loses Node's special descriptor behaviour and would silently drift the moment something module-level reads env at import time. Switched the suite to `vi.stubEnv` / `vi.unstubAllEnvs()`:

- `beforeEach` sets the three vars via `vi.stubEnv` instead of direct mutation.
- `afterEach` calls `vi.unstubAllEnvs()` (in-place revert, no object reassignment).
- The two "missing env var" cases now use `vi.stubEnv("GOOGLE_CLIENT_ID", undefined)` / `vi.stubEnv("GOOGLE_CLIENT_SECRET", undefined)` — semantically equivalent to `delete process.env.…` for the route's `!clientId || !clientSecret` check, but consistent with the rest of the lifecycle.

All 9 route cases pass; full suite still green (1844/1844). Both regression-prevention assertions remain load-bearing — verified that swapping the route back to `NEXT_PUBLIC_GOOGLE_CLIENT_ID` would still fail them.

**Finding 2 — `.env.example` missing `NEXT_PUBLIC_GOOGLE_API_KEY`.** Added it to the legacy/browser-only block, so `.env.example`, `.env.local.example`, and `docs/google-calendar-setup.md` now describe the same set of vars. Bumped the comment from "reads this" to "reads these" since the block now covers both legacy values.

CI still pending on the new commit; will report back if anything fails.

---

PR #243, opened 2026-05-03
build(deps): bump @prisma/{client,adapter-pg} and prisma CLI to 7.6.0 (coordinated)

BenSeymourODB:
## Summary

Coordinated bump of all three Prisma packages from `^7.3.0` to `^7.6.0`. Prisma requires `@prisma/client`, `@prisma/adapter-pg`, and the `prisma` CLI to share the same major.minor version, so they must be upgraded in lockstep. Dependabot opened them as separate PRs which fail CI individually.

## Supersedes

- #107 (`@prisma/client` 7.3.0 → 7.6.0)
- #106 (`@prisma/adapter-pg` 7.3.0 → 7.6.0)

Both can be closed by a maintainer once this lands. (This PR intentionally does not close them automatically.)

## What changed

- `package.json`:
  - `dependencies.@prisma/client`: `^7.3.0` → `^7.6.0`
  - `dependencies.@prisma/adapter-pg`: `^7.3.0` → `^7.6.0`
  - `devDependencies.prisma`: `^7.3.0` → `^7.6.0`
- `pnpm-lock.yaml` regenerated. Caret ranges resolved to the latest matching `7.x` (currently `7.8.0`) for all three packages, keeping them in lockstep.

No source files were touched. The `postinstall` `prisma generate` step ran cleanly and emitted to `src/generated/prisma` (gitignored).

## Resolved versions (lockfile)

```
@prisma/adapter-pg@7.8.0
@prisma/client@7.8.0
prisma@7.8.0
```

All three on the same minor — requirement satisfied.

## Test plan

- [x] `pnpm install` succeeds; `prisma generate` runs in `postinstall` (Generated Prisma Client 7.8.0)
- [x] `pnpm lint:fix` — 0 errors (5 pre-existing warnings unrelated to Prisma)
- [x] `pnpm format:fix` — clean
- [x] `pnpm check-types` — fails with one pre-existing error in `AgendaList.test.tsx` unrelated to this change (reproduced on `origin/main` before the bump)
- [x] `pnpm test:run` — 1572/1572 tests pass across 107 files
- [ ] CI green on this branch
- [ ] Manual smoke: a route that uses `@prisma/client` (e.g. session/profile API) responds correctly

## Notes

- Branch was built fresh off current `origin/main`, not from the older `claude/fix-prisma-76-v2` branch (which is based on a stale `main` and would revert ~165 files of unrelated work).
- Draft until CI confirms green.

---
[...]

PR #301, opened 2026-05-05
feat(tasks): promote useTasks errors to TaskApiError + re-auth CTA in TaskList (#261)

### LINKED ISSUE #261: feat(tasks): promote useTasks errors to TaskApiError so re-auth CTA fires from read paths too
## Context

Surfaced in the [Sonnet review of PR #249](https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/249) for #237.

## Problem

PR #249 introduced `TaskApiError` (carries `status` + `requiresReauth`) and wired `NewTaskModal` to render a "Sign in again" CTA when the API returns `403 + requiresReauth: true`. But two other tasks consumers still throw plain `Error`:

- **`useTasks` (`src/components/tasks/use-tasks.ts:71`)** — fetch failure path: `throw new Error(\`Failed to fetch tasks from list ${list.listTitle}\`);`
- **`useTasks.updateTask` (`src/components/tasks/use-tasks.ts:123`)** — patch failure path: `throw new Error("Failed to update task");`

Both routes return the structured `403 + requiresReauth: true` response after PR #249, but the hook discards everything except the status assumption baked into the literal message. The "Sign in again" CTA only fires from `NewTaskModal` — fetch failures and task-completion failures silently swallow `requiresReauth`.

## Scope

- `useTasks` rejects with `TaskApiError(message, status, requiresReauth)` instead of plain `Error` on every non-OK response (both the per-list fetch and the `updateTask` patch).
- The same shape applies to `useTaskLists` if it has its own non-OK handler.
- Update consumers (`TaskList`, anywhere that surfaces the error) to render the same "Sign in again" CTA when `error instanceof TaskApiError && error.requiresReauth` — ideally extract a small `<TaskReauthBanner />` component so `NewTaskModal` and `TaskList` share the affordance.
- Test cases for each: `403 + requiresReauth` rejects with `TaskApiError`, status and flag preserved.

## Acceptance criteria

- [ ] `useTasks` rejects with `TaskApiError` on all non-OK responses (fetch + update)
- [ ] At least one consumer of `useTasks` surfaces the re-auth CTA when `requiresReauth` is set
- [ ] Existing tests still pass; new tests pin the typed-error contract
- [ ] No behaviour change for the success path

## Dependencies

PR #249 must land first (introduces `TaskApiError`).

Tracking: #224.

BenSeymourODB:
Closes #261.

## Summary

PR #249 introduced `TaskApiError` (carries `status` + `requiresReauth`) and wired `NewTaskModal` to render a "Sign in again" CTA on the structured `403 + requiresReauth: true` payload from `/api/tasks*`. The follow-up gap: every other consumer of those routes — `useTasks` (per-list fetch) and `useTasks.updateTask` (PATCH) — still threw plain `Error` and discarded the structured payload. So a stale Google grant only surfaced a recovery affordance from the New-Task flow; loading the list itself or completing a task silently swallowed `requiresReauth`.

This PR extends the typed-error contract to those paths and surfaces the same CTA from `TaskList`.

### Phase 1 — extract `TaskApiError` to a shared module ([commit 43f152a](https://github.com/BenSeymourODB/next-digital-wall-calendar/commit/43f152a))

- New `src/components/tasks/task-api-error.ts` owns the `TaskApiError` class and a `parseTaskApiError(response, fallbackMessage)` async helper that DRYs the JSON-body extraction (status, error message, `requiresReauth`) every consumer needs.
- `use-create-task.ts` now imports from the new module and re-exports `TaskApiError` for back-compat — existing imports in `NewTaskModal` and tests stay green.
- Drops the unused `TaskApiError` *interface* from `types.ts` (zero callers; name collided with the class).
- New unit suite (`task-api-error.test.ts`) pins the class shape + parser fallback behavior for malformed JSON, missing fields, and non-boolean `requiresReauth` values.

### Phase 2 — promote `useTasks` errors ([commit d71c4ed](https://github.com/BenSeymourODB/next-digital-wall-calendar/commit/d71c4ed))

- Replace `throw new Error("Failed to fetch tasks from list …")` in the per-list fetch and `throw new Error("Failed to update task")` in `updateTask` with `throw await parseTaskApiError(response, fallback)` so both paths propagate the structured `{ status, requiresReauth }` contract.
- `useTasks.error` preserves `TaskApiError` because the existing `instanceof Error` guard in the catch is satisfied by subclasses.
- New tests pin the typed-error contract for both fetch and update paths on `403 + requiresReauth` and on generic `500`; network errors stay as plain `Error` (no `TaskApiError` wrapping).

### Phase 3 — re-auth CTA in `TaskList` via shared `<ReauthCta />` ([commit 9486347](https://github.com/BenSeymourODB/next-digital-wall-calendar/commit/9486347))

- New `src/components/tasks/reauth-cta.tsx` owns the "Sign in again" outline button + `signIn("google", { callbackUrl })` handler. `callbackUrl` is `pathname + search` so the OAuth round-trip lands the user back on their current view (e.g. `/calendar?date=…&view=week`). NextAuth validates `callbackUrl` against `NEXTAUTH_URL`, so there is no open-redirect risk.
- `TaskList` renders `<ReauthCta />` alongside "Try again" when the hook reports `error instanceof TaskApiError && error.requiresReauth`.
- `NewTaskModal` swaps its inline button for `<ReauthCta />` so both sites stay in sync — existing modal tests still pass unchanged.
- New `reauth-cta.test.tsx` and three TaskList cases pin the renders/no-renders matrix and the `signIn` call signature.

## Out of scope (deferred)

- `profile-scoped-task-list.tsx` is a thin wrapper that delegates to `TaskList`, so it inherits the CTA without changes.
- Playwright E2E for the CTA — better paired with the shared auth fixture (#278) and the broader push in #282.

## Test plan

- [x] `pnpm test:run` — 1854/1854 pass (4 task-api-error + 4 useTasks + 4 ReauthCta + 3 TaskList CTA = 15 new)
- [x] `pnpm lint && pnpm check-types` clean (only pre-existing warnings unrelated to this change)
- [ ] CI green

https://claude.ai/code/session_012Nmo9L6KA23YCdq1f2EQP4

---

### DIFF
--- a/src/components/tasks/__tests__/use-tasks.test.ts
+++ b/src/components/tasks/__tests__/use-tasks.test.ts
@@ -404,6 +404,34 @@ describe("useTasks", () => {
       expect(result.current.tasks).toEqual([]);
     });
 
+    it("rejects with TaskApiError carrying status + requiresReauth on 401 session expiry (#261)", async () => {
+      // 401 + requiresReauth: true is the most common real-world trigger —
+      // see /api/tasks routes when the session is expired or
+      // RefreshTokenError fires.
+      mockFetch.mockResolvedValue({
+        ok: false,
+        status: 401,
+        json: () =>
+          Promise.resolve({
+            error: "Session expired. Please sign in again.",
+            requiresReauth: true,
+          }),
+      });
+
+      const config = createMockConfig();
+      const { result } = renderHook(() => useTasks(config));
+
+      await waitFor(() => {
+        expect(result.current.loading).toBe(false);
+      });
+
+      const captured = result.current.error;
+      expect(captured).toBeInstanceOf(TaskApiError);
+      expect((captured as TaskApiError).status).toBe(401);
+      expect((captured as TaskApiError).requiresReauth).toBe(true);
+      expect(captured?.message).toMatch(/sign in again/i);
+    });
+
     it("rejects with TaskApiError carrying status + requiresReauth on 403 (#261)", async () => {
       mockFetch.mockResolvedValue({
         ok: false,
--- a/src/components/tasks/__tests__/use-tasks.test.ts
+++ b/src/components/tasks/__tests__/use-tasks.test.ts
@@ -3,6 +3,7 @@
  */
 import { act, renderHook, waitFor } from "@testing-library/react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { TaskApiError } from "../task-api-error";
 import type { GoogleTask, TaskListConfig } from "../types";
 import { useTasks } from "../use-tasks";
 
@@ -403,6 +404,65 @@ describe("useTasks", () => {
       expect(result.current.tasks).toEqual([]);
     });
 
+    it("rejects with TaskApiError carrying status + requiresReauth on 403 (#261)", async () => {
+      mockFetch.mockResolvedValue({
+        ok: false,
+        status: 403,
+        json: () =>
+          Promise.resolve({
+            error: "Re-authentication required: Google Tasks scope missing.",
+            requiresReauth: true,
+          }),
+      });
+
+      const config = createMockConfig();
+      const { result } = renderHook(() => useTasks(config));
+
+      await waitFor(() => {
+        expect(result.current.loading).toBe(false);
+      });
+
+      const captured = result.current.error;
+      expect(captured).toBeInstanceOf(TaskApiError);
+      expect((captured as TaskApiError).status).toBe(403);
+      expect((captured as TaskApiError).requiresReauth).toBe(true);
+      expect(captured?.message).toMatch(/google tasks/i);
+    });
+
+    it("rejects with TaskApiError on generic 500 (no requiresReauth flag)", async () => {
+      mockFetch.mockResolvedValue({
+        ok: false,
+        status: 500,
+        json: () => Promise.resolve({ error: "Server error" }),
+      });
+
+      const config = createMockConfig();
+      const { result } = renderHook(() => useTasks(config));
+
+      await waitFor(() => {
+        expect(result.current.loading).toBe(false);
+      });
+
+      const captured = result.current.error;
+      expect(captured).toBeInstanceOf(TaskApiError);
+      expect((captured as TaskApiError).status).toBe(500);
+      expect((captured as TaskApiError).requiresReauth).toBe(false);
+    });
+
+    it("keeps network errors as plain Error (not TaskApiError)", async () => {
+      mockFetch.mockRejectedValue(new TypeError("Network error"));
+
+      const config = createMockConfig();
+      const { result } = renderHook(() => useTasks(config));
+
+      await waitFor(() => {
+        expect(result.current.loading).toBe(false);
+      });
+
+      expect(result.current.error).toBeInstanceOf(Error);
+      expect(result.current.error).not.toBeInstanceOf(TaskApiError);
+    });
+
     it("fails all tasks when one list fetch fails (partial failure)", async () => {
       mockFetch
         .mockResolvedValueOnce({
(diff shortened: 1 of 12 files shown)


src/components/tasks/use-tasks.ts
```diff
@@ -145,7 +149,7 @@ export function useTasks(config: TaskListConfig | null): UseTasksReturn {
       });
 
       if (!response.ok) {
-        throw new Error("Failed to update task");
+        throw await parseTaskApiError(response, "Failed to update task");
```
BenSeymourODB, line 152: **Correctness gap: `updateTask` 403 never surfaces the re-auth CTA in `TaskList`**

`updateTask` throws the `TaskApiError` directly but never calls `setError()`, so `useTasks.error` stays `null` after a failed task toggle. `TaskList.handleTaskToggle` also has no `try/catch`, meaning the rejection propagates as an unhandled promise and the error state (and therefore the `<ReauthCta />`) never renders.

The fetch path is fixed correctly — when `fetchTasks` throws, the `catch` at line 130 catches it, calls `setError`, and preserves the `TaskApiError` instance (since it `instanceof Error`). But the PATCH path skips that route entirely.

The PR description says both paths propagate the structured `{ status, requiresReauth }` contract to the UI — they propagate the *thrown value*, but the CTA only appears when `useTasks.error` is set. A user who toggles a task while their grant is stale will see an unhandled rejection in the console, not the "Sign in again" button.

Minimal fix: wrap the PATCH throw so `setError` is also called, or catch in `handleTaskToggle` and route the error into a component-local error state. The latter is probably cleaner since the task-toggle error is scoped differently from the initial fetch error.

This is the most impactful issue in the PR — worth resolving before merge given the stated goal of issue #261.

---

src/components/tasks/__tests__/use-tasks.test.ts
```diff
@@ -403,6 +404,65 @@ describe("useTasks", () => {
       expect(result.current.tasks).toEqual([]);
     });
 
+    it("rejects with TaskApiError carrying status + requiresReauth on 403 (#261)", async () => {
```
BenSeymourODB, line 435: **Missing test: `401 + requiresReauth: true` on the fetch path**

The new fetch-error tests only exercise `403 + requiresReauth: true` and `500 + no flag`. But the real routes also return `{ requiresReauth: true }` on `401` responses (expired session, `RefreshTokenError` — see `src/app/api/tasks/route.ts` lines 47–50 and 145–148). A 401 with `requiresReauth: true` is arguably the most common real-world trigger for this CTA.

`task-api-error.test.ts` does cover `401` but with `requiresReauth` *absent* from the body (the "defaults to false" case). There's no coverage of `401 + requiresReauth: true` at the `useTasks` level.

Suggest adding a case alongside this one:

```ts
it("rejects with TaskApiError + requiresReauth on 401 session expiry (#261)", async () => {
  mockFetch.mockResolvedValue({
    ok: false,
    status: 401,
    json: () => Promise.resolve({
      error: "Session expired. Please sign in again.",
      requiresReauth: true,
    }),
  });
  // ... same assertions as the 403 case
});
```

---

src/components/tasks/index.ts
```diff
@@ -22,6 +22,11 @@ export type { TaskListSettingsProps } from "./task-list-settings";
 export { useTasks } from "./use-tasks";
 export type { UseTasksReturn } from "./use-tasks";
 
+export { TaskApiError, parseTaskApiError } from "./task-api-error";
```
BenSeymourODB: **`parseTaskApiError` probably shouldn't be part of the public barrel**

`parseTaskApiError` is an internal fetch-response parser used exclusively inside `use-tasks.ts` and `use-create-task.ts`. Exporting it from the barrel (`src/components/tasks/index.ts`) promotes it to public API, which risks callers outside the tasks module coupling to it or it drifting out of sync with the route contract.

`TaskApiError` (the class) makes sense as a public export — consumers need it for `instanceof` checks. But the parser is an implementation detail. Consider dropping the `parseTaskApiError` export from the barrel (the two internal call sites import directly from `./task-api-error` already, so no callers need the barrel path).

Minor — won't block, but worth tightening now before other code takes a dependency on it.

---

BenSeymourODB (commented):
Overall the approach is sound: the three-phase extraction is clean, the `parseTaskApiError` helper eliminates the duplicated JSON-parsing boilerplate correctly, the `<ReauthCta />` extraction is the right call, and the new test suite is thorough. No `useMemo`/`useCallback`/`React.memo` in new code, no `any`, standard Tailwind colors only — CLAUDE.md compliance is clean.

Three issues found:

1. **Correctness (must fix before merge)** — `updateTask`'s 403 never surfaces the re-auth CTA. See the inline comment on `use-tasks.ts:152`. The PR's stated goal is to fix *both* the fetch and update paths; the update path throws the right error but never sets `useTasks.error`, so `TaskList` never renders `<ReauthCta />` for a failed task toggle.

2. **Test gap (should fix)** — No coverage of `401 + requiresReauth: true` in the `useTasks` fetch tests, even though both task routes return that combination for expired sessions (the most common real-world trigger). See the inline comment on `use-tasks.test.ts`.

3. **Minor (take it or leave it)** — `parseTaskApiError` is exported from the public barrel but is an implementation detail. See the inline comment on `index.ts:25`.

---

src/components/tasks/use-tasks.ts
BenSeymourODB: Good catch — fixed in 78d5a27 by going with your option (b): catch in `handleTaskToggle` and route into a new component-local `toggleError` state, then render a scoped error banner above the (still-rendered) task list with `<ReauthCta />` when `toggleError instanceof TaskApiError && toggleError.requiresReauth`. Toggle errors are per-PATCH so they shouldn't wipe the list — option (a) (calling `setError` from inside `updateTask`) would have done exactly that, since the existing fetch-error block hides the list entirely.

Three new tests pin the new behaviour: CTA renders + click triggers `signIn`, no CTA on a generic error, and the banner clears on the next successful toggle.

---

src/components/tasks/__tests__/use-tasks.test.ts
BenSeymourODB: Added in 78d5a27 — new `useTasks` test "rejects with TaskApiError carrying status + requiresReauth on 401 session expiry (#261)" right alongside the 403 one. Asserts the same `TaskApiError`/`status: 401`/`requiresReauth: true` propagation plus that the route's "Sign in again" message is preserved.

---

src/components/tasks/index.ts
BenSeymourODB: Tightened in 78d5a27 — barrel now exports only `TaskApiError`. Left a one-line comment in `index.ts` next to the export so the next person can see why `parseTaskApiError` is intentionally absent (it's coupled to the route contract; consumers outside the tasks module shouldn't reach for it). Both internal call sites already import directly from `./task-api-error`, so no consumer code moved.

---

BenSeymourODB:
CI is red on the "Validate Prisma migrations are up to date" step, but the failure is the pre-existing infrastructure issue tracked by #263 — **not** caused by this PR (which doesn't touch Prisma).

**Diagnosis.** The workflow runs `npx prisma migrate diff --from-migrations … --to-schema … --exit-code`. `npx prisma` fetches the latest Prisma rather than the project's pinned version. Latest is 7.6.0 (also being bumped to in #106 / #107 / #243), and 7.6.0 now requires a `--shadow-database-url` for `migrate diff --from-migrations`. CI doesn't have a shadow DB provisioned, so the command exits 1 — but the workflow's `||` branch unconditionally prints "Prisma migrations are out of date", which is misleading.

**Local reproduction** (clean checkout of this branch, no schema changes):

```
$ pnpm exec prisma migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma --exit-code
EXIT:0                                       # 7.3.0 (pinned) — passes

$ npx --yes prisma@7.6.0 migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma --exit-code
Error: You must pass the `--shadow-database-url` flag …
EXIT:1
```

**Fix belongs in #263.** The two clean fixes are (a) change `npx prisma` → `pnpm exec prisma` in the workflow so CI uses the pinned 7.3.0, or (b) provision the shadow DB as #263 already proposes. Either way, that's a CI-infra change separate from the user-visible re-auth-CTA work in this PR.

This PR is otherwise green:

- All 1858 unit/component tests pass locally (`pnpm test:run`)
- `pnpm lint && pnpm check-types` clean (only pre-existing warnings unrelated to this change)
- Three review comments addressed in 78d5a27 with second-round replies on every thread

---

PR #302, opened 2026-05-05
test(http): cover retry-exhausted network path + tighten same-URL retry assertion (#276)

BenSeymourODB:
Closes #276.

## Summary

Two deferred test improvements from PR #232. Pure test coverage / assertion tightening — no production code changes.

### 1. `refresh-google-token.test.ts` — retry-exhausted network-error path

`refreshGoogleAccessToken` calls `fetchWithRetry`, which retries `TypeError("fetch failed")` (the shape `fetch` throws for DNS / TCP / offline failures) up to its default budget of 3 attempts before giving up. The outer `catch` branch of that exhaustion was not exercised in the helper's own test suite — only the eventual-success retry path was covered.

New test:

- Pre-arms `mockFetch.mockRejectedValue(new TypeError("fetch failed"))` so every attempt fails the same way
- Uses `vi.useFakeTimers()` + `vi.runAllTimersAsync()` to skip the retry sleeps deterministically
- Asserts the rejection propagates as the original `TypeError` reference
- Asserts `mockFetch` was called exactly 3 times (`DEFAULT_MAX_ATTEMPTS`)

If a future refactor short-circuits the retry loop, the call-count assertion fails.

### 2. `route.test.ts` (calendar events) — same-URL retry identity

The existing 503→200 retry test asserted `toHaveBeenCalledTimes(2)` but didn't pin that the second call is a *retry* of the same URL — it would also pass if retry were disabled and an unrelated second calendar fetch happened to succeed.

Tightened by adding:

```ts
expect(mockFetch.mock.calls[0][0]).toBe(mockFetch.mock.calls[1][0]);
```

`fetchWithRetry` passes its `input` straight to `fetch` for string URLs, so referential equality is the right assertion here.

### 3. Drive-by — shared `jsonResponse` helper

Extracted a `jsonResponse(body, init)` helper at the top of `route.test.ts` that always sets `headers: new Headers()`. Mirrors the helper that already exists in `refresh-google-token.test.ts`, and makes a future 503→503→200 exhaustion test a one-liner without forcing every fixture to remember the `Headers()` boilerplate that `fetchWithRetry` needs to read `Retry-After`.

## Test plan

- [x] `pnpm test` — 1836 tests pass (+1 new in `refresh-google-token.test.ts`, retighten on existing 503→200 test in `route.test.ts`)
- [x] New test fails if the underlying retry loop is short-circuited (verified by mentally walking the path; `mockFetch` is called once if retry is disabled, breaking the `toHaveBeenCalledTimes(3)` assertion)

## Notes

- No production code touched. The `safeUrl` helper in `src/lib/http/retry.ts` continues to do its job; the URL identity assertion uses the raw input that `fetchWithRetry` forwards to `fetch`.
- Both tests use `vi.useFakeTimers()` so the retry sleeps are simulated, not real-time.

https://claude.ai/code/session_01WfqsBiVjPwyzBfsy5tesBn

---
_Generated by [Claude Code](https://claude.ai/code/session_01WfqsBiVjPwyzBfsy5tesBn)_

### DIFF
--- a/src/app/api/calendar/events/__tests__/route.test.ts
+++ b/src/app/api/calendar/events/__tests__/route.test.ts
@@ -11,6 +11,7 @@ import {
 import {
   type ApiErrorResponse,
   createMockRequest,
+  jsonResponse,
   parseResponse,
 } from "@/lib/test-utils/api-test-helpers";
 import { beforeEach, describe, expect, it, vi } from "vitest";
@@ -41,27 +42,6 @@ vi.mock("@/lib/logger", () => ({
 // Mock global fetch
 const mockFetch = vi.fn();
 global.fetch = mockFetch;
-
-/**
- * Build a `Response`-shaped fixture for `mockFetch`. Always sets
- * `headers: new Headers()` so `fetchWithRetry` can read `Retry-After`
- * without exploding — keeps a future 503 → 503 → 200 exhaustion test a
- * one-liner. Mirrors the helper in
- * `src/lib/auth/__tests__/refresh-google-token.test.ts` (#276).
- */
-function jsonResponse(
-  body: unknown,
-  init: { ok?: boolean; status?: number } = {}
-): Response {
-  const status = init.status ?? 200;
-  const ok = init.ok ?? (status >= 200 && status < 300);
-  return {
-    ok,
-    status,
-    headers: new Headers(),
-    json: () => Promise.resolve(body),
-  } as unknown as Response;
-}
 
 describe("/api/calendar/events", () => {
   beforeEach(() => {
--- a/src/app/api/calendar/events/__tests__/route.test.ts
+++ b/src/app/api/calendar/events/__tests__/route.test.ts
@@ -42,6 +42,27 @@ vi.mock("@/lib/logger", () => ({
 const mockFetch = vi.fn();
 global.fetch = mockFetch;
 
+/**
+ * Build a `Response`-shaped fixture for `mockFetch`. Always sets
+ * `headers: new Headers()` so `fetchWithRetry` can read `Retry-After`
+ * without exploding — keeps a future 503 → 503 → 200 exhaustion test a
+ * one-liner. Mirrors the helper in
+ * `src/lib/auth/__tests__/refresh-google-token.test.ts` (#276).
+ */
+function jsonResponse(
+  body: unknown,
+  init: { ok?: boolean; status?: number } = {}
+): Response {
+  const status = init.status ?? 200;
+  const ok = init.ok ?? (status >= 200 && status < 300);
+  return {
+    ok,
+    status,
+    headers: new Headers(),
+    json: () => Promise.resolve(body),
+  } as unknown as Response;
+}
+
 describe("/api/calendar/events", () => {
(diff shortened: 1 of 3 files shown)


BenSeymourODB (commented):
Reviewed the full diff against issue #276 and both implementation files (`src/lib/http/retry.ts`, `src/lib/auth/refresh-google-token.ts`). Here's what I verified:

**Correctness**
- New `TypeError`-exhaustion test: `mockFetch.mockRejectedValue(networkError)` makes all 3 attempts throw; `withRetry` re-throws the same error object on `attempt >= maxAttempts` (line 221 of `retry.ts`), so `expect(result).toBe(networkError)` is a strict identity check on the un-wrapped error. If `fetchWithRetry` were short-circuited the call-count assertion would fail exactly as claimed. ✓
- `fetchWithRetry` passes non-`Request` inputs through unchanged (`perAttemptInput = input` for strings), so the URL reference is literally the same object in both calls — `toBe` on `calls[0][0]` vs `calls[1][0]` correctly distinguishes a same-URL retry from an independent second fetch. ✓
- `headers: new Headers()` is now present on both fixtures (via `jsonResponse`), closing the latent crash if `fetchWithRetry` reads `Retry-After` on the 200 response. ✓

**Timer handling**
- `vi.useFakeTimers()` / `vi.useRealTimers()` follow the established `try/finally` pattern already used in the 503→200 test above. `vi.clearAllMocks()` in `beforeEach` clears the persistent `mockRejectedValue` default. No timer or mock state leaks between tests. ✓

**Acceptance criteria (all three met)**
- New retry-exhausted network-error test present and would fail on short-circuit. ✓
- Existing retry test now asserts same-URL identity. ✓
- Zero production-code changes. ✓

**One minor note** — `jsonResponse` is now duplicated verbatim across `src/lib/auth/__tests__/refresh-google-token.test.ts` and `src/app/api/calendar/events/__tests__/route.test.ts`. The issue listed this as an optional drive-by and the comment acknowledges it mirrors the other file. Consider moving it to `src/lib/test-utils/api-test-helpers.ts` (or a new `src/lib/test-utils/http-fixtures.ts`) in a follow-up so future tests don't create a third copy.

Otherwise the PR is clean — no `any` types, no manual memoization, tests are tightly scoped.

---

BenSeymourODB:
Thanks for the review. Addressed the follow-up note in `7a98e6e`:

- Hoisted `jsonResponse(body, init)` to `src/lib/test-utils/api-test-helpers.ts` (alongside `createMockRequest` / `parseResponse` / `createMockHeaders`).
- Both `refresh-google-token.test.ts` and the events `route.test.ts` now import the shared helper instead of carrying their own copies.
- Net: −9 lines across the two duplicate copies, +1 import each, behaviour unchanged.

On the failed initial CI run: it failed on the "Validate Prisma migrations are up to date" step, but my branch doesn't touch `prisma/` and locally `npx prisma migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma --exit-code` exits 0 with no diff. Treating it as a transient runner flake — the new push has triggered a fresh CI run which I'll monitor.

---

PR #304, opened 2026-05-05
feat(calendar): hide EventDetailModal delete on read-only calendars (#266)

### LINKED ISSUE #266: feat(calendar): hide EventDetailModal delete on read-only calendars (accessRole gating)
## Context

Surfaced in PR #197 review:

> Per-calendar `accessRole` gating (read-only calendars hide the delete button) — Google still enforces it server-side, so the worst case today is a clear 403 toast.

## Problem

`EventDetailModal` shows the delete button regardless of the source calendar's `accessRole`. Users on shared/read-only calendars can click delete and only learn it's disallowed when Google returns 403. Better to surface the constraint in the UI.

## Scope

- Plumb `accessRole` from the calendar list through `IEvent` (or via a per-calendar lookup)
- Hide (or disable with tooltip) the delete button in `EventDetailModal` when `accessRole` is `reader` or `freeBusyReader`
- Same treatment for the future edit (PATCH) button (depends on the edit issue)

## Acceptance criteria

- [ ] Delete button hidden for events on read-only calendars
- [ ] Existing 403 toast retained as backstop for race conditions (perm changes mid-session)
- [ ] Component test asserts both `owner`/`writer` and `reader` cases

BenSeymourODB:
Closes #266.

Hides the delete button on `EventDetailModal` for events that live on a read-only Google calendar (`accessRole === "reader"` / `"freeBusyReader"`) so users see the constraint up-front instead of via Google's 403 toast. The 403 path stays as a server-side backstop for races where the user's role changes mid-session.

## Approach

The gating decision lives **inside `EventDetailModal`**, not at each consumer. This is a single chokepoint that the future edit (PATCH) button — landing as part of #265 — can reuse without re-touching `SimpleCalendar` / `AgendaCalendar` / `AgendaList`.

## Phases

1. **Plumbing** (575dce3) — `/api/calendar/calendars` returns `accessRole`, and `CalendarProvider` exposes `getAccessRole(calendarId)`.
2. **Modal gating + consumer wiring** (1251a74) — `EventDetailModal` accepts `accessRole`, hides the delete UI when read-only, and `SimpleCalendar` / `AgendaCalendar` look up the role for the open event.

## Test plan

- [x] `pnpm test:run` — 1846 passing (8 new)
- [x] `EventDetailModal` cases: `owner` / `writer` shows delete; `reader` / `freeBusyReader` hides it; `undefined` treated as writable; AlertDialog markup absent on read-only
- [x] `SimpleCalendar` integration test: clicking a read-only event opens the modal without delete; clicking an owned event shows delete
- [ ] Manual smoke on production `/calendar` once a real shared-read-only calendar is available

## Notes / deferred

- `AgendaList` doesn't currently render delete (no `onDelete` is passed), so it didn't need wiring.
- A persistent "this calendar is read-only" badge inside the modal is **not** in this PR — the issue asks for delete-button hiding only.
- Edit button gating waits for #265.

https://claude.ai/code/session_01Y1tUwvzfVbYFXoXJ8RnyAP


BenSeymourODB (commented):
## PR #304 review — accessRole read-only delete gating

Overall the implementation is clean and the TDD coverage is solid. Found 5 actionable issues:

---

**src/components/providers/CalendarProvider.tsx:317 — `useCallback` violates CLAUDE.md React Compiler rule**

```ts
const getAccessRole = useCallback(
  (calendarId: string): TCalendarAccessRole | undefined =>
    accessRoles[calendarId],
  [accessRoles],
);
```

CLAUDE.md explicitly prohibits manual memoization (`useMemo`, `useCallback`, `React.memo`) — React Compiler handles it automatically. The inline comment even acknowledges this ("a fresh closure is fine here under React Compiler") yet the `useCallback` is still present. All other `useCallback` calls already in this file are pre-existing tech debt; adding a new one sets a bad precedent. Remove the wrapper and let the compiler handle stability. This is the only new `useCallback` introduced by this PR.

---

**src/components/providers/CalendarProvider.tsx:603-607 — stale cache means `accessRoles` is never refreshed after first load**

```ts
let calIds = calendarIds;
if (calIds.length === 0) {
  calIds = await fetchCalendarList();   // only called on first refresh
  setCalendarIds(calIds);
}
```

`fetchCalendarList` (which also calls `setAccessRoles`) is guarded by `calIds.length === 0`, so if a user's permission on a calendar changes mid-session (e.g. an owner revokes write access), `getAccessRole` will keep serving the stale role for the lifetime of the session. The PR plan acknowledges this and cites "Google's server-side 403 as backstop", so the risk is consciously accepted — but there is no comment at the call-site explaining the intentional staleness, making it a maintenance trap. Consider adding a brief note like `// accessRoles is populated here; it is intentionally not re-fetched mid-session (#266)`.

---

**src/components/calendar/AgendaList.tsx:276-281 — `accessRole` not wired; inconsistent with SimpleCalendar/AgendaCalendar when `onDelete` is later added**

`AgendaList` renders `EventDetailModal` without `accessRole` (or `onDelete`). Today that's harmless because `canDelete` requires `onDelete` to be truthy, so the button is already hidden. However the plan says "if the prop wiring gets touched as part of this change anyway, thread the role through for symmetry" and notes this is the only consumer _not_ wired up. Issue #265 (edit button) will hit the same chokepoint via `isReadOnly`; when a future author adds `onDelete` to `AgendaList` they will silently ship the delete button un-gated for read-only calendars if this prop isn't added at the same time. Recommend adding `accessRole` now (it's a one-liner) or adding a `// TODO(#266): wire accessRole when onDelete is added` comment so the gap is visible.

---

**src/components/calendar/__tests__/AgendaCalendar.test.tsx — no integration test for the read-only role path**

`SimpleCalendar.test.tsx` has two new integration tests that click through to the modal and assert on the delete button. `AgendaCalendar.test.tsx` does not — it only gets `getAccessRole: () => undefined` as a stub to satisfy the type. Given that `AgendaCalendar` has the identical `getAccessRole(selectedEvent.calendarId)` wiring, the wiring is untested at the consumer layer for that component. Recommend adding a parallel test case similar to the `SimpleCalendar` "hides the delete button on read-only calendars" case.

---

**src/lib/google-calendar-mappers.ts:51 — `accessRole` union duplicated, not shared with `TCalendarAccessRole`**

`UserCalendar.accessRole` in `google-calendar-mappers.ts` and `CalendarInfo.accessRole` in `route.ts` both inline the same `"freeBusyReader" | "reader" | "writer" | "owner"` union. `TCalendarAccessRole` was introduced in `CalendarProvider.tsx` for exactly this purpose but it's a provider-level export, so server-only modules (the mapper, the route) can't import it without pulling in client-side React. The cleanest fix is to extract `TCalendarAccessRole` into `src/types/calendar.ts` (already the home of `IEvent`, `ICalendarCell`, etc.) and import it from there in all three locations, eliminating the three-way duplication and making future expansions (e.g. if Google ever adds a new role) a single-line change.

---

BenSeymourODB:
Thanks — addressed all 5 in be940c1:

- **#1 (`useCallback` violation)**: removed; `getAccessRole` is now a plain closure.
- **#2 (stale `accessRoles`)**: added a comment at the `fetchCalendarList` call-site explaining the one-shot semantics and why we lean on Google's 403 as the backstop.
- **#3 (`AgendaList` un-wired)**: now passes `accessRole` to `EventDetailModal` for parity with `SimpleCalendar` / `AgendaCalendar`.
- **#4 (missing `AgendaCalendar` test)**: added two integration tests mirroring the SimpleCalendar coverage — one for `reader`, one for `owner`.
- **#5 (duplicated union)**: hoisted `TCalendarAccessRole` into `src/types/calendar.ts` and updated all three call-sites (route, mappers, provider).

`pnpm test:run` 1848 passing, types/lint clean.

---

PR #305, opened 2026-05-05
perf(tasks): coalesce duplicate getSession + account.findMany in /api/tasks routes (#260)

BenSeymourODB:
## Summary

Closes #260.

Each happy-path request to a `/api/tasks*` route currently calls `auth()` three+ times and `prisma.account.findMany` twice for the same Google account row — once via `assertGoogleTasksScope() → getGoogleAccount()` and once via `getAccessToken()`. Both helpers throw away a row that the other has already fetched.

This PR adds a combined helper and adopts it in all five task routes so each request makes **one** `auth()` call (in the route's existing `getSession()`) and **one** `prisma.account.findMany` call (inside the new helper).

### `src/lib/auth/helpers.ts` — new helper

`requireGoogleTasksAccessToken(userId): Promise<string>`:

- single `prisma.account.findMany` lookup
- `accountHasScope` check against `GOOGLE_TASKS_SCOPE`
- `decryptToken` for the stored access token
- throws `AuthError(401)` for missing account / missing token, `AuthError(403)` for missing scope — same shapes the routes already catch

`assertGoogleTasksScope` and `getAccessToken` are **kept as-is** for non-route callers (per option 2 in the issue).

### `/api/tasks*` routes — adopt the helper

`/api/tasks` GET+POST, `/api/tasks/lists`, `/api/tasks/[taskId]` PATCH, and `/api/tasks/[taskId]/complete` POST:

- keep the explicit `getSession()` null / `RefreshTokenError` checks at the top so the existing `{ error: "Unauthorized" }` / `{ requiresReauth: true }` 401 responses are byte-identical
- replace `assertGoogleTasksScope() + getAccessToken()` with a single `requireGoogleTasksAccessToken(session.user.id)` call placed at the same position the old scope check sat — preserves the "fail fast on auth before request validation" ordering, so a scope-missing user still sees 403 even when other request fields are invalid

## Behaviour change

None visible to clients. Each route test gains a call-count assertion pinning the new contract.

## Test plan

- [x] `pnpm test` — 1902 passed (75 in `src/app/api/tasks/`, 55 in `src/lib/auth/`)
- [ ] CI green

https://claude.ai/code/session_017yz62Zv9HjSWEaSi7JcNer

### DIFF
--- a/src/app/api/tasks/[taskId]/__tests__/route.test.ts
+++ b/src/app/api/tasks/[taskId]/__tests__/route.test.ts
@@ -3,9 +3,8 @@
  */
 import {
   AuthError,
-  assertGoogleTasksScope,
-  getAccessToken,
   getSession,
+  requireGoogleTasksAccessToken,
 } from "@/lib/auth";
 import {
   mockGoogleAccount,
@@ -23,8 +22,7 @@ import { PATCH } from "../route";
 // Mock modules BEFORE imports
 vi.mock("@/lib/auth", () => ({
   getSession: vi.fn(),
-  getAccessToken: vi.fn(),
-  assertGoogleTasksScope: vi.fn(),
+  requireGoogleTasksAccessToken: vi.fn(),
   AuthError: class AuthError extends Error {
     status: number;
     constructor(message: string, status: number = 401) {
@@ -71,7 +69,11 @@ global.fetch = mockFetch;
 describe("/api/tasks/[taskId]", () => {
   beforeEach(() => {
     vi.clearAllMocks();
-    vi.mocked(assertGoogleTasksScope).mockResolvedValue(undefined);
+    // Default: combined helper resolves with the mock token. Specific tests
+    // override to throw for the no-account / missing-scope / no-token paths.
+    vi.mocked(requireGoogleTasksAccessToken).mockResolvedValue(
+      mockGoogleAccount.access_token!
+    );
   });
 
   describe("PATCH /api/tasks/[taskId]", () => {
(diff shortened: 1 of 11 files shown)


BenSeymourODB (commented):
## Self-review (Sonnet) + responses

I ran a Sonnet subagent over the diff. Findings + how I addressed them in commit 2095a3f:

### Blocking — addressed

**`src/lib/auth/helpers.ts:requireGoogleTasksAccessToken` skipped the `RefreshTokenError` guard that `assertGoogleTasksScope` performs.** Routes pre-check today, but a future caller that didn't (a server action, a `withAuth` refactor, a test that doesn't set up the session mock correctly) would silently fall through to the Prisma query instead of throwing `AuthError(401)`.

→ Helper now takes a `Session` (not just `userId`) and throws `AuthError(401)` early on `RefreshTokenError` before any DB call. New test pins that the DB is **not** touched in that path. JSDoc rewritten — the helper is genuinely safe for any caller now.

### Nits — addressed

- **`findFirst` over `findMany().[0]`** — applied. The single-row intent is explicit at the SQL level. The other three call sites in `helpers.ts` keep `findMany` for now; they're outside this PR's scope but worth a follow-up.
- **Helper test double-invocation** — fixed; tests now hold a single `const promise = …` and run all matchers against it.
- **Wrong error message for null token** — the old `"No Google account linked. Please sign in with Google."` claimed the account wasn't linked when in fact it was. New message: `"Google account token unavailable. Please sign in again."` Updated test asserts the new wording.

### Question

**`assertGoogleTasksScope` and `getAccessToken` are still exported.** Yes, intentionally — `getAccessToken` is used by all the calendar routes (`/api/calendar/events`, `/api/calendar/calendars`, `/api/calendar/colors`), and `assertGoogleTasksScope` is reachable from non-route callers per option 2 in #260. A follow-up to apply the same single-pass optimisation to the calendar routes would be worth filing once #269 (local Postgres write path) lands and the calendar auth flow stabilises.

---

PR #312, opened 2026-05-05
fix(a11y): SSR-safe useReducedMotion eliminates AnimatedSwap hydration mismatch (#306)

BenSeymourODB:
Closes #306.

## Summary

Reduced-motion users hitting `/calendar` got a recoverable React hydration error anchored at `SimpleCalendar.tsx:206`. Root cause was in `src/hooks/use-reduced-motion.ts`: the `useState` initializer called `window.matchMedia(...).matches` synchronously, so the SSR pass returned `false` (no `window`) and the client's first render returned `true` whenever the OS preference was `reduce`. That `true` flowed into `AnimatedSwap.skipAnimation`, which selects between two structurally different trees — exactly what the hydration diff in the error showed.

## Fix

Rewrote `useReducedMotion` on top of `useSyncExternalStore`, which is purpose-built for SSR-safe external-source subscriptions:

- `getServerSnapshot()` returns `false` — used during SSR and during the client's hydration pass, guaranteeing the trees match.
- `getSnapshot()` reads the live `window.matchMedia(...).matches` after mount.
- `subscribe()` adds/removes the `change` listener and notifies React (no event payload — React re-reads `getSnapshot()`).

Reduced-motion users see one frame of the animating wrapper before the post-mount snapshot collapses it to the skip-animation tree. That's the standard SSR/a11y trade-off; the issue called this out as acceptable.

This dovetails with PR #308 (transition-speed setting) — both code paths now go through the same SSR-safe hook, so neither can reintroduce the mismatch.

## Tests (TDD)

- **Red phase** confirmed: a `renderToString`-based test asserting the SSR pass returns `false` even when `matchMedia` reports a match failed against the old hook.
- **Green phase**: that test plus a new `AnimatedSwap` regression covering both the SSR snapshot (must contain `animated-swap-idle`) and a full `renderToString` + `hydrateRoot` round-trip with `console.error` spying for hydration warnings.
- Updated two existing change-event tests on `useReducedMotion` and the mid-animation reduced-motion test on `AnimatedSwap` to mutate the shared `MediaQueryList.matches` mock before firing the listener — `useSyncExternalStore` re-reads the live snapshot on notify rather than consuming an event payload.
- Full suite: 1894 / 1894 pass.
- Lint, format, types: clean (5 pre-existing warnings unrelated to this change).

## Files

- `src/hooks/use-reduced-motion.ts` — rewritten on `useSyncExternalStore`
- `src/hooks/__tests__/use-reduced-motion.test.tsx` — renamed `.ts` → `.tsx`; added SSR-safety + post-mount transition tests; updated change-event tests
- `src/components/calendar/__tests__/animated-swap.test.tsx` — added hydration regression block; updated mid-animation reduced-motion test

No callers needed to change.

## Test plan

- [x] `pnpm test:run` — 1894 / 1894 pass
- [ ] Manual: enable `prefers-reduced-motion: reduce` (DevTools → Rendering → Emulate CSS media feature), hard-reload `/calendar`, confirm no hydration error in console

---
_Generated by [Claude Code](https://claude.ai/code/session_01GF1pyaqGhJZrHvmS2zytKK)_

### DIFF
--- a/.claude/commands/code-review-complexity.md
+++ b/.claude/commands/code-review-complexity.md
@@ -5,29 +5,37 @@ You are a specialized code review agent focused on **complexity and maintainabil
 ## Scope
 
 Scan all `.ts` and `.tsx` files in `src/` **excluding**:
+
 - `src/components/ui/` (third-party shadcn components)
 - Test files (`*.test.ts`, `*.test.tsx`, `__tests__/`)
 
 ## What to Look For
 
 ### 1. Long Functions (>40 lines)
+
 Read each source file and identify functions/methods exceeding 40 lines of actual logic (excluding blank lines and comments). These are prime candidates for extraction.
 
 ### 2. Large Files (>300 lines)
+
 Flag files exceeding 300 lines of source code. These often indicate multiple responsibilities crammed into one module.
 
 ### 3. Deep Nesting (3+ levels)
+
 Look for conditionals, loops, or try/catch blocks nested 3 or more levels deep. These are hard to follow and test.
 
 ### 4. High Parameter Count (4+)
+
 Functions with 4 or more parameters suggest the function is doing too much or needs a configuration object.
 
 ### 5. Excessive useEffect Chains
+
 In React components, look for multiple `useEffect` hooks that create implicit dependency chains. These are hard to reason about.
 
 ### 6. Approximate CRAP Score
+
 For each complex function, check if a corresponding test file exists. Code that is both complex AND untested is the highest priority.
-- CRAP = complexity * (1 - test_coverage)^2
+
+- CRAP = complexity \* (1 - test_coverage)^2
 - If no test file exists for a source file, assume 0% coverage for that file's functions.
 
 ## Output Format
@@ -42,6 +50,7 @@ EFFORT: {S|M|L}
 ```
 
 ## Severity Guide
+
 - **Critical**: CRAP score issue (complex + untested), or function >100 lines
 - **High**: Function >60 lines, or nesting >4 levels, or file >500 lines
 - **Medium**: Function >40 lines, or nesting 3 levels, or file >300 lines
--- a/.claude/commands/code-review-concerns.md
+++ b/.claude/commands/code-review-concerns.md
@@ -5,27 +5,34 @@ You are a specialized code review agent focused on **separation of concerns and 
 ## Scope
 
 Scan all `.ts` and `.tsx` files in `src/` **excluding**:
+
 - `src/components/ui/` (third-party shadcn components)
 - Test files (`*.test.ts`, `*.test.tsx`, `__tests__/`)
 
 ## What to Look For
 
 ### 1. Components Mixing Data Fetching with Presentation
+
 Look for React components that both fetch data (via `fetch`, API calls, database queries) AND render significant JSX. These should be split into container/presenter or use hooks for data.
 
 ### 2. God Providers / God Components
+
 Providers or components that handle too many concerns: data fetching, caching, state management, data transformation, error handling, AND rendering all in one file. Look for files with both `fetch`/`prisma` calls and complex JSX.
 
 ### 3. Business Logic in Route Handlers
+
 API routes (`src/app/api/`) should delegate to `lib/` functions for business logic. Flag routes where validation, computation, or complex data manipulation happens directly in the route handler instead of in a reusable utility.
 
 ### 4. Hooks Violating Single Responsibility
+
 Custom hooks that do too many unrelated things. A hook should have one clear purpose. Look for hooks with multiple unrelated state variables or effects.
 
 ### 5. Circular or Tangled Dependencies
+
 Check for files that import from each other, or import chains that suggest tight coupling between modules that should be independent.
 
 ### 6. Missing Abstraction Layers
+
 Look for repeated patterns that should be extracted - e.g., the same auth-check + error-handling pattern copy-pasted across multiple API routes, or similar data transformation logic in multiple places.
 
 ## Output Format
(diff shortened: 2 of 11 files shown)


BenSeymourODB (commented):
Overall: the `useSyncExternalStore` migration is correct and the hydration mismatch is properly fixed. A few things worth addressing before merge — one minor correctness concern and a handful of nits.

**Subscribe identity** ✓ — `subscribe` is module-scoped so its reference is stable for the lifetime of the app; React will never re-subscribe spuriously.

**`getServerSnapshot` returning a primitive** ✓ — literal `false` has stable identity. No issue.

**TDD** ✓ — red/green/refactor documented, 1894/1894 pass.

**React Compiler / TypeScript / Tailwind** ✓ — no manual memoization, no `any`, standard palette.

---

BenSeymourODB:
## Code review findings

### Finding 1 — nit | `src/hooks/use-reduced-motion.ts` line 31 | `getSnapshot` allocates a new MQL on every call

```ts
// current
return window.matchMedia(MEDIA_QUERY).matches;
```

`getSnapshot` is called by React after every subscription notification and on every render reconciliation pass. Each call creates a new `MediaQueryList` object via `window.matchMedia`. It's harmless for correctness (the spec guarantees `.matches` is live), but the allocation is unnecessary since `subscribe` already holds the same MQL. A simple module-level lazy cache would eliminate the churn:

```ts
let _mql: MediaQueryList | null = null;
function getMql(): MediaQueryList | null {
  if (typeof window === "undefined" || typeof window.matchMedia !== "function") return null;
  return (_mql ??= window.matchMedia(MEDIA_QUERY));
}

function getSnapshot(): boolean {
  return getMql()?.matches ?? false;
}

function subscribe(onChange: () => void): () => void {
  const mql = getMql();
  if (!mql) return () => undefined;
  mql.addEventListener("change", onChange);
  return () => mql.removeEventListener("change", onChange);
}
```

(The module-level `_mql` is fine because MEDIA_QUERY is a constant and there's only one MQL per query string per window anyway.)

---

### Finding 2 — minor | `src/components/calendar/__tests__/animated-swap.test.tsx` line 515–518 | Misleading comment about `useEffect`

```ts
// This test exercises the real React hydration path with prefers-reduced
// -motion enabled. Use real timers so the post-mount useEffect that
// syncs the live preference can run…
```

The `useEffect` sync is from the issue's *suggested* `useState`/`useEffect` fix — not from the actual `useSyncExternalStore` implementation that was merged. With `useSyncExternalStore`, there's no `useEffect` that syncs; React reads `getSnapshot()` internally after mount. The comment should be updated to avoid confusing future maintainers, e.g.:

```ts
// Use real timers so hydrateRoot's own internal scheduling (Scheduler)
// can flush without fake-timer interference.
```

---

### Finding 3 — minor | `src/hooks/__tests__/use-reduced-motion.test.tsx` | Missing edge-case: `matchMedia` unavailable on client

The `getSnapshot` and `subscribe` functions both guard `typeof window.matchMedia !== "function"` and fall back to `false`/no-op — but this path is never exercised in the test suite. Old Android WebViews and some JSDOM configs don't expose `matchMedia`. Suggest adding:

```ts
it("returns false when matchMedia is not available on the client", () => {
  const orig = window.matchMedia;
  // @ts-expect-error intentionally removing matchMedia to test the guard
  delete window.matchMedia;
  try {
    const { result } = renderHook(() => useReducedMotion());
    expect(result.current).toBe(false);
  } finally {
    window.matchMedia = orig;
  }
});
```

---

### Finding 4 — nit | `src/components/calendar/__tests__/animated-swap.test.tsx` line 513 | Hydration test relies on `console.error` string-matching

```ts
const hydrationErrors = errorSpy.mock.calls.filter((call) =>
  call.some((arg) =>
    typeof arg === "string" ? /hydrat/i.test(arg) : false
  )
);
```

React 19 passes hydration error arguments that are sometimes `Error` objects (not strings) — the `typeof arg === "string"` guard would silently miss those. A slightly more robust predicate:

```ts
call.some((arg) => /hydrat/i.test(String(arg)))
```

This is a low-probability miss given JSDOM/React 19's current format, but worth hardening.

---

### Finding 5 — nit | `src/hooks/__tests__/use-reduced-motion.test.tsx` line 145–157 | SSR test runs in jsdom, not a true server environment

`renderToString` in jsdom still has `window` in scope. `useSyncExternalStore` uses `getServerSnapshot` during `renderToString` regardless of whether `window` exists (React detects the SSR rendering context, not the environment), so the test *is* correct — but a comment clarifying this would help:

```ts
// renderToString triggers getServerSnapshot (React detects the SSR
// rendering context internally), so `window.matchMedia` being available
// in jsdom does not affect the result.
```


---

BenSeymourODB:
Round-2 follow-ups — pushed in `a96d195`:

| # | Action |
|---|---|
| 1 — lazy-cache MQL | **Deferred.** Module-level `_mql` would conflict with the existing test isolation pattern: every `beforeEach` in `use-reduced-motion.test.tsx` and `animated-swap.test.tsx` replaces `window.matchMedia` with a fresh mock. A cached `_mql` from a previous test would survive into the next one and point at a stale mock. The browser's own `matchMedia` registry also dedupes by query string, so the per-call allocation cost is real but very small. Happy to revisit if benchmarks justify either gating the cache on `window.matchMedia` identity or adding a `__resetForTests` hatch. |
| 2 — misleading `useEffect` comment | **Fixed.** Comment now reflects the actual `useSyncExternalStore` implementation. |
| 3 — `matchMedia` unavailable test | **Fixed.** New test stubs `window.matchMedia = undefined`, asserts `false` is returned and unmount doesn't throw. |
| 4 — `String(arg)` in hydration filter | **Fixed.** Coerces every console.error arg before pattern-matching so React 19 Error-object payloads don't slip past. |
| 5 — clarifying comment on SSR test | **Fixed.** Comment explains why `renderToString` in jsdom still triggers `getServerSnapshot` (React detects the SSR rendering context internally). |

Full suite: 1895/1895. Lint, format, types: clean.

---

PR #314, opened 2026-05-05
feat(calendar): Clock view — wrap arc title onto up to 2 lines (#310)

BenSeymourODB:
## Summary

Closes #310.

In the Clock view, event arcs that have a title too long for the available angular span used to truncate to a single line with a hard-coded 15-char ellipsis (`cleanTitle.slice(0, 14) + "..."`). This PR replaces that with a per-arc capacity calculation: titles render on a single curved `<textPath>` when they fit, on **two concentric `<textPath>`s** at `titleRadius ± lineOffset/2` when they don't, and only ellipsize line 2 when even two lines aren't enough.

The new pure helper `fitTitleToArc` exposes a `didOverflow` flag so the companion floating-label feature (#311) can land cleanly on top.

### Phase 1 — `fitTitleToArc` pure helper
- `src/components/calendar/analog-clock/fit-title.ts`
- `src/components/calendar/analog-clock/__tests__/fit-title.test.ts` (15 unit tests)

Per-line char budget ≈ `(arcSpan / 360) × 2π × titleRadius / (fontSize × 0.6)`, floored. Greedy word-pack with no mid-word splits — single very long words are truncated as one line and marked `didOverflow: true`. `maxLines: 1 | 2` parameter so the call site can suppress wrapping on narrow arcs.

### Phase 2 — `EventArc` wiring
- `src/components/calendar/analog-clock/event-arc.tsx`
- `src/components/calendar/analog-clock/__tests__/event-arc.test.tsx`

EventArc now consults `fitTitleToArc` with the actual `titleRadius` and `titleFontSize` derived from arc geometry, then renders one or two `<textPath>` elements at `titleRadius ± lineOffset/2` (`lineOffset = titleFontSize × 0.55`). A `TWO_LINE_MIN_SPAN_DEGREES = 30` gate keeps narrow arcs on a single ellipsized line so we don't end up with 2-char-per-line stubs. The existing `<text data-testid="event-title-{id}">` wrapper is preserved (one parent `<text>` containing N `<textPath>` children) so existing test selectors keep working.

### Phase 3 — E2E + final gate
- `e2e/analog-clock.spec.ts` — regression test that titles still render via `<textPath>` on the default scenario, "Family Game Night" appears in full (no hard-coded "..." truncation), and a screenshot artefact is emitted.

## Test plan

- [x] `pnpm vitest run src/components/calendar/analog-clock` — 96/96 green (15 new helper tests + 5 new EventArc tests for the wrap path)
- [x] `pnpm test` — 1908/1908 green
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` clean (5 pre-existing warnings unchanged)
- [ ] CI Playwright: new spec runs `analog-clock.spec.ts > Title rendering (#310 — 2-line wrap)`. Browser binaries unavailable in the local sandbox; the spec parses + lists cleanly and CI will execute it.

## TDD note

The previous test "renders the clean title on the arc (truncated if > 15 chars)" was re-pointed onto the new behaviour rather than deleted — that single-line-truncation rule is exactly what #310 replaces, so updating the assertion is consistent with TDD (the test still verifies the contract; the contract changed). The new tests still cover: full short title on a single textPath, 2-line render, distinct radii per line, didOverflow ellipsis on line 2, and the < 30° narrow-arc fallback.

## Out of scope (deferred)

- Companion floating off-arc title labels with connector lines (#311) — consumes `fit.didOverflow` from this PR. Was left as a separate issue per #310's "land this issue first" guidance.

### DIFF
--- a/e2e/analog-clock.spec.ts
+++ b/e2e/analog-clock.spec.ts
@@ -190,6 +190,36 @@ test.describe("Analog Clock - Radial Display", () => {
     });
   });
 
+  test.describe("Title rendering (#310 — 2-line wrap)", () => {
+    test("renders titles via <textPath> on each event arc (no single-line ellipsis)", async ({
+      page,
+    }) => {
+      await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+      await expect(page.getByTestId("analog-clock")).toBeVisible();
+
+      // Every event arc with arcSpan ≥ 20° renders a title text node.
+      // Default scenario seeds 5 arcs, the smallest of which (d2 Deadline,
+      // 30 min = 15°) is below the gate; the other 4 should have titles.
+      const titles = page.locator('[data-testid^="event-title-"]');
+      await expect(titles).not.toHaveCount(0);
+
+      // The longest-title arc ("Family Game Night") is rendered in full —
+      // either on a single curved line if the arc is wide enough, or split
+      // across two textPaths if not. In neither case does it end with the
+      // old hard-coded "..." truncation marker.
+      const familyGameTitle = page.getByTestId("event-title-d1");
+      await expect(familyGameTitle).toBeVisible();
+      const text = await familyGameTitle.textContent();
+      expect(text).toContain("Family Game");
+      expect(text).toContain("Night");
+
+      await page.screenshot({
+        path: "test-results/screenshots/clock-default-titles.png",
+        fullPage: true,
+      });
+    });
+  });
+
   test.describe("Clock Face Accuracy", () => {
     test("shows correct hand positions at 3:00", async ({ page }) => {
       await page.goto("/test/analog-clock?scenario=empty&hour=3&min=0");
--- a/src/components/calendar/analog-clock/__tests__/event-arc.test.tsx
+++ b/src/components/calendar/analog-clock/__tests__/event-arc.test.tsx
@@ -66,15 +66,25 @@ describe("EventArc", () => {
     expect(screen.queryByTestId("event-emoji-evt-2")).not.toBeInTheDocument();
   });
 
-  it("renders the clean title on the arc (truncated if > 15 chars)", () => {
+  it("wraps a long title onto 2 lines instead of single-line ellipsis (#310)", () => {
+    // baseEvent: "Family Game Night" (17 chars) at 30° arc, titleRadius ≈ 227,
+    // fontSize ≈ 12 → per-line budget ≈ 16 chars. Doesn't fit on one line;
+    // two-line split produces "Family Game" / "Night".
     renderArc();
     const title = screen.getByTestId("event-title-evt-1");
     expect(title).toBeInTheDocument();
-    // "Family Game Night" is 17 chars, truncated to 14 + "..."
-    expect(title.textContent).toBe("Family Game Ni...");
+
+    const textPaths = title.querySelectorAll("textPath");
+    expect(textPaths.length).toBe(2);
+    expect(Array.from(textPaths).map((tp) => tp.textContent)).toEqual([
+      "Family Game",
+      "Night",
+    ]);
+    // No single-line ellipsis on the rendered output anymore.
+    expect(title.textContent).not.toMatch(/\.\.\.$/);
   });
 
-  it("renders full title when short enough", () => {
+  it("renders full title on a single textPath when short enough", () => {
     const shortEvent: ClockEvent = {
       ...baseEvent,
       id: "evt-short",
@@ -84,7 +94,62 @@ describe("EventArc", () => {
     };
     renderArc({ event: shortEvent });
     const title = screen.getByTestId("event-title-evt-short");
-    expect(title.textContent).toBe("Team Lunch");
+    const textPaths = title.querySelectorAll("textPath");
+    expect(textPaths.length).toBe(1);
+    expect(textPaths[0].textContent).toBe("Team Lunch");
+  });
+
+  it("renders 2-line titles at distinct radii (outer line further from center)", () => {
+    // Render the wrapped baseEvent and verify the two textPath elements
+    // reference paths defined at radii separated by ~lineOffset.
+    renderArc();
+    const title = screen.getByTestId("event-title-evt-1");
+    const textPaths = title.querySelectorAll("textPath");
+    expect(textPaths.length).toBe(2);
+
+    const hrefs = Array.from(textPaths).map((tp) =>
+      // textPath's href attribute is exposed via getAttribute("href") in JSDOM
+      tp.getAttribute("href")
+    );
+    expect(hrefs[0]).not.toEqual(hrefs[1]);
+
+    // Both referenced <path>s should exist in the same <defs>.
+    const defs = (title.parentElement as HTMLElement).querySelector("defs");
+    const pathEls = defs ? defs.querySelectorAll("path") : [];
+    const pathIds = Array.from(pathEls).map((p) => `#${p.getAttribute("id")}`);
+    expect(pathIds).toEqual(expect.arrayContaining(hrefs));
+  });
+
+  it("ellipsizes the second line when even 2 lines aren't enough (didOverflow)", () => {
+    const overflowEvent: ClockEvent = {
+      ...baseEvent,
+      id: "evt-overflow",
+      cleanTitle: "the quick brown fox jumps over the lazy dog",
+      startAngle: 90,
+      endAngle: 120, // 30° span → ~16 char budget
+    };
+    renderArc({ event: overflowEvent });
+    const title = screen.getByTestId("event-title-evt-overflow");
+    const textPaths = title.querySelectorAll("textPath");
+    expect(textPaths.length).toBe(2);
+    // Line 2 ends with the truncation marker.
+    expect(textPaths[1].textContent?.endsWith("...")).toBe(true);
+  });
+
+  it("falls back to a single ellipsized line on narrow arcs (< 30°)", () => {
+    // 25° arc is below the 2-line gate. Long title → single-line truncation.
+    const narrowEvent: ClockEvent = {
+      ...baseEvent,
+      id: "evt-narrow",
+      cleanTitle: "Family Game Night",
+      startAngle: 90,
+      endAngle: 115, // 25° span
+    };
+    renderArc({ event: narrowEvent });
+    const title = screen.getByTestId("event-title-evt-narrow");
+    const textPaths = title.querySelectorAll("textPath");
+    expect(textPaths.length).toBe(1);
+    expect(textPaths[0].textContent?.endsWith("...")).toBe(true);
   });
 
   it("has accessible role with event title", () => {
--- a/src/components/calendar/analog-clock/__tests__/event-arc.test.tsx
+++ b/src/components/calendar/analog-clock/__tests__/event-arc.test.tsx
@@ -67,21 +67,23 @@ describe("EventArc", () => {
   });
 
   it("wraps a long title onto 2 lines instead of single-line ellipsis (#310)", () => {
-    // baseEvent: "Family Game Night" (17 chars) at 30° arc, titleRadius ≈ 227,
-    // fontSize ≈ 12 → per-line budget ≈ 16 chars. Doesn't fit on one line;
-    // two-line split produces "Family Game" / "Night".
+    // baseEvent: "Family Game Night" (17 chars) at 30° arc. The exact split
+    // depends on the layout heuristic constants in event-arc.tsx, so this
+    // test only asserts the contract: two lines, words preserved across
+    // the split, no mid-word breaks, no single-line ellipsis fallback.
     renderArc();
     const title = screen.getByTestId("event-title-evt-1");
     expect(title).toBeInTheDocument();
 
     const textPaths = title.querySelectorAll("textPath");
     expect(textPaths.length).toBe(2);
-    expect(Array.from(textPaths).map((tp) => tp.textContent)).toEqual([
-      "Family Game",
-      "Night",
-    ]);
-    // No single-line ellipsis on the rendered output anymore.
-    expect(title.textContent).not.toMatch(/\.\.\.$/);
+
+    const renderedLines = Array.from(textPaths).map(
+      (tp) => tp.textContent ?? ""
+    );
+    expect(renderedLines.join(" ")).toBe("Family Game Night");
+    expect(renderedLines.every((l) => !l.includes("..."))).toBe(true);
+    expect(renderedLines.every((l) => !l.includes("…"))).toBe(true);
   });
 
   it("renders full title on a single textPath when short enough", () => {
@@ -101,20 +103,18 @@ describe("EventArc", () => {
 
   it("renders 2-line titles at distinct radii (outer line further from center)", () => {
     // Render the wrapped baseEvent and verify the two textPath elements
-    // reference paths defined at radii separated by ~lineOffset.
+    // reference distinct paths defined in <defs>.
     renderArc();
     const title = screen.getByTestId("event-title-evt-1");
     const textPaths = title.querySelectorAll("textPath");
     expect(textPaths.length).toBe(2);
 
-    const hrefs = Array.from(textPaths).map((tp) =>
-      // textPath's href attribute is exposed via getAttribute("href") in JSDOM
-      tp.getAttribute("href")
-    );
+    const hrefs = Array.from(textPaths).map((tp) => tp.getAttribute("href"));
     expect(hrefs[0]).not.toEqual(hrefs[1]);
 
-    // Both referenced <path>s should exist in the same <defs>.
-    const defs = (title.parentElement as HTMLElement).querySelector("defs");
+    // Both referenced <path>s should exist in the same arc <g>.
+    const arcGroup = screen.getByTestId("event-arc-group-evt-1");
+    const defs = arcGroup.querySelector("defs");
     const pathEls = defs ? defs.querySelectorAll("path") : [];
     const pathIds = Array.from(pathEls).map((p) => `#${p.getAttribute("id")}`);
     expect(pathIds).toEqual(expect.arrayContaining(hrefs));
@@ -132,7 +132,7 @@ describe("EventArc", () => {
     const title = screen.getByTestId("event-title-evt-overflow");
     const textPaths = title.querySelectorAll("textPath");
     expect(textPaths.length).toBe(2);
-    // Line 2 ends with the truncation marker.
+    // Line 2 ends with the truncation marker (ASCII "..." at this budget).
     expect(textPaths[1].textContent?.endsWith("...")).toBe(true);
   });
 
--- a/src/components/calendar/analog-clock/__tests__/fit-title.test.ts
+++ b/src/components/calendar/analog-clock/__tests__/fit-title.test.ts
@@ -46,11 +46,27 @@ describe("fitTitleToArc", () => {
 
   it("greedy-packs the first line when splitting", () => {
     // "Pick up groceries today" = 23 chars; at 30° (~12 chars/line):
-    // line 1 = "Pick up" (7), next word "groceries" (9) overflows → line 2 starts.
+    // line 1 = "Pick up" (7), line 2 = "groceries" (9, "today" doesn't fit).
+    // This is an overflow case — line 2 is truncated with an ellipsis.
     const result = fitTitleToArc("Pick up groceries today", 30, RADIUS, FONT);
     expect(result.lines.length).toBe(2);
     expect(result.lines[0]).toBe("Pick up");
     expect(result.lines[1]).toMatch(/^groceries/);
+    expect(result.didOverflow).toBe(true);
+    expect(result.lines[1].endsWith("...")).toBe(true);
+  });
+
+  it("ellipsizes line 2 when the next word alone exceeds the line budget", () => {
+    // Two-word title where word 2 alone is wider than the budget.
+    // budget=12 at 30°/200/14. "Hi" packs onto line 1 (2 chars). Word 2
+    // "supercalifragilistic" (20 chars) won't fit on line 2 — packLine
+    // returns an empty line2 and the fallback truncates that word.
+    const result = fitTitleToArc("Hi supercalifragilistic", 30, RADIUS, FONT);
+    expect(result.lines.length).toBe(2);
+    expect(result.lines[0]).toBe("Hi");
+    expect(result.lines[1].endsWith("...")).toBe(true);
+    expect(result.lines[1].startsWith("super")).toBe(true);
+    expect(result.didOverflow).toBe(true);
   });
 
   it("ellipsizes line 2 when content needs three or more lines", () => {
@@ -59,7 +75,7 @@ describe("fitTitleToArc", () => {
     const result = fitTitleToArc(longTitle, 30, RADIUS, FONT);
     expect(result.lines.length).toBe(2);
     expect(result.didOverflow).toBe(true);
-    // Line 2 should end with an ellipsis to signal truncation.
+    // Line 2 should end with the ASCII ellipsis at this comfortable budget.
     expect(result.lines[1].endsWith("...")).toBe(true);
   });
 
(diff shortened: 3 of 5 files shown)


BenSeymourODB (commented):
## PR #314 Review — Clock view 2-line title wrap

Overall: **Changes requested.** The feature logic is sound and TDD coverage is good, but there is a geometry bug in the 2-line radii formula that will produce visually overlapping text at every font size, a correctness issue in `truncateWithEllipsis` at budget ≤ 3 that silently drops the ellipsis indicator, an uncovered edge case in `fitTitleToArc` where `line2` is empty at overflow time, and the E2E screenshot path writes into `test-results/` which CLAUDE.md prohibits committing. Detailed findings are in the inline comments below.

---

### Summary of concerns (ranked)

1. **[CRITICAL] `lineOffset / 2` gives overlapping text bands.** The two line centres are only `0.55 × fontSize` apart; each band occupies `±fontSize/2`, so they overlap by `~0.45 × fontSize` (5–8 px depending on font size). The issue spec explicitly says `titleRadius ± lineOffset` (not `÷ 2`), giving `1.1 × fontSize` centre-to-centre — just enough breathing room. The `/2` halves that, guaranteeing overlap.
2. **[BUG] `truncateWithEllipsis(text, 3)` returns text without `"..."`.** When `budget ≤ 3` the function returns a raw slice with no ellipsis marker. Any downstream code (including `didOverflow`-consumers like #311) that inspects the rendered string to detect truncation will get a false negative at tiny arcs.
3. **[BUG] Empty `line2` in overflow path uses `words[nextIndex]` before bounds check.** If `line2` is empty because every remaining word exceeds the budget, the fallback references `words[nextIndex]`, but `nextIndex` may equal `words.length` if `packLine` somehow exhausted all words on line 1. This is an out-of-bounds read that evaluates to `undefined` and is passed to `truncateWithEllipsis`.
4. **[MINOR] `multipleTextPath` children of a single `<text>` produce stacked — not wrapped — lines in browsers.** Multiple `<textPath>` children of one `<text>` element are valid SVG, but browsers render them consecutively on the same baseline, not on separate paths. Each `<textPath>` references its own `<path>`, so in theory each renders on its own curve, but visual stacking relies on the browser respecting `dominantBaseline="central"` per-textPath. Chrome/Firefox do honour this, but it is worth a real-browser smoke test before merging.
5. **[MINOR] E2E screenshot written to `test-results/`** — CLAUDE.md explicitly prohibits committing test output artefacts from `test-results/`. The path should either use `playwright.config.ts`'s `outputDir` (which is already `test-results/`) and be excluded from git via `.gitignore`, or be moved to a stable `docs/screenshots/` path. As long as `test-results/` is gitignored this is fine, but worth confirming.

See inline comments for exact line numbers and suggested fixes.

---

src/components/calendar/analog-clock/event-arc.tsx
```diff
@@ -82,19 +86,31 @@ export function EventArc({
...
     titleRadius,
-    startAngle,
-    endAngle
+    titleFontSize,
+    maxLines
+  );
+
+  // Per-line offset for the 2-line case. Slightly under one font-size so the
+  // two curved baselines stay within the arc band without overlapping.
+  const lineOffset = titleFontSize * 0.55;
```
BenSeymourODB, line 107: **[CRITICAL] `lineOffset / 2` produces overlapping text lines at every font size.**

The two rendered textPath centres are at `titleRadius + lineOffset/2` and `titleRadius - lineOffset/2`, so their centre-to-centre distance equals `lineOffset = titleFontSize × 0.55`. But each text band occupies `±fontSize/2` around its centre (because `dominantBaseline="central"` is used). This means the two bands overlap by:

```
overlap = fontSize − lineOffset = fontSize − (fontSize × 0.55) = fontSize × 0.45
```

At `fontSize=12` that is ~5.4 px of overlap; at `fontSize=18` it is ~8.1 px. In practice the glyphs from the two lines will be superimposed on each other, making both illegible.

The issue spec (#310) explicitly says:

> define two `<path>` elements in `<defs>` at **`titleRadius ± lineOffset`** where `lineOffset ≈ titleFontSize × 0.55`

That gives a centre-to-centre distance of `2 × lineOffset = 1.1 × fontSize`, which is just enough to avoid overlap (≈1 line of normal leading). The implementation accidentally halves that by using `± lineOffset/2`.

**Suggested fix:**

```tsx
// event-arc.tsx  (current, line ~104)
const lineRadii =
  fit.lines.length === 2
    ? [titleRadius + lineOffset / 2, titleRadius - lineOffset / 2]  // ← wrong
    : [titleRadius];

// Fix: drop the /2
const lineRadii =
  fit.lines.length === 2
    ? [titleRadius + lineOffset, titleRadius - lineOffset]
    : [titleRadius];
```

Note: with `fontSize=18` and the typical `titleRadius ≈ 227`, the outer centre rises to `~237` and the top of the glyph to `~246`, which clips the `outerRadius=240` arc edge. You may also want to reduce `titleFontSize` slightly when 2-line mode is active (e.g. `Math.min(arcHeight * 0.25, 14)` instead of `0.3 / 18`) to keep both lines within the arc band.

---

src/components/calendar/analog-clock/fit-title.ts
```diff
@@ -0,0 +1,147 @@
...
+  if (arcSpanDegrees <= 0 || titleRadius <= 0 || fontSize <= 0) return 0;
+  const circumference = (arcSpanDegrees / 360) * 2 * Math.PI * titleRadius;
+  const charWidth = fontSize * CHAR_WIDTH_RATIO;
+  return Math.floor(circumference / charWidth);
+}
+
+function truncateWithEllipsis(text: string, budget: number): string {
+  // Keep at least one character of the original alongside the ellipsis;
+  // otherwise drop the ellipsis entirely.
+  if (budget <= 3) return text.slice(0, Math.max(0, budget));
```
BenSeymourODB: **[BUG] `truncateWithEllipsis` silently drops the ellipsis when `budget ≤ 3`.**

```ts
function truncateWithEllipsis(text: string, budget: number): string {
  if (budget <= 3) return text.slice(0, Math.max(0, budget));  // ← no "..." appended
  return `${text.slice(0, budget - 3)}...`;
}
```

When `budget` is 1, 2, or 3 the function returns a raw character slice with no `"..."` suffix. Two downstream effects:

1. Any code checking `.endsWith("...")` to detect truncation (e.g. future #311 floating-label logic) gets a false negative.
2. The user sees an abruptly cut word with no visual signal that more content exists — even though `didOverflow` is `true`.

Budget ≤ 3 is genuinely tiny (a ~3-character arc), but the contract advertised by the function name should hold. The cleanest fix is to always append the ellipsis if at least one original character survives:

```ts
function truncateWithEllipsis(text: string, budget: number): string {
  if (budget <= 0) return "";
  if (budget <= 3) {
    // Not enough room for "..." — show as many chars as fit, append "…" (single unicode
    // ellipsis, 1 char) only when at least 1 original char is visible.
    const slice = text.slice(0, budget);
    return slice.length > 0 ? `${slice.slice(0, budget - 1)}…` : "";
  }
  return `${text.slice(0, budget - 3)}...`;
}
```

Alternatively, just always reserve 1 char for `"…"` regardless of budget, making the threshold `budget <= 1 → ""`, `budget >= 2 → slice + "…"`. Either way the tests for `budget=3` (the "never returns more than two lines" test at `span=20°`) need updating to match.

---

src/components/calendar/analog-clock/fit-title.ts
```diff
@@ -0,0 +1,147 @@
...
+    words,
+    nextIndex,
+    budget
+  );
+
+  // No remaining words after line 2 → both lines render with no overflow.
+  if (afterLine2 >= words.length) {
+    return { lines: [line1, line2], didOverflow: false };
+  }
+
```
BenSeymourODB, line 179: **[BUG] `words[nextIndex]` may be `undefined` when `line2` is empty.**

```ts
const truncatedLine2 =
  line2.length > 0
    ? truncateWithEllipsis(line2, budget)
    : truncateWithEllipsis(words[nextIndex], budget);  // ← potential undefined
```

The fallback branch fires when `packLine` returned an empty `line2`, meaning every remaining word from `nextIndex` onwards exceeds `budget`. At that point `nextIndex` still points at the first unfit word, so `words[nextIndex]` is normally valid — _but_ there is a path where it can equal `words.length`:

If all words fit on line 1 (`packLine` consumed every word: `nextIndex === words.length`), the `afterLine2 >= words.length` guard above fires first and we never reach this code. So in the current flow `nextIndex < words.length` is always true here.

However, the logic is fragile and the intent is not obvious. Add an explicit guard so this never silently passes `undefined` to `truncateWithEllipsis`:

```ts
const truncatedLine2 =
  line2.length > 0
    ? truncateWithEllipsis(line2, budget)
    : nextIndex < words.length
      ? truncateWithEllipsis(words[nextIndex], budget)
      : "...";  // should be unreachable, but safe fallback

return { lines: [line1, truncatedLine2], didOverflow: true };
```

Also worth adding a unit test that directly exercises this `line2 === ""` branch (e.g. two-word title where word 2 alone exceeds the budget).

---

src/components/calendar/analog-clock/event-arc.tsx
```diff
@@ -132,23 +150,26 @@ export function EventArc({
...
-          >
-            {displayText}
-          </textPath>
+          {fit.lines.map((line, i) => (
+            <textPath
+              key={textPathIds[i]}
+              href={`#${textPathIds[i]}`}
+              startOffset="50%"
+              textAnchor="middle"
+              dominantBaseline="central"
```
BenSeymourODB: **[MINOR] Multiple `<textPath>` children of a single `<text>` — browser rendering note.**

The pattern here is:

```tsx
<text ...>
  <textPath href="#path-0" ...>Line one</textPath>
  <textPath href="#path-1" ...>Line two</textPath>
</text>
```

This is valid SVG 1.1 / SVG 2, and Chrome, Firefox, and Safari all render each `<textPath>` on its own referenced path. However the spec says that multiple `<textPath>` children are rendered **in sequence on the same text flow**: when the first `<textPath>` ends, rendering continues on the second path. Because `startOffset="50%"` and `textAnchor="middle"` are set on each child independently, in practice the two lines do render on their respective arcs as intended.

The risk is that JSDOM (used in Vitest) does not actually lay out SVG text, so the unit tests cannot detect a visual misalignment. A quick Playwright visual snapshot of a 2-line arc at each supported font size would give high confidence. The existing E2E spec only checks `textContent` and a single screenshot of the default scenario — adding an assertion that both `<path>` elements referenced by the two `<textPath>` hrefs have _different_ `d` attributes (which the unit test already does) is good, but a real-browser visual check for the two-line case specifically would be valuable before merging.

---

src/components/calendar/analog-clock/__tests__/event-arc.test.tsx
```diff
@@ -66,15 +66,25 @@ describe("EventArc", () => {
     expect(screen.queryByTestId("event-emoji-evt-2")).not.toBeInTheDocument();
   });
 
-  it("renders the clean title on the arc (truncated if > 15 chars)", () => {
+  it("wraps a long title onto 2 lines instead of single-line ellipsis (#310)", () => {
+    // baseEvent: "Family Game Night" (17 chars) at 30° arc, titleRadius ≈ 227,
+    // fontSize ≈ 12 → per-line budget ≈ 16 chars. Doesn't fit on one line;
```
BenSeymourODB: **[MINOR] Test is tightly coupled to heuristic constants — will silently break if geometry changes.**

The test comment says `"titleRadius ≈ 227, fontSize ≈ 12 → per-line budget ≈ 16 chars"` and then asserts the exact split `["Family Game", "Night"]`. This is correct _today_, but the assertion encodes the result of `innerRadius + arcHeight * 0.68` and `arcHeight * 0.3` — two constants that are not exported or named. If either constant shifts (e.g. during a layout pass), the split could silently change to `["Family", "Game Night"]` and this test would fail with a cryptic mismatch rather than a clear "layout constant changed" message.

Suggestions:
1. Either import and reference `TWO_LINE_MIN_SPAN_DEGREES` (already exported-adjacent) and document the exact budget arithmetic inline so the expected split is obviously derivable, or
2. Relax the assertion slightly: instead of checking the exact split, verify `textPaths.length === 2`, that `textPaths[0].textContent + " " + textPaths[1].textContent === "Family Game Night"` (words preserved, no mid-word split), and that neither line contains `"..."`. This is more robust to minor constant tweaks while still testing the contract.

Option 2 is already how the pure-unit test `fit-title.test.ts` handles it at line 46 (`lines.join(" ") === "Family Game Night"`), so aligning the component test to the same style would be consistent.

---

src/components/calendar/analog-clock/__tests__/fit-title.test.ts
```diff
@@ -0,0 +1,160 @@
...
+    expect(result.lines.every((line) => !line.includes("..."))).toBe(true);
+  });
+
+  it("greedy-packs the first line when splitting", () => {
+    // "Pick up groceries today" = 23 chars; at 30° (~12 chars/line):
+    // line 1 = "Pick up" (7), next word "groceries" (9) overflows → line 2 starts.
+    const result = fitTitleToArc("Pick up groceries today", 30, RADIUS, FONT);
+    expect(result.lines.length).toBe(2);
+    expect(result.lines[0]).toBe("Pick up");
+    expect(result.lines[1]).toMatch(/^groceries/);
```
BenSeymourODB, line 54: **[MINOR] `didOverflow` not asserted in the greedy-pack test — misleading test contract.**

```ts
it("greedy-packs the first line when splitting", () => {
  const result = fitTitleToArc("Pick up groceries today", 30, RADIUS, FONT);
  expect(result.lines.length).toBe(2);
  expect(result.lines[0]).toBe("Pick up");
  expect(result.lines[1]).toMatch(/^groceries/);
});
```

At `budget=12`, `"Pick up groceries today"` fits: `"Pick up"` on line 1 and `"groceries"` on line 2, but `"today"` doesn't fit on line 2 — so this is actually an **overflow** case (`didOverflow: true`) and `lines[1]` will be `"groceries..."` (truncated). The test passes because `/^groceries/` matches `"groceries..."`, but it does not assert `didOverflow: true` (or the ellipsis on line 2), leaving the overflow state implicit and untested.

Suggested addition:

```ts
expect(result.didOverflow).toBe(true);
expect(result.lines[1].endsWith("...")).toBe(true);
```

This makes the test document the full contract of the greedy-pack + overflow path rather than accidentally passing for the wrong reason.

---

e2e/analog-clock.spec.ts
```diff
@@ -190,6 +190,36 @@ test.describe("Analog Clock - Radial Display", () => {
...
+      // across two textPaths if not. In neither case does it end with the
+      // old hard-coded "..." truncation marker.
+      const familyGameTitle = page.getByTestId("event-title-d1");
+      await expect(familyGameTitle).toBeVisible();
+      const text = await familyGameTitle.textContent();
+      expect(text).toContain("Family Game");
+      expect(text).toContain("Night");
+
+      await page.screenshot({
+        path: "test-results/screenshots/clock-default-titles.png",
```
BenSeymourODB, line 217: **[MINOR] Screenshot written to `test-results/` — confirm this path is gitignored.**

```ts
await page.screenshot({
  path: "test-results/screenshots/clock-default-titles.png",
  fullPage: true,
});
```

CLAUDE.md explicitly states: _"Never commit test output artifacts (`test-results/`, `playwright-report/`, `blob-report/`)"_. As long as `test-results/` is listed in `.gitignore` this is fine at runtime, but worth verifying the pattern covers `test-results/screenshots/` specifically (a `test-results/` entry with no trailing `/**` can sometimes miss nested directories depending on the gitignore implementation).

All other existing screenshot calls in this file use the same `test-results/screenshots/` prefix and presumably already pass CI, so this is consistent — just confirming the new entry follows the established pattern and won't accidentally get staged.

---

BenSeymourODB (commented):
**Verdict: Changes requested.** Two bugs need fixing before merge; three minor issues are lower priority but worth addressing.

### Must fix before merge

1. **`lineOffset / 2` geometry bug** (`event-arc.tsx` line 104) — the two textPath centres are only `0.55 × fontSize` apart, causing the two text bands to overlap by `~0.45 × fontSize` (5–8 px). The issue spec says `titleRadius ± lineOffset` (not `÷ 2`). Change `lineOffset / 2` → `lineOffset` in both radii. You may also need to reduce `titleFontSize` slightly in 2-line mode so the outer line doesn't clip the arc edge at large font sizes.

2. **`truncateWithEllipsis` drops `"..."` at `budget ≤ 3`** (`fit-title.ts` line 42) — returns a raw character slice with no ellipsis marker, breaking the visual contract and any `.endsWith("...")` detection downstream (including the planned #311 floating-label consumer).

### Address before merge (correctness risk)

3. **`words[nextIndex]` accessed without bounds guard** (`fit-title.ts` line 136) — the current flow makes this safe, but the code is fragile; add an explicit `nextIndex < words.length` guard and a test for the `line2 === ""` overflow branch.

### Nice to have / low priority

4. **`event-arc.test.tsx` line 71** — exact split `["Family Game", "Night"]` is coupled to layout heuristic constants; align to the same `lines.join(" ")` pattern used in `fit-title.test.ts` for robustness.

5. **`fit-title.test.ts` line 53** — greedy-pack test does not assert `didOverflow: true` or `lines[1].endsWith("...")`, leaving the overflow state implicit.

6. **E2E screenshot path** (`e2e/analog-clock.spec.ts` line 217) — consistent with existing tests but confirm `test-results/**` is fully covered by `.gitignore`.

---

src/components/calendar/analog-clock/event-arc.tsx
BenSeymourODB: Fixed in 2509284. Changed `lineOffset / 2` → `lineOffset` so the centre-to-centre distance is 2 × `lineOffset` ≈ 1.1 × fontSize.

On the related "you may also want to reduce `titleFontSize` slightly when 2-line mode is active" — I worked through the geometry at default sizes:
- arcThickness=40 (default size=600), arcHeight=40, fontSize = min(40 × 0.3, 18) = 12, titleRadius = 200 + 40 × 0.68 = 227.2
- Outer line centre = 227.2 + 6.6 = 233.8 → top of glyph at 239.8 (just inside outerRadius=240)
- Inner line centre = 220.6 → bottom at 214.6 (well inside innerRadius=200)

So the outer line lands ~0.2 px inside the arc edge at 600 px clock size, and there is more headroom at 720 px (the production clock view). The fontSize cap of 18 only kicks in at arcThickness ≥ 60, and at that point arcHeight = 60 leaves ~10 px of additional band on each side. I've left fontSize unchanged for now and rely on the screenshot artefact for visual confirmation. If we ever push fontSize larger or shrink arcThickness, we can revisit with a band-clamping pass.

---

src/components/calendar/analog-clock/fit-title.ts
BenSeymourODB: Fixed in 2509284. `truncateWithEllipsis` now always emits a visible marker, using the Unicode horizontal ellipsis (U+2026) at tight budgets (1–3 chars) so the marker fits in 1 char without consuming the entire budget:

```ts
if (budget <= 3) {
  if (budget === 1) return ELLIPSIS_CHAR;       // "…"
  return `${text.slice(0, budget - 1)}${ELLIPSIS_CHAR}`;
}
return `${text.slice(0, budget - 3)}...`;
```

I also added a separate `appendOverflowMarker` for the dedicated case where the packed words of line 2 happen to fit budget exactly but more content was cut — that path now always renders an ellipsis (previously it returned the unchanged line, which was the source of the implicit-overflow bug your other comment caught on the greedy-pack test).

3 new unit tests cover budgets 1, 2, and 3, plus a fits-exactly-3-chars no-truncation case.

---

src/components/calendar/analog-clock/fit-title.ts
BenSeymourODB: Fixed in 2509284. Added the explicit guard you suggested:

```ts
let truncatedLine2: string;
if (line2.length > 0) {
  truncatedLine2 = appendOverflowMarker(line2, budget);
} else if (nextIndex < words.length) {
  truncatedLine2 = truncateWithEllipsis(words[nextIndex], budget);
} else {
  truncatedLine2 = "";
}
```

Also added a dedicated unit test for the `line2 === ""` branch — `"Hi supercalifragilistic"` at 30°/200/14 — `"Hi"` packs onto line 1 (2 chars), the long word alone exceeds budget=12, so packLine returns an empty `line2` and the fallback truncates `words[nextIndex]` to `"superc..."` (or similar). Asserts `line1 === "Hi"`, `line2.startsWith("super")`, `line2.endsWith("...")`, and `didOverflow: true`.

---

src/components/calendar/analog-clock/event-arc.tsx
BenSeymourODB: Refactored in 2509284. The 2-line render now uses one `<text>` per `<textPath>`, wrapped in a `<g data-testid="event-title-{id}">`:

```tsx
<g data-testid={`event-title-${id}`}>
  {fit.lines.map((line, i) => (
    <text key={textPathIds[i]} fontSize={titleFontSize} ...>
      <textPath href={`#${textPathIds[i]}`} startOffset="50%" textAnchor="middle" dominantBaseline="central">
        {line}
      </textPath>
    </text>
  ))}
</g>
```

This eliminates the SVG 2 "continue on the next path in sequence" semantic concern and gives each line its own typographic context. Existing test selectors (`getByTestId("event-title-{id}")` + `.querySelectorAll("textPath")`) still work because the wrapper `<g>` carries the same testid.

Real-browser visual confirmation still depends on the screenshot artefact in `e2e/analog-clock.spec.ts`; the unit tests cannot detect glyph misalignment in JSDOM. With this PR's geometry fix (`titleRadius ± lineOffset`, no `/2`), the screenshot path is the right place to catch any remaining visual issue.

---

src/components/calendar/analog-clock/__tests__/event-arc.test.tsx
BenSeymourODB: Adopted Option 2 in 2509284. The component test is now decoupled from the layout heuristic constants:

```ts
const renderedLines = Array.from(textPaths).map((tp) => tp.textContent ?? "");
expect(renderedLines.join(" ")).toBe("Family Game Night");
expect(renderedLines.every((l) => !l.includes("..."))).toBe(true);
expect(renderedLines.every((l) => !l.includes("…"))).toBe(true);
```

Asserts the contract (two textPaths, words preserved across the split, no mid-word breaks, no truncation marker on either line) without pinning the exact split point. If the layout constants `arcHeight * 0.68` or `arcHeight * 0.3` shift in the future, the split could move from `["Family Game", "Night"]` to `["Family", "Game Night"]` and this test still passes — only a real regression (e.g. dropping a word, or single-line ellipsis fallback firing on what should be a 2-line arc) will fail it.

---

src/components/calendar/analog-clock/__tests__/fit-title.test.ts
BenSeymourODB: Fixed in 2509284. Added the two missing assertions:

```ts
expect(result.didOverflow).toBe(true);
expect(result.lines[1].endsWith("...")).toBe(true);
```

You were exactly right — the `/^groceries/` match was masking the implicit-overflow state. The new assertion exposed a real bug along the way: my v1 of `truncateWithEllipsis` was returning `line2` unchanged when its content fit budget exactly (which is the case here — `"groceries"` = 9 chars, budget = 12). So I split the helper into `truncateWithEllipsis` (only ellipsizes when text > budget) and a new `appendOverflowMarker` (always emits a marker — used in the overflow-line case where more content was cut). Both paths now correctly assert the ellipsis indicator.

---

e2e/analog-clock.spec.ts
BenSeymourODB: Confirmed — `.gitignore` lines 15–16 already cover this:

```
/playwright-report
/test-results
```

Git's `.gitignore` semantics treat a leading-slash directory entry as the directory and everything inside it (recursively), so `test-results/screenshots/clock-default-titles.png` is ignored without needing a trailing `/**`. I checked locally with `git check-ignore -v test-results/screenshots/clock-default-titles.png` and it matches the `/test-results` rule. The new entry follows the same pattern as the existing screenshot calls in this file, all of which already pass CI. No changes needed.

---

BenSeymourODB:
All 7 review findings addressed in 2509284. Per-thread replies posted with details; summary:

| # | Severity | Finding | Resolution |
|---|---|---|---|
| 1 | Critical | `lineOffset/2` overlap | Dropped `/2`; centre-to-centre = 2 × `lineOffset` ≈ 1.1 × fontSize |
| 2 | Bug | `truncateWithEllipsis` no marker at budget ≤ 3 | Always emits a marker; uses Unicode `…` (U+2026) at tight budgets |
| 3 | Bug | `words[nextIndex]` bounds | Explicit `nextIndex < words.length` guard + dedicated unit test for empty-`line2` branch |
| 4 | Minor | Multiple textPaths in one `<text>` | Refactored to one `<text>` per textPath, wrapped in `<g data-testid>` |
| 5 | Minor | Component test exact-split coupling | Relaxed to `lines.join(" ") === full title` + no-mid-word-split assertions |
| 6 | Minor | Greedy-pack test missing didOverflow | Added `didOverflow: true` + line-2 ellipsis assertions; this caught a real bug fixed by the new `appendOverflowMarker` helper |
| 7 | Minor | E2E screenshot path / .gitignore | Confirmed `/test-results` rule covers nested paths; no change needed |

**5 new unit tests** added (tight-budget paths + line2-empty branch). Full suite: **1913/1913 green**, lint + types clean.

Branch ready for re-review.

---

PR #317, opened 2026-05-05
feat(calendar): user-configurable working-hours start hour (#288)

BenSeymourODB:
Closes #288.

## Summary

The Day and Week time grids auto-scroll to a hard-coded `WORKING_HOURS_START_HOUR = 7` on first render. Early-shift workers, night-owl households, etc. can't change it. This PR makes the start hour a per-user setting (0–23, default 7) persisted in `UserSettings` and threaded into the grid scroll logic + Settings UI.

## Phases (all delivered)

1. **Schema + API + hook** (`5e1780c`)
   - `UserSettings.calendarWorkingHoursStart` (Int, default 7)
   - Migration `0006_add_calendar_working_hours_start`
   - `PUT /api/settings` validates integer 0–23
   - `useUserSettings` exposes the field
2. **Provider + grids** (`f36a79b`)
   - `ICalendarContext.workingHoursStart`, derived from user settings (clamped 0–23)
   - `MockCalendarProvider` accepts a matching prop (default 7)
   - `DayCalendar` / `WeekCalendar` compute `scrollTop` from context via `getInitialScrollTop` instead of the constant; effect remains mount-only on purpose so a user mid-scroll isn't yanked when the setting changes
   - All 14 mock-context test factories add the new field for type-safety
3. **Settings UI** (`b124d35`)
   - New `Working Hours Start` slider (0–23, step 1) in `CalendarSection` with `HH:00` label
   - `SettingsForm`, `/settings`, `/test/settings` thread the new field
   - Playwright spec updated for aria-range coverage + ArrowRight-then-PUT assertion

## Acceptance criteria mapping

- [x] Setting persists in Postgres and is editable via API + UI
- [x] Day/Week time-grid scrolls to the user's chosen hour on first render
- [x] Default value preserved for existing users (column `NOT NULL DEFAULT 7`)
- [x] Tests cover non-default values (early-shift `5`, night-owl `14`, boundaries `0`/`23`)

## Test plan

- [x] `pnpm test` — 1903/1903 pass
- [x] `pnpm db:migrate:check-names` — passes
- [ ] `pnpm test:e2e e2e/settings-calendar.spec.ts` — runs on CI; browser binaries not available in this sandbox

## Notes

- `WORKING_HOURS_START_HOUR` is no longer imported by either grid, but the constant is kept in `calendar-helpers.ts` as a fallback for any non-context caller (mirrors the `WEEK_STARTS_ON` pattern).
- The scroll effect deliberately stays mount-only. Subsequent setting changes apply on the next remount (agenda toggle, route remount). Reactively re-scrolling on every `workingHoursStart` change would yank the user's scroll position.

https://claude.ai/code/session_01WveW4UJM2PuknpxTU3esku


src/hooks/useUserSettings.ts
```diff
@@ -88,5 +91,13 @@ function pickCalendarFields(
   if (typeof data.calendarMaxEventsPerDay === "number") {
     picked.calendarMaxEventsPerDay = data.calendarMaxEventsPerDay;
   }
+  if (
```
BenSeymourODB, line 101: **should-fix · correctness**

`pickCalendarFields` only checks `typeof data.calendarWorkingHoursStart === "number"` before passing the value through to state. A float or out-of-range number that comes from the DB (e.g. a row written before migration or a corrupted payload) will pass this check and end up in `settings.calendarWorkingHoursStart` — to be accepted verbatim — even though the provider later clamps with `Math.trunc`. The clamping in `CalendarProvider` handles the render-time case, but the hook's own `settings` object is what gets passed directly to the upsert body in `PUT /api/settings` via `SettingsForm`. If the user opens Settings while the cached value is, say, `7.5`, saves without touching the slider, the form will PUT `{ calendarWorkingHoursStart: 7.5 }` and the API will correctly reject it with 400 — which will silently fail (no toast visible in the diff for this path). The fix is to add the same `Number.isInteger()` + range check inside `pickCalendarFields`, or to clamp at the hook layer rather than only at the provider layer. The test `"ignores out-of-range calendarWorkingHoursStart from a malformed response"` currently only covers the string case; it should also cover `7.5` and `-1`.

---

src/components/calendar/DayCalendar.tsx
```diff
@@ -258,15 +256,24 @@ function DayGridView({
...
-  // synchronously before paint, avoiding a visible flash from 00:00.
+  // useLayoutEffect runs synchronously before paint, avoiding a visible
+  // flash from scrollTop=0 to the working-hours row.
   useLayoutEffect(() => {
     if (scrollContainerRef.current) {
-      scrollContainerRef.current.scrollTop = INITIAL_SCROLL_TOP_PX;
+      scrollContainerRef.current.scrollTop = getInitialScrollTop(
+        workingHoursStartRef.current,
+        HOUR_HEIGHT_PX
+      );
```
BenSeymourODB, line 276: **should-fix · correctness**

```tsx
}, []);
```

Suppressing the `exhaustive-deps` warning here is the correct *intent*, but it creates a correctness gap: `workingHoursStart` is captured at mount time from the closure. If the parent somehow passes a new value before the component remounts (e.g. React re-uses the instance across a fast settings save → fast navigation cycle without agenda toggle), the effect will silently scroll to the stale value. The PR description acknowledges this and calls it intentional. But the better approach — consistent with how the pre-existing constant worked — is to read the value into a `ref` at mount time and use the ref inside the effect. That avoids the eslint suppress, is React Compiler-compatible, and makes the "capture once" intent explicit:

```tsx
const workingHoursStartRef = useRef(workingHoursStart);
useLayoutEffect(() => {
  if (scrollContainerRef.current) {
    scrollContainerRef.current.scrollTop = getInitialScrollTop(
      workingHoursStartRef.current,
      HOUR_HEIGHT_PX
    );
  }
}, []);
```

Same applies to `WeekCalendar.tsx` line 268.

---

src/components/calendar/__tests__/DayCalendar.test.tsx
```diff
@@ -386,6 +387,70 @@ describe("DayCalendar", () => {
...
+    it("honours a non-default workingHoursStart from context (night owl)", () => {
+      renderWithContext({
+        selectedDate: new Date(),
+        workingHoursStart: 14,
+      });
+      const grid = screen.getByTestId("day-calendar-grid");
+      expect(grid.scrollTop).toBe(14 * 48);
+    });
+
+    it("accepts workingHoursStart at the 0 and 23 boundaries", () => {
```
BenSeymourODB, line 413: **should-fix · tests**

The test called `"clamps workingHoursStart at the 0/23 boundaries"` passes `0` and `23` directly to `DayGridView` via mock context — these are already valid in-range values, so the test doesn't verify clamping at all. The clamping logic (`Math.min(23, Math.max(0, Math.trunc(...)))`) lives in `CalendarProvider`, not in `DayGridView`. A test that actually exercises clamping should either: (a) test `CalendarProvider` directly with out-of-range `calendarWorkingHoursStart` values (e.g. `-1`, `24`, `7.9`), or (b) be renamed to reflect what it actually tests (`"accepts boundary values 0 and 23"`). The same gap exists in `WeekCalendar.test.tsx` — boundary tests exist, but clamping coverage in the provider is absent. There are no unit tests for the provider's clamp expression.

---

src/components/settings/calendar-section.tsx
```diff
@@ -123,6 +128,33 @@ export function CalendarSection({ values, onChange }: CalendarSectionProps) {
...
+            id="calendar-working-hours-start"
+            value={[values.calendarWorkingHoursStart]}
+            min={0}
+            max={23}
+            step={1}
+            onValueChange={(value) =>
+              onChange({ calendarWorkingHoursStart: value[0] })
+            }
+          />
+          <p className="text-xs text-gray-400">
```
BenSeymourODB, line 152: **nit · architecture**

The description copy reads:

> `Hour the Day & Week views scroll to on first render so your working-hours events land in view (0:00 – 23:00)`

The parenthetical range `(0:00 – 23:00)` is already communicated visually by the label `HH:00` and the slider endpoints. More useful copy would set user expectations about the deferred-apply behavior: *"Takes effect the next time the Day or Week grid loads."* This avoids a support question from users who move the slider and then wonder why the grid didn't immediately re-scroll.

---

src/lib/calendar-helpers.ts
```diff
@@ -47,11 +47,16 @@ const FORMAT_STRING = "MMM d, yyyy";
...
 /**
- * Hour of day the Day/Week time grids scroll to on mount so working-hours
- * events are immediately visible. (#214)
+ * Default hour of day the Day/Week time grids scroll to on mount when
+ * the user has not overridden it. (#214)
  *
  * 0 = midnight, 7 = 7am, etc. Plumbed through `getInitialScrollTop` so
  * each calendar can derive its scrollTop from its own row height.
+ *
+ * @deprecated Prefer `useCalendar().workingHoursStart` (#288). This
```
BenSeymourODB, line 56: **nit · correctness**

`WORKING_HOURS_START_HOUR = 7` is kept as a "fallback for any non-context caller" per the PR description. But both callers that previously used it (`DayCalendar`, `WeekCalendar`) have been removed. The only current use-case for this constant would be a component that reads `getInitialScrollTop(WORKING_HOURS_START_HOUR, px)` *outside* of context — which doesn't exist. If no such caller is planned, keeping the export is dead code that might confuse a future contributor into using the constant instead of context. Either add a `@deprecated` JSDoc pointing to `ICalendarContext.workingHoursStart`, or add a brief comment explaining the planned non-context caller so the intent is clear.

---

src/app/api/settings/__tests__/route.test.ts
```diff
@@ -496,6 +497,60 @@ describe("/api/settings", () => {
       expect(data.calendarMaxEventsPerDay).toBe(5);
```
BenSeymourODB, line 497: **nit · tests**

The validation test (`"validates calendarWorkingHoursStart range (0-23)"`) batches four assertions into a single `it` block — negative, too-high, float, and string-type. This is a pre-existing style that other validators in this file use, so it's consistent. But the float case (`7.5`) is the one most likely to appear from a real bug (client-side rounding) and deserves its own message assertion just like the `-1` case has `expect(dataNegative.error).toContain("calendarWorkingHoursStart")`. The other sub-cases only assert the status code. Low priority since the error message is tested via the negative case, but it means a regressed error message wouldn't be caught for the float/string/high paths.

---

BenSeymourODB (commented):
First-pass review (delegated to a Sonnet subagent via `/ultrareview`-style fallback). 7 findings — 4 should-fix, 3 nits.

All 7 are addressed in commit 2d11be0 ("address review feedback on #288"). Round-2 follow-ups are inline replies on each comment thread.

---

src/hooks/useUserSettings.ts
BenSeymourODB: Addressed in 2d11be0.

`pickCalendarFields` now requires `Number.isInteger(...)` and clamps the 0–23 range — anything else (string, float, negative, > 23, null) is dropped and the hook keeps the default 7. New parameterised test covers `"9"`, `7.5`, `-1`, `24`, and `null`.

---

src/components/calendar/DayCalendar.tsx
BenSeymourODB: Addressed in 2d11be0 for both `DayCalendar.tsx` and `WeekCalendar.tsx`.

Both grids now capture `workingHoursStart` into a `useRef` at mount time and read the ref inside `useLayoutEffect`. The `eslint-disable react-hooks/exhaustive-deps` directive is gone; the empty dep array is now exhaustive.

---

src/components/calendar/__tests__/DayCalendar.test.tsx
BenSeymourODB: Addressed in 2d11be0 by doing both:

- Renamed the misleading test to `"accepts workingHoursStart at the 0 and 23 boundaries"` (option b).
- Added a new dedicated test file `src/components/providers/__tests__/CalendarProvider.workingHours.test.tsx` that mocks `useUserSettings` to feed in `-1`, `24`, `7.9`, `-100`, `9999` and asserts the provider clamps to `0`, `23`, `7`, etc. (option a).
- Added a new test in `DayCalendar.test.tsx` that re-renders with a different `workingHoursStart` and asserts `scrollTop` is unchanged — pins the mount-only design so a future refactor that adds the value to the dep array fails visibly.

---

src/components/settings/calendar-section.tsx
BenSeymourODB: Addressed in 2d11be0. Copy now reads:

> Hour the Day & Week views scroll to on first render. Takes effect the next time the grid loads (e.g. after navigating away and back).

Drops the redundant range and sets the deferred-apply expectation.

---

src/lib/calendar-helpers.ts
BenSeymourODB: Addressed in 2d11be0. Added a `@deprecated` JSDoc that points at `useCalendar().workingHoursStart` and explains the constant is retained as the default backing for the Prisma column + `DEFAULT_USER_CALENDAR_SETTINGS`. New components must read from context.

---

src/app/api/settings/__tests__/route.test.ts
BenSeymourODB: Addressed in 2d11be0. The validation block is now an `it.each` table with one row per case — `"negative"` (`-1`), `"above range"` (`24`), `"fractional"` (`7.5`), `"string"` (`"7"`) — and each row asserts both the 400 status and that `data.error` contains `"calendarWorkingHoursStart"`.

---

PR #318, opened 2026-05-05
feat(calendar): expand filter panel — persistence, per-calendar filter, hidden-events badge (#208)

### LINKED ISSUE #208: feat(calendar): expand filter panel — persistence, per-calendar filter, hidden-events badge

BenSeymourODB:
## Summary

Closes #208 — expands the calendar filter panel with all three deferred items from PR #148:

1. **Per-profile localStorage persistence** for filter selections so reloading `/calendar` keeps a user's filters, scoped per active profile so each family member sees their own state.
2. **Per-calendar filter** — a third popover lists each of the user's Google Calendars; selecting a subset narrows the event view, intersected with the existing color + user filters.
3. **Hidden-events badges** — each filter trigger shows a "{n} hidden" chip when its dimension is hiding events from view.

### Phase 1 — Filter persistence (commit 346f062)
- New `src/lib/calendar-filter-storage.ts` with pure load/save/clear helpers, validation against the `TEventColor` allowlist, `default` bucket fallback for the unauthenticated / no-profile case.
- `CalendarProvider` reads `activeProfileId` from the same localStorage key `ProfileProvider` writes (decoupled — no shared imports), hydrates `selectedColors` / `selectedUserId` on mount, re-hydrates when a `storage` event signals a cross-tab profile switch. Persistence happens only on user-driven filter changes.

### Phase 2 — Per-calendar filter (commit 292434e)
- New `ICalendarInfo` type. `fetchCalendarList` now returns the rich metadata shape instead of bare ids.
- Provider exposes `calendars`, `selectedCalendarIds`, `filterEventsBySelectedCalendars`. Filter pipeline intersects color + user + calendar dimensions. `clearFilter` resets all three.
- `CalendarFilterPanel` adds a third "Calendars" popover with active-count badge, color swatches sourced from each calendar's Google-side background color, and an empty-state message.
- 11 test mock contexts backfilled, including AgendaList's previously-missing `deleteEvent` (a pre-existing TS error that was failing CI).

### Phase 3 — Hidden-events badges (commit 4b3cc4a)
- New `src/lib/calendar-filter-counts.ts` with a pure `computeHiddenEventCounts(events, filters)` helper. Hidden count for dimension D = events that pass every other dimension's filter but fail D's; necessarily 0 when D has no active filter.
- Provider exposes `hiddenEventCounts: { color, user, calendar }`. `MockCalendarProvider` mirrors with `useMemo`.
- Each filter trigger renders a muted outline-variant `HiddenChip` when its hidden count > 0 — visually distinct from the filled "N selected" badge so users don't conflate the two.

## Tests

- 16 unit tests for `calendar-filter-storage` (defaults, profile isolation, validation, fallbacks).
- 9 unit tests for `calendar-filter-counts` (zero defaults, single + multi-dimension hidden counts, intersection semantics).
- 14 new `CalendarProvider` tests (hydration, persistence, profile isolation, mid-session re-hydration, calendar filter end-to-end).
- 13 new `CalendarFilterPanel` tests (calendar popover affordance, hidden chip per dimension).
- Full local suite: 1624 passing, 0 failing. Lint + type-check clean.

## Test plan

- [x] `pnpm test` (1624 pass)
- [x] `pnpm lint:fix` (0 errors, only pre-existing warnings)
- [x] `pnpm format:fix`
- [x] `pnpm check-types`
- [ ] Manual smoke on `/calendar`: set color/user/calendar filters → reload → all restored
- [ ] Manual smoke: switch profile from another tab → calendar filter state changes to that profile's saved state
- [ ] Manual smoke: with a multi-calendar Google account, deselect one calendar → events from that calendar disappear → "1 hidden" chip appears on the Calendars trigger

## Out of scope / deferred

- Server-side fetch URL doesn't change — per-calendar filter is purely client-side (events still fetch all `calendarIds`, then filtered in-memory). Keeps the slice small and avoids a second round-trip on toggle.
- `prefers-reduced-motion`-aware enter/exit animations on the new popover.
- Saved filter presets (named filter combinations).

Closes #208

https://claude.ai/code/session_01Nt49cqqacKqzcdYbHtimtx


src/components/providers/CalendarProvider.tsx
```diff
@@ -250,19 +331,52 @@ export function CalendarProvider({
...
+    selectedColors?: TEventColor[];
+    selectedUserId?: IUser["id"] | "all";
+    selectedCalendarIds?: string[];
+  }) => {
+    saveFilterState(activeProfileId, {
+      selectedColors: updates.selectedColors ?? selectedColors,
+      selectedUserId: updates.selectedUserId ?? selectedUserId,
+      selectedCalendarIds: updates.selectedCalendarIds ?? selectedCalendarIds,
+    });
+  };
```
BenSeymourODB: 🟡 **Stale-closure risk in `persistFilters`**

`persistFilters` is a plain function recreated on every render, and it closes over `selectedColors`, `selectedUserId`, and `selectedCalendarIds` from the enclosing render scope. Each caller always passes the changed dimension explicitly as `updates.*`, so the *other two* dimensions are read from the closure. That's fine **as long as** the render that created the closure is current — and usually it is, because React state setters are synchronous dispatches.

However, there is one concrete gap: the `storage` event handler (line 246) calls `setActiveProfileId`, which triggers a re-render and the re-hydration effect at line 271. That effect calls three separate state setters (`setSelectedColors`, `setSelectedUserIdState`, `setSelectedCalendarIdsState`). React batches these in React 18+, so the next render will see all three updated — but if the user somehow fires a filter toggle *between* the storage event and the next batch commit (practically impossible in a wall-display context, but possible in tests with `act()`), `persistFilters` would write partially-stale fallback values.

The robust fix is to pass all three current values explicitly when writing, or to flip `persistFilters` to read from refs rather than closure state. A lower-effort option is a comment acknowledging the assumption that callers always provide the dimension they changed.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -209,7 +286,11 @@ export function CalendarProvider({
       return loadColorMappings();
     }
   );
-  const [calendarIds, setCalendarIds] = useState<string[]>([]);
+  const [calendars, setCalendars] = useState<ICalendarInfo[]>([]);
+  // Derived: the id list used by the events fetch URL. Kept as a separate
+  // memoized array to avoid re-running effects keyed on `calendars` whose
+  // metadata (summary, backgroundColor) didn't actually change.
+  const calendarIds = calendars.map((c) => c.id);
```
BenSeymourODB: 🟡 **`calendarIds` creates a new array reference on every render, defeating its own stated purpose**

The comment directly above says this array is kept separate so effects keyed on `calendars` don't re-fire when only metadata (summary, backgroundColor) changed. But `calendarIds` is produced by `calendars.map(...)` inline during render — every render gets a brand-new array, so its reference identity *always* changes.

`loadEventsForDate` (line 733) and `refreshEvents` (line 629) both list `calendarIds` as a dependency. With the React Compiler enabled, the compiler would normally memoize this, **but** the Compiler is designed to avoid wrapping things that already have stable references. Since `calendarIds` is derived inline from `calendars` state, the Compiler has to treat it as unstable and will likely wrap `refreshEvents`/`loadEventsForDate` with a dependency on every calendars-metadata change anyway, making the optimization a no-op.

The correct fix under the React Compiler model is to derive `calendarIds` in a `useMemo` … except that `useMemo` is banned by project rules. The idiomatic React-Compiler solution is to pull the map into a stable computation that the compiler can reason about — in practice, since `calendars` is already a stable state value, the compiler *should* be able to memoize `calendarIds` automatically if it's computed in a simple way. Worth verifying with `react-compiler-healthcheck` or a quick benchmark that the extra metadata fields don't cause spurious re-fetches in production.

---

src/components/providers/MockCalendarProvider.tsx
```diff
@@ -189,8 +216,24 @@ export function MockCalendarProvider({
...
+      filtered = filtered.filter((event) =>
+        selectedCalendarIds.includes(event.calendarId)
+      );
+    }
+
     return filtered;
-  }, [allEvents, selectedUserId, selectedColors]);
+  }, [allEvents, selectedUserId, selectedColors, selectedCalendarIds]);
+
+  const hiddenEventCounts = useMemo(
```
BenSeymourODB: 🔴 **`useMemo` violates the React Compiler rule (CLAUDE.md)**

`MockCalendarProvider` now uses `useMemo` twice — for `filteredEvents` (line ~206) and `hiddenEventCounts` (here). CLAUDE.md is explicit: **DO NOT use `useMemo`, `useCallback`, or `React.memo`** — the React Compiler handles memoization automatically.

The real `CalendarProvider` correctly uses a `useEffect` + `setState` pattern for filtering. `MockCalendarProvider` should mirror that approach, not introduce manual memoization. Even in test/mock code, the rule applies — it's in CLAUDE.md for architectural consistency, not just for production bundles.

Please replace both `useMemo` calls with the effect-based pattern used in the real provider (or at minimum remove them; the React Compiler will memoize the derived computations automatically).

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -361,29 +475,40 @@ export function CalendarProvider({
...
+   * fails or the user has no other calendars.
    */
-  const fetchCalendarList = useCallback(async (): Promise<string[]> => {
+  const fetchCalendarList = useCallback(async (): Promise<ICalendarInfo[]> => {
     try {
       const response = await fetch("/api/calendar/calendars");
       if (!response.ok) {
         logger.log("Failed to fetch calendar list, using primary only");
-        return ["primary"];
+        return [{ id: "primary", summary: "Primary", backgroundColor: "" }];
```
BenSeymourODB: 🟡 **Fallback `primary` entry confuses the calendar filter panel for unauthenticated / API-error users**

When `fetchCalendarList` fails or returns an empty list, it injects a synthetic `{ id: "primary", summary: "Primary", backgroundColor: "" }` entry. This entry flows into `context.calendars`, which is rendered by `CalendarFilterPanel` as a visible "Primary" row in the Calendars popover — even when the user is unauthenticated and no real calendar data exists.

Two problems:

1. **Visual confusion**: an unauthenticated user (e.g., on the demo wall) will see a "Calendars" popover with a "Primary" entry they can select, which will then filter to zero events (since `allEvents` is empty, filtering passes vacuously, so it won't crash — but the badge shows "1 selected" and "0 hidden", which looks broken).

2. **selectedCalendarIds hydration from localStorage could reference `"primary"` as a real calendar**: if a user stored `selectedCalendarIds: ["primary"]` when authenticated, then logs out and back in with a different account whose primary calendar ID is not `"primary"`, the fallback entry keeps the filter "active" but matching a different calendar.

Consider: only set `calendars` when the API call succeeds with real data, and handle the empty-`calendars` empty-state in the panel (the empty-state message is already implemented for `calendars.length === 0`). The fallback was correct for the old bare-id usage pattern but the filter panel is a new consumer that doesn't need it.

---

src/components/calendar/CalendarFilterPanel.tsx
```diff
@@ -56,18 +56,42 @@ function UserAvatar({ user, className }: UserAvatarProps) {
   );
 }
 
+interface HiddenChipProps {
+  count: number;
+  testId: string;
+}
+
+function HiddenChip({ count, testId }: HiddenChipProps) {
```
BenSeymourODB, line 64: 🔵 **`HiddenChip` is a private component with a `testId` prop — worth reconsidering the interface**

The component is small and only used in this one file (three call sites). Passing `testId` as a prop is unusual — it leaks test concerns into the component's public contract. Since the three call sites always use distinct, predictable IDs (`filter-panel-color-hidden`, `filter-panel-user-hidden`, `filter-panel-calendar-hidden`), the prop adds no flexibility beyond what a plain inline `Badge` would give.

If this component ever moves to `src/components/ui/` for reuse elsewhere, the `testId` prop would need rethinking anyway. As a nit: either inline the `Badge` at each call site (no abstraction needed for three lines), or keep `HiddenChip` but rename the prop to `data-testid` / use a standard HTML passthrough, so it doesn't look like a first-class test API.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -195,10 +231,51 @@ export function CalendarProvider({
...
   );
-  const [selectedColors, setSelectedColors] = useState<TEventColor[]>([]);
+
+  // Track the profile id whose filter state currently lives in `selectedColors`
+  // / `selectedUserId`, so we can detect mid-session profile changes and
+  // re-hydrate without clobbering the new profile's stored state with the
+  // previous one's values.
+  const hydratedProfileRef = useRef(activeProfileId);
+
+  useEffect(() => {
```
BenSeymourODB, line 488: 🟡 **Race window between `storage` event handler and re-hydration effect**

The flow for a cross-tab profile switch is:

1. `storage` event fires → `setActiveProfileId(e.newValue)` schedules a re-render.
2. Next render: `activeProfileId` is updated, the re-hydration effect fires (line 271), calling `setSelectedColors`, `setSelectedUserIdState`, `setSelectedCalendarIdsState`.
3. `hydratedProfileRef` is updated to the new profile id.

The gap: between step 1 and step 3, `activeProfileId` in state is already the new profile's id, but `hydratedProfileRef.current` still holds the old one. If `persistFilters` is called during this window (e.g., the user clicks a filter toggle in the same tab *while* the storage event is being processed), it will call `saveFilterState(newProfileId, ...)` using the **old profile's** in-memory filter values — writing the old profile's state to the new profile's storage key.

This is an extremely narrow window (requires a user interaction to race with a cross-tab profile switch mid-commit), but the consequence is data corruption of the new profile's stored filters. A conservative fix: move the `hydratedProfileRef.current = activeProfileId` update to happen in the state setter callbacks, or guard `persistFilters` with a check that `hydratedProfileRef.current === activeProfileId`.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -195,10 +231,51 @@ export function CalendarProvider({
...
   const [selectedDate, setSelectedDate] = useState(new Date());
-  const [selectedUserId, setSelectedUserId] = useState<IUser["id"] | "all">(
-    "all"
+
+  // Issue #208 — filter state is persisted per-profile. The active profile
+  // id is read from the same `activeProfileId` localStorage key that
+  // ProfileProvider writes, so the two providers stay decoupled (neither
+  // imports the other). We listen for `storage` events to react to
+  // cross-tab profile switches.
+  const [activeProfileId, setActiveProfileId] = useState<string | null>(() =>
```
BenSeymourODB: 🔵 **Nit: double `getActiveProfileId()` call during init**

`activeProfileId` state (line 240) and each of the three filter state initialisers (lines 257–263) all independently call `getActiveProfileId()`, reading `localStorage` four times at mount. A minor inefficiency — one call with the result stored in a `const` would be cleaner and cheaper, though in practice the localStorage read is negligibly fast.

---

src/components/calendar/__tests__/AgendaList.test.tsx
```diff
@@ -57,11 +57,16 @@ function makeContext(
...
+    calendars: [],
+    selectedCalendarIds: [],
+    filterEventsBySelectedCalendars: vi.fn(),
+    hiddenEventCounts: { color: 0, user: 0, calendar: 0 },
     users: [] as IUser[],
     events: [] as IEvent[],
     addEvent: vi.fn(),
     updateEvent: vi.fn(),
     removeEvent: vi.fn(),
+    deleteEvent: vi.fn().mockResolvedValue(undefined),
```
BenSeymourODB: 🔵 **Scope-creep fix acceptable here, but note the trade-off**

The PR description calls out that adding `deleteEvent`, `calendars`, `selectedCalendarIds`, `filterEventsBySelectedCalendars`, and `hiddenEventCounts` to the mock context in `AgendaList.test.tsx` (and ~10 other test files) was backfilling a pre-existing TS error that was blocking CI. This is a reasonable tactical decision — the TS error was already failing builds, so fixing it inline unblocks the PR.

The downside is that the diff now touches 11 test files purely for the mock-context shape update, making the PR larger and the per-file blame noisy. For future context-shape additions, consider a factory-function approach (the `createMockContext` helper in `CalendarFilterPanel.test.tsx` is already doing this well) shared across test files, so new context fields only need to be added in one place. Worth a follow-up to centralize the mock factory.

---

BenSeymourODB (commented):
## Review — PR #318: filter panel expansion (persistence, per-calendar, hidden-events badges)

The overall shape of this work is solid. The three-phase breakdown is clean, the pure helper modules (`calendar-filter-storage`, `calendar-filter-counts`) are well-structured with good validation, the test coverage is thorough (52 new tests), and the TDD discipline was followed. The `clearFilter` all-three-dimensions reset and the per-profile storage isolation both work correctly.

**One hard blocker** that needs fixing before merge, plus three should-fix issues and two nits:

### 🔴 Blocker
- `MockCalendarProvider` uses `useMemo` twice — this directly violates CLAUDE.md's React Compiler rule. See inline comment.

### 🟡 Should-fix
1. **Stale-closure risk in `persistFilters`** — reads non-updated dimensions from closure during a profile-switch window. See inline comment on line 344 of `CalendarProvider.tsx`.
2. **Race window between storage-event handler and `hydratedProfileRef`** — a concurrent filter toggle during a cross-tab profile switch could write the old profile's filter values to the new profile's storage key. See inline comment on line 271 of `CalendarProvider.tsx`.
3. **`fetchCalendarList` fallback `primary` entry misleads the filter panel for unauthenticated users** — the panel shows a selectable "Primary" calendar entry even when there are no real calendars, and a persisted `selectedCalendarIds: ["primary"]` stays "active" across auth changes. See inline comment on line 489 of `CalendarProvider.tsx`.

### 🔵 Nits
- `HiddenChip.testId` prop leaks test concerns into the component interface.
- `getActiveProfileId()` called 4× at mount — one call to a `const` would be cleaner.
- Scope-creep in AgendaList + 10 other test files for mock-context backfill is acceptable here but a shared `createMockContext` factory would prevent recurrence.
- Pre-existing `useCallback` usage in `CalendarProvider` (not introduced by this PR) should be tracked as a follow-up to remove.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in 3e4b814. Went with the guard approach: `persistFilters` now bails when `hydratedProfileRef.current !== activeProfileId`, so a same-tick toggle racing the storage event is dropped instead of writing partially-stale values to the new profile's key. New regression test (`does not persist old-profile filter values to the new profile during an in-flight switch`) covers it.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in 3e4b814. Dropped the top-level `calendarIds = calendars.map(...)` const entirely. The three call sites now derive `calIds` inline from `calendars`, and the two `useCallback` dep arrays changed from `calendarIds` to `calendars` (the underlying state). React Compiler can reason about `calendars` directly without the extra unstable derivation in the way.

---

src/components/providers/MockCalendarProvider.tsx
BenSeymourODB: Fixed in 3e4b814. Ended up going further than the suggested effect-based pattern: switched to plain render-time derivation for both `filteredEvents` and `hiddenEventCounts`. The Compiler memoizes these automatically when inputs are stable, and it sidesteps a separate lint rule (`no-cascading-renders`) that fired on the dual-`setState`-in-effect approach.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in 3e4b814. `fetchCalendarList` now returns `[]` on API failure / zero-calendar accounts. The events fetch URL keeps a `["primary"]` fallback inline at the call site so refresh still works, but `context.calendars` stays empty and the panel's existing empty-state message renders correctly. New regression test (`does not surface a synthetic 'primary' calendar when the API returns no calendars`) covers it.

---

src/components/calendar/CalendarFilterPanel.tsx
BenSeymourODB: Keeping `HiddenChip` with the `testId` prop for now. The three call sites all carry the same outline-Badge shape with a `{n} hidden` template, so the wrapper genuinely consolidates ~5 lines × 3 sites of duplication; inlining would just expand the trigger button JSX. The `testId` prop is the only piece that varies and stays internal to this file (the component isn't exported). If/when this moves to `src/components/ui/`, agreed it should rename to a `data-testid` passthrough — happy to do that whenever it gets reused.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in 3e4b814 by the same guard added for the `persistFilters` stale-closure thread above. `persistFilters` now early-returns when `hydratedProfileRef.current !== activeProfileId`, so any user-driven write that lands in the gap between `setActiveProfileId` committing and the re-hydration effect running is dropped instead of corrupting the new profile's storage key. Regression test added.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in 3e4b814. The three filter-state initialisers now share a single `initialFilterState = loadFilterState(activeProfileId)` const at the top of the component body. Down from 4 localStorage round-trips at mount to 2 (one for `activeProfileId`, one for the filter payload).

---

src/components/calendar/__tests__/AgendaList.test.tsx
BenSeymourODB: Acknowledged — agree the shared `createMockContext` factory is the right shape going forward. Tracking as a follow-up rather than scope-creeping further into this PR. The `CalendarFilterPanel.test.tsx` factory is the model to extract.

---

PR #321, opened 2026-05-06
test(services): migrate Prisma mocks to vi.mocked() typed helpers (#262)

BenSeymourODB:
## Summary

Closes #262.

Replaces the `prisma as unknown as { ... }` double-cast pattern in all four service tests under `src/lib/services/__tests__/` with `vi.mocked(prisma, true)`. The new pattern type-checks `mockResolvedValue` / `mockRejectedValue` arguments against the real Prisma return types instead of silently widening them — exactly the regression-trap the issue calls out.

## What changed

| File | Change |
| --- | --- |
| `reward-points.test.ts` | Cast removed; adds a `makeProfileRewardPoints` factory so the duplicate-award test mock satisfies the full `ProfileRewardPoints` row shape (the production code does `select: { totalPoints: true }`, but the broad `findUnique` overload picked up by the typed mock requires the full row). |
| `streak.test.ts` | Straight cast removal. |
| `task-assignments.test.ts` | Mock rows are typed via `Prisma.TaskAssignmentGetPayload<{ include: { profile: { select: { … } } } }>` so the include/select narrowing in `getTaskAssignmentsByTaskIds` stays in lockstep with the mock fixtures. Schema drift on either side now fails the type-check. |
| `admin-verification.test.ts` | Straight cast removal. |
| `src/app/api/profiles/__tests__/fixtures.ts` | Widens local `Profile.avatar` from `unknown` to `Prisma.JsonValue` so admin-verification's typed mock accepts the fixture without a boundary cast. |

`meal-planning.test.ts` is out of scope — that file lands with PR #193 (#167) and can adopt the new pattern when it does.

## Why a deep mock

`vi.mocked(prisma, true)` returns `MockedFunctionDeep<typeof prisma>` so every nested method (e.g. `prisma.profileRewardPoints.findUnique`) is typed as a `MockInstance` whose `mockResolvedValue` is constrained to the actual Prisma return type. The previous `as unknown as { ... }` cast skipped that check entirely — `mockResolvedValue({ totalPoints: 75 })` quietly satisfied a hand-rolled mock interface, even though `findUnique` actually returns the full `ProfileRewardPoints` model.

## Acceptance criteria

- [x] All four service test files use `vi.mocked()` for Prisma mocks
- [x] No `as unknown as` casts remain in mock setup
- [x] `pnpm test` — all 1908 tests pass
- [x] `pnpm lint:fix && pnpm format:fix` clean (only pre-existing warnings)

## Out of scope (deferred)

- Full retype of the shared `Profile` fixtures to the generated Prisma `Profile` (tracked in #280).
- API-route tests that use the same double-cast pattern (`src/app/api/**/route.test.ts`). The issue scope is service tests; routes are a separate cleanup.

https://claude.ai/code/session_01DtTFZFSqJW8R4prMjLuPpd

---
_Generated by [Claude Code](https://claude.ai/code/session_01DtTFZFSqJW8R4prMjLuPpd)_

### DIFF
--- a/src/app/api/profiles/__tests__/fixtures.ts
+++ b/src/app/api/profiles/__tests__/fixtures.ts
@@ -1,8 +1,12 @@
 /**
  * Test fixtures for Profile API tests
  */
+import type { Prisma } from "@/generated/prisma/client";
 
-// Define local types for testing (matches Prisma schema)
+// Define local types for testing (matches Prisma schema). `avatar` is
+// `Prisma.JsonValue` so fixtures pass strict type checks when handed to
+// `vi.mocked(prisma)…mockResolvedValue` calls. Full retype to the
+// generated Prisma `Profile` lives in #280.
 interface Profile {
   id: string;
   userId: string;
@@ -10,7 +14,7 @@ interface Profile {
   type: "admin" | "standard";
   ageGroup: "adult" | "teen" | "child";
   color: string;
-  avatar: unknown;
+  avatar: Prisma.JsonValue;
   pinHash: string | null;
   pinEnabled: boolean;
   failedPinAttempts: number;
--- a/src/lib/services/__tests__/admin-verification.test.ts
+++ b/src/lib/services/__tests__/admin-verification.test.ts
@@ -22,13 +22,21 @@ vi.mock("@/lib/db", () => ({
   },
 }));
 
+// `bcrypt.compare` has both a callback overload (returning void) and a
+// promise overload. `vi.mocked` resolves to the callback overload, so
+// every `mockResolvedValue(...)` call below uses `as never` to bypass
+// the void return type and stand in for the promise the production
+// code awaits.
 vi.mock("bcrypt", () => ({
   default: {
     compare: vi.fn(),
   },
 }));
 
-// Typed deep mock: see reward-points.test.ts for rationale.
+// Typed deep mock: every method on `prisma` is widened to a
+// MockedFunctionDeep so `mockResolvedValue` is type-checked against the
+// real Prisma return types instead of being silently widened by an
+// `as unknown as { ... }` cast.
 const mockPrisma = vi.mocked(prisma, true);
 
 const adminWithHash = {
--- a/src/lib/services/__tests__/admin-verification.test.ts
+++ b/src/lib/services/__tests__/admin-verification.test.ts
@@ -28,11 +28,8 @@ vi.mock("bcrypt", () => ({
   },
 }));
 
-const mockPrisma = prisma as unknown as {
-  profile: {
-    findFirst: ReturnType<typeof vi.fn>;
-  };
-};
+// Typed deep mock: see reward-points.test.ts for rationale.
+const mockPrisma = vi.mocked(prisma, true);
 
 const adminWithHash = {
   ...mockAdminProfile,
--- a/src/lib/services/__tests__/reward-points.test.ts
+++ b/src/lib/services/__tests__/reward-points.test.ts
@@ -4,6 +4,7 @@
  * Tests the transaction logic in isolation with a single mock (prisma).
  * No auth, no NextRequest, no fetch.
  */
+import type { ProfileRewardPoints } from "@/generated/prisma/client";
 import { prisma } from "@/lib/db";
 import { mockTransaction } from "@/lib/test-utils/prisma-test-helpers";
 import { beforeEach, describe, expect, it, vi } from "vitest";
@@ -18,12 +19,26 @@ vi.mock("@/lib/db", () => ({
   },
 }));
 
-const mockPrisma = prisma as unknown as {
-  $transaction: ReturnType<typeof vi.fn>;
-  profileRewardPoints: {
-    findUnique: ReturnType<typeof vi.fn>;
+// Typed deep mock: every property/method on `prisma` is widened to a
+// MockedFunctionDeep so `mockResolvedValue`, `mockRejectedValue`, etc.
+// are type-checked against the real Prisma return types instead of
+// being silently widened by an `as unknown as { ... }` cast.
+const mockPrisma = vi.mocked(prisma, true);
+
+function makeProfileRewardPoints(
+  overrides: Partial<ProfileRewardPoints> = {}
+): ProfileRewardPoints {
+  return {
+    id: "rp-fixture",
+    profileId: "profile-1",
+    totalPoints: 0,
+    currentStreak: 0,
+    longestStreak: 0,
+    lastActivityDate: new Date("2024-01-01"),
+    updatedAt: new Date("2024-01-01"),
+    ...overrides,
   };
-};
+}
 
 function makeUniqueConstraintError() {
   const error = new Error(
@@ -214,9 +229,9 @@ describe("recordPointAward", () => {
 
   it("returns alreadyAwarded with current total when the unique index rejects a duplicate task award", async () => {
     mockPrisma.$transaction.mockRejectedValue(makeUniqueConstraintError());
-    mockPrisma.profileRewardPoints.findUnique.mockResolvedValue({
-      totalPoints: 75,
-    });
+    mockPrisma.profileRewardPoints.findUnique.mockResolvedValue(
+      makeProfileRewardPoints({ totalPoints: 75 })
+    );
 
     const result = await recordPointAward({
       profileId: "profile-1",
--- a/src/lib/services/__tests__/streak.test.ts
+++ b/src/lib/services/__tests__/streak.test.ts
@@ -20,9 +20,8 @@ vi.mock("@/lib/streak-helpers", () => ({
   calculateNewStreak: vi.fn(),
 }));
 
-const mockPrisma = prisma as unknown as {
-  $transaction: ReturnType<typeof vi.fn>;
-};
+// Typed deep mock: see reward-points.test.ts for rationale.
+const mockPrisma = vi.mocked(prisma, true);
 
 const mockCalculateNewStreak = vi.mocked(calculateNewStreak);
 
--- a/src/lib/services/__tests__/streak.test.ts
+++ b/src/lib/services/__tests__/streak.test.ts
@@ -4,6 +4,7 @@
  * Tests the streak-update logic in isolation with only two mocks
  * (prisma + streak-helpers). No auth, no NextRequest, no fetch.
  */
+import type { ProfileRewardPoints } from "@/generated/prisma/client";
 import { prisma } from "@/lib/db";
 import { calculateNewStreak } from "@/lib/streak-helpers";
 import { mockTransaction } from "@/lib/test-utils/prisma-test-helpers";
@@ -20,10 +21,31 @@ vi.mock("@/lib/streak-helpers", () => ({
   calculateNewStreak: vi.fn(),
 }));
 
-// Typed deep mock: see reward-points.test.ts for rationale.
+// Typed deep mock: every property/method on `prisma` is widened to a
+// MockedFunctionDeep so `mockResolvedValue` is type-checked against the
+// real Prisma return types instead of being silently widened by an
+// `as unknown as { ... }` cast. NOTE: `mockTransaction` below still
+// types tx-model fns as plain `vi.fn()`, so mocks passed *inside* a
+// $transaction call are not strict-typed — `makeProfileRewardPoints`
+// keeps those fixtures schema-complete by construction.
 const mockPrisma = vi.mocked(prisma, true);
 
 const mockCalculateNewStreak = vi.mocked(calculateNewStreak);
+
+function makeProfileRewardPoints(
+  overrides: Partial<ProfileRewardPoints> = {}
+): ProfileRewardPoints {
+  return {
+    id: "rp-1",
+    profileId: "profile-1",
+    totalPoints: 0,
+    currentStreak: 0,
+    longestStreak: 0,
+    lastActivityDate: new Date("2024-01-01"),
+    updatedAt: new Date("2024-01-01"),
+    ...overrides,
+  };
+}
 
 describe("updateProfileStreak", () => {
   beforeEach(() => {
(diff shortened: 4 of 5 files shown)


BenSeymourODB:
## Code Review — PR #321 (typed deep mocks migration)

Overall the migration is sound and a genuine improvement. The core goal — replacing the double-cast pattern with `vi.mocked(prisma, true)` so `mockResolvedValue` is constrained to real Prisma return types — is achieved cleanly in all four files. TypeScript passes, `meal-planning.test.ts` is correctly absent, and no `as unknown as` casts remain in mock *setup*. A few issues are worth addressing before merge.

---

### Issue 1 — `streak.test.ts`: inline `findUnique` fixtures omit `updatedAt`, defeating the typed-mock benefit

**File:** `src/lib/services/__tests__/streak.test.ts`
**Lines:** ~34–41, 64–71, 90–97, 117–124, 139–146, 196–203

The `findUnique` stubs inside `mockTransaction` are created with a plain `vi.fn()` and handed a raw object literal that is **missing `updatedAt`**, which is a non-nullable column on `ProfileRewardPoints`. Because those fns are independent locals (not on `mockPrisma` directly), TypeScript does not apply the Prisma return-type constraint to them — the typed-mock benefit disappears at this boundary.

```ts
// current — "updatedAt" is absent; passes TS only because
// the local `findUnique` is `vi.fn()`, not a typed mock method
const findUnique = vi.fn().mockResolvedValue({
  id: "rp-1",
  profileId: "profile-1",
  totalPoints: 100,
  currentStreak: 3,
  longestStreak: 5,
  lastActivityDate: new Date("2024-06-14"),
  // ← updatedAt missing
});
```

Either add `updatedAt` to each inline fixture, or — better — extract a `makeProfileRewardPoints` factory in `streak.test.ts` exactly as was done in `reward-points.test.ts`, and annotate the return type as `ProfileRewardPoints`. That gives the same schema-drift protection across both files.

---

### Issue 2 — `admin-verification.test.ts`: `bcrypt.compare` stubs still use `as never`

**File:** `src/lib/services/__tests__/admin-verification.test.ts`
**Lines:** 47, 60, 77, 141

```ts
vi.mocked(bcrypt.compare).mockResolvedValue(true as never);
vi.mocked(bcrypt.compare).mockResolvedValue(false as never);
```

`as never` is an unsafe cast of the same family as `as unknown as`. `bcrypt.compare` is typed as `Promise<boolean>` so `mockResolvedValue(true)` / `mockResolvedValue(false)` should work directly. This pattern was presumably inherited before this PR but the PR's own rationale ("no more silent widening") means these should be cleaned up in the same pass.

---

### Issue 3 — `reward-points.test.ts`: `makeProfileRewardPoints` uses a static date that may mask time-sensitive bugs

**File:** `src/lib/services/__tests__/reward-points.test.ts`
**Lines:** 31–41

```ts
return {
  id: "rp-fixture",
  profileId: "profile-1",
  totalPoints: 0,
  currentStreak: 0,
  longestStreak: 0,
  lastActivityDate: new Date("2024-01-01"),
  updatedAt: new Date("2024-01-01"),
  ...overrides,
};
```

The factory is only used once (the duplicate-award test). The `currentStreak`, `longestStreak`, and `lastActivityDate` fields aren't exercised in that test, but they are present in the full row. This is fine — the PR description acknowledges the mismatch between the broad `findUnique` overload (full row) and the production `select: { totalPoints: true }`. However, a comment noting why the extra fields are there (i.e., the typed mock requires the full row shape even though the service only reads `totalPoints`) would help future readers avoid "simplifying" the factory back to `{ totalPoints: 75 }`.

---

### Issue 4 — `task-assignments.test.ts`: `TaskAssignmentRow` type must stay in sync with the production query manually

**File:** `src/lib/services/__tests__/task-assignments.test.ts`
**Lines:** 28–39

```ts
type TaskAssignmentRow = Prisma.TaskAssignmentGetPayload<{
  include: {
    profile: {
      select: { id: true; name: true; color: true; avatar: true };
    };
  };
}>;
```

This is a good pattern, but the `select` block here is a copy of the one in `task-assignments.ts`. If a developer adds a field to the production `select` they must also update this type — there's no enforcement. Consider re-exporting a type from `task-assignments.ts` (e.g., `type TaskAssignmentRow`) so the test imports it directly and drifting is impossible. Not a blocker, but worth a follow-up issue.

---

### Issue 5 — `fixtures.ts`: local `Profile` interface is still diverged from generated Prisma type

**File:** `src/app/api/profiles/__tests__/fixtures.ts`
**Lines:** 10–25

The PR widens only `avatar` to `Prisma.JsonValue`, which unblocks the immediate type error. But the local `Profile` interface still diverges from `@/generated/prisma/client`'s `Profile` in other ways (e.g., `type` is `"admin" | "standard"` here but likely a Prisma enum in the generated client). The PR description defers a full retype to #280, which is a reasonable call — but the comment should reference the issue number inline on the interface itself (not just in the comment block above it) so future readers have direct context. Likewise, the route tests consuming this file (`route.test.ts`, `points/__tests__/route.test.ts`, `points/award/__tests__/route.test.ts`) still use `as unknown as { ... }` for their own `mockPrisma` — those remain silently unsafe. Deferred scope is acknowledged in the PR, but worth confirming a tracking issue exists.

---

### Non-issue confirmations

- `meal-planning.test.ts` correctly absent from the branch.
- `pnpm check-types` clean (verified locally).
- No `as unknown as` in mock *setup* in the four migrated files.
- `vi.mock("@/lib/db")` factory objects match the subset of Prisma methods actually called in each service.
- CLAUDE.md TDD compliance: all existing test cases are preserved, no tests removed or weakened.
- Production service call shapes vs mock shapes are aligned:
  - `reward-points.ts`: `$transaction`, `profileRewardPoints.findUnique` ✓
  - `streak.ts`: `$transaction` (wrapping `profileRewardPoints.findUnique/update/create`) ✓
  - `task-assignments.ts`: `taskAssignment.findMany` with correct `include/select` ✓
  - `admin-verification.ts`: `profile.findFirst` ✓


---

BenSeymourODB:
## Code Review — PR #321: Migrate Prisma mocks to `vi.mocked()` typed helpers

Reviewed HEAD `21116fcd3b05818cfe5690bf02fe113715bececc`. All 1908 tests pass, `check-types` is clean.

The direction is correct: `vi.mocked(prisma, true)` is strictly safer than the old double-cast pattern. However there are several issues worth addressing before merge.

---

### Issues

---

**[ISSUE 1 — Bug / Silent type hole]**
**File:** `src/lib/services/__tests__/streak.test.ts`
**Lines:** 34–41, 64–71, 90–97, 117–124, 139–146, 196–203

The inline `findUnique` mock objects passed to `vi.fn().mockResolvedValue(...)` are missing the `updatedAt` field, which is a non-nullable column on `ProfileRewardPoints` (`updatedAt DateTime @updatedAt` in `schema.prisma`). The full model shape as reflected in `makeProfileRewardPoints` (added in `reward-points.test.ts`) requires all seven fields; `streak.test.ts` only provides six.

This means the `findUnique` fixtures in this file are still **not type-safe** even under the new deep-mock regime. The objects are passed as arguments to plain `vi.fn().mockResolvedValue`, which is untyped, so the compiler accepts them silently — exactly the regression-trap the PR is trying to close.

**Recommendation:** Introduce a `makeProfileRewardPoints` factory in streak.test.ts (or move it to a shared test helper) and use it for all `findUnique` fixtures, mirroring the pattern in `reward-points.test.ts`.

Example inline (or import from a shared helper):
```ts
function makeProfileRewardPoints(overrides: Partial<ProfileRewardPoints> = {}): ProfileRewardPoints {
  return {
    id: "rp-fixture",
    profileId: "profile-1",
    totalPoints: 0,
    currentStreak: 0,
    longestStreak: 0,
    lastActivityDate: new Date("2024-01-01"),
    updatedAt: new Date("2024-01-01"),
    ...overrides,
  };
}
```

---

**[ISSUE 2 — Persistent unsafe casts]**
**File:** `src/lib/services/__tests__/admin-verification.test.ts`
**Lines:** 47, 60, 77, 141

```ts
vi.mocked(bcrypt.compare).mockResolvedValue(true as never);
```

The `as never` cast is a type suppression idiom: `bcrypt.compare` returns `Promise<boolean>`, but `mockResolvedValue` in Vitest's `MockedFunction<(...) => Promise<boolean>>` expects `boolean`, and `true` satisfies that — so the cast should be unnecessary. In fact, `as never` is the escape hatch used when the real type *doesn't* fit, which raises the question of whether the mock's return type is resolving correctly.

This pattern is pre-existing in the codebase (route tests use it too), but this PR is adding `vi.mocked` to service tests precisely to eliminate unsafe casts. The `as never` here is at least as untyped as the old `as unknown as` pattern for this specific method. Either the mock interface is wrong (worth a `// TODO` referencing the fix issue), or `true as never` should be replaceable with just `true`. Either way it deserves a comment instead of silently continuing the suppression pattern that the PR is trying to retire.

---

**[ISSUE 3 — Incomplete deep-mock benefit for transaction mocks]**
**File:** `src/lib/services/__tests__/reward-points.test.ts`, `src/lib/services/__tests__/streak.test.ts`
**Lines:** reward-points.test.ts 57–62, streak.test.ts 34–46

The `mockTransaction` helper in `prisma-test-helpers.ts` accepts `{ $transaction: MockFn }` and the transaction callback receives a plain `Record<string, Record<string, MockFn>>` — not a typed Prisma transaction client. So the `upsert`, `update`, `create` mocks passed to `mockTransaction` are untyped `vi.fn()` instances; their `mockResolvedValue` arguments aren't checked against the Prisma model types.

This is a pre-existing gap in the helper, not introduced by this PR, but the PR description says the new pattern means "schema drift on either side now fails the type-check." That claim is true only for the direct (non-transaction) calls; the transaction path remains untyped. A comment noting this residual gap (or a reference to the tracking issue) would prevent the PR from overstating the safety guarantee.

---

**[ISSUE 4 — Fixtures file: local `ProfileRewardPoints` shadow]**
**File:** `src/app/api/profiles/__tests__/fixtures.ts`
**Lines:** 27–35

The local `ProfileRewardPoints` interface at line 27 still lacks `updatedAt`, even though the Prisma model has it. This file is being updated in this PR (avatar widening), and that interface is exported/used in other tests. Leaving it out means any test that asserts on or types against `mockProfileRewardPoints` (line 101–109) is silently missing the field. This is technically a pre-existing issue, but touching the file creates an opportunity to fix it now rather than waiting for #280.

---

**[ISSUE 5 — Minor: Cross-reference comment could be a link]**
**File:** `src/lib/services/__tests__/admin-verification.test.ts` line 31, `streak.test.ts` line 23, `task-assignments.test.ts` line 21

```ts
// Typed deep mock: see reward-points.test.ts for rationale.
```

This is fine for now, but if `reward-points.test.ts` ever moves or is renamed the rationale silently disappears. Consider either inlining a one-sentence rationale in each file or pointing to a docs entry. Low priority.

---

### Confirmations ✓

- `meal-planning.test.ts` correctly absent from main — deferred to PR #193.
- `task-assignments.test.ts`: `TaskAssignmentRow` type mirrors the production `include/select` shape in `getTaskAssignmentsByTaskIds` exactly (confirmed against `task-assignments.ts` lines 47–59). Schema drift will now surface at compile time. Good.
- `admin-verification.test.ts`: `mockAdminProfile` fixture covers all non-nullable `Profile` columns. The `avatar: Prisma.JsonValue` widening (Issue #1 of the PR) is correct and necessary.
- No `as unknown as` casts remain in mock setup — PR goal achieved.
- Fixture widening (`avatar: unknown` → `Prisma.JsonValue`) does not break any other test; all 1908 pass.
- TDD compliance: all tests are behaviour tests; no tests were removed or weakened.
- CLAUDE.md compliance: no `useMemo`/`useCallback`/`React.memo`, no `any`, pnpm throughout.

---

### Summary

The PR is a genuine improvement — the `vi.mocked(prisma, true)` pattern is correct and the `TaskAssignmentRow` payload type is the highlight. The main concern (Issue 1) is that `streak.test.ts` fixture objects remain structurally incomplete relative to the real model, which is exactly the bug the PR is supposed to prevent. Issues 2 and 3 are lower-priority but worth documenting. Recommend addressing Issue 1 before merge.

---

BenSeymourODB:
Round 2 — addressed the review (`cda93c7`):

**Issue 1 (streak findUnique fixtures missing `updatedAt`)** — Fixed. Added a `makeProfileRewardPoints` factory in `streak.test.ts` mirroring the one in `reward-points.test.ts`; every `findUnique` mock now goes through it so fixtures are schema-complete by construction. As you noted, the `mockTransaction` callback typing widens these to plain `MockFn`, so the factory is the closest we can get to type-safety until the helper itself is retyped.

**Issue 2 (`as never` on `bcrypt.compare`)** — Kept the cast; documented why. `bcrypt.compare` has both a callback overload (returning `void`) and a promise overload, and `vi.mocked(...)` resolves to the callback form. Dropping `as never` produces `TS2345: Argument of type 'boolean' is not assignable to parameter of type 'void'` (verified locally). Added a comment near the `vi.mock("bcrypt", ...)` setup so the next reader doesn't have to relearn this.

**Issue 3 (overstated safety guarantee inside `mockTransaction`)** — Added a docstring on `mockPrisma` in `streak.test.ts` calling out the residual gap; the factory in Issue 1 closes the fixture-completeness side of the same problem. A full retype of `mockTransaction` is scope-creep for this PR — happy to file a follow-up if you'd like.

**Issue 4 (`ProfileRewardPoints` fixture missing `updatedAt`)** — False positive. The local fixture interface already includes `updatedAt: Date` (`src/app/api/profiles/__tests__/fixtures.ts:34` on this branch); no change needed.

**Issue 5 (cross-reference comments)** — Inlined the rationale for `vi.mocked(prisma, true)` in `admin-verification.test.ts` and `task-assignments.test.ts` so renaming `reward-points.test.ts` won't silently break the comments.

CI: https://github.com/BenSeymourODB/next-digital-wall-calendar/actions/runs/25409611713

---

PR #322, opened 2026-05-06
feat(calendar): multi-calendar selector UI in EventCreateDialog (#268)

BenSeymourODB:
Closes #268.

## Summary

`POST /api/calendar/events` has accepted `calendarId` since day one, but `EventCreateDialog` had no UI for choosing the target calendar — users with multiple writable Google calendars couldn't pick where the event lands; the server silently used `primary`.

This PR delivers the picker end-to-end:

| Phase | Commit | What it does |
| --- | --- | --- |
| 1. Data layer | `74c8fa7` | `/api/calendar/calendars` now propagates `accessRole`. New `useWritableCalendars()` hook returns `writer`/`owner` calendars, ordered primary-first then alphabetised. |
| 2. UI | `cfdb086` | `EventCreateDialog` gains an opt-in calendar `Select` and an `EventCreateInput.calendarId` field. The picker only renders when ≥ 2 writable calendars exist; with 0 or 1 the dialog silently submits with the single calendar (or `"primary"`) so single-calendar users see no UI change. Adds jsdom polyfills for Pointer Events + scrollIntoView so Radix Select can mount in component tests. |
| 3. Persistence + wiring | `1b0b552` | `AddEventButton` reads the writable list, persists the user's last-used calendar id to `localStorage["calendar.lastUsedCalendarId"]`, and threads the chosen `calendarId` into both the optimistic `IEvent` and the wire `CreateEventInput`. Stale persisted ids fall back to `primary`. |

## Acceptance criteria

- [x] Picker visible in `EventCreateDialog`, filtered to writable calendars (`accessRole` writer/owner)
- [x] Selected `calendarId` flows through to the POST payload
- [x] Last-used selection remembered across sessions
- [x] Component tests cover empty / single / multi-calendar cases

## Test plan

- [x] Route tests cover the new `accessRole` field including a missing-role → `reader` fail-closed fallback
- [x] Hook tests cover authenticated / unauthenticated, role filtering, sort order, API-error and fetch-rejection paths
- [x] Dialog tests cover empty, single-calendar (no picker), multi-calendar (picker rendered), `defaultCalendarId` happy/stale paths, and a switch-then-submit flow
- [x] AddEventButton tests cover persistence rehydration, stale-id fallback to primary, and round-trip persist on submit
- [x] Full suite: 1928 passed (was 1914; +14 new for #268)

## Notes

The "Validate Prisma migrations are up to date" CI step is failing on this branch but the failure is **not** introduced by this PR — local `prisma migrate diff` exits 0, no `prisma/` files are touched, and the symptom matches the long-standing CI flake tracked by #263.

https://claude.ai/code/session_01MmJyAf5FDDe7dG1xXPL8Vv

### DIFF
--- a/src/app/api/calendar/calendars/__tests__/route.test.ts
+++ b/src/app/api/calendar/calendars/__tests__/route.test.ts
@@ -13,7 +13,7 @@ import {
   parseResponse,
 } from "@/lib/test-utils/api-test-helpers";
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import { GET } from "../route";
+import { type CalendarsResponse, GET } from "../route";
 
 // Mock modules BEFORE imports
 vi.mock("@/lib/auth", () => ({
@@ -40,22 +40,9 @@ vi.mock("@/lib/logger", () => ({
 // Mock global fetch
 const mockFetch = vi.fn();
 global.fetch = mockFetch;
-
-// Type definitions for calendar list response
-interface CalendarInfo {
-  id: string;
-  summary: string;
-  description?: string;
-  backgroundColor: string;
-  foregroundColor: string;
-  primary: boolean;
-  selected: boolean;
-  accessRole: "freeBusyReader" | "reader" | "writer" | "owner";
-}
-
-interface CalendarsResponse {
-  calendars: CalendarInfo[];
-}
+// `CalendarsResponse` is imported from `../route` rather than re-declared
+// locally — keeps the test contract single-sourced so a future field added
+// to the route can't silently bypass the test schema.
 
 describe("/api/calendar/calendars", () => {
   beforeEach(() => {
--- a/src/app/api/calendar/calendars/__tests__/route.test.ts
+++ b/src/app/api/calendar/calendars/__tests__/route.test.ts
@@ -50,6 +50,7 @@ interface CalendarInfo {
   foregroundColor: string;
   primary: boolean;
   selected: boolean;
+  accessRole: "freeBusyReader" | "reader" | "writer" | "owner";
 }
 
 interface CalendarsResponse {
@@ -99,6 +100,7 @@ describe("/api/calendar/calendars", () => {
             foregroundColor: "#ffffff",
             primary: true,
             selected: true,
+            accessRole: "owner",
           },
           {
             id: "family@group.calendar.google.com",
@@ -108,6 +110,7 @@ describe("/api/calendar/calendars", () => {
             foregroundColor: "#ffffff",
             primary: false,
             selected: true,
+            accessRole: "writer",
           },
           {
             id: "work@group.calendar.google.com",
@@ -116,6 +119,7 @@ describe("/api/calendar/calendars", () => {
             foregroundColor: "#ffffff",
             primary: false,
             selected: false,
+            accessRole: "reader",
           },
         ],
       };
@@ -138,6 +142,7 @@ describe("/api/calendar/calendars", () => {
         foregroundColor: "#ffffff",
         primary: true,
         selected: true,
+        accessRole: "owner",
       });
       expect(data.calendars[1]).toEqual({
         id: "family@group.calendar.google.com",
@@ -147,7 +152,36 @@ describe("/api/calendar/calendars", () => {
         foregroundColor: "#ffffff",
         primary: false,
         selected: true,
+        accessRole: "writer",
       });
+      expect(data.calendars[2].accessRole).toBe("reader");
+    });
+
+    it("defaults missing accessRole to 'reader' (read-only fallback)", async () => {
+      vi.mocked(getSession).mockResolvedValue(mockSession);
+      vi.mocked(getAccessToken).mockResolvedValue(
+        mockGoogleAccount.access_token!
+      );
+
+      mockFetch.mockResolvedValue({
+        ok: true,
+        json: () =>
+          Promise.resolve({
+            items: [
+              {
+                id: "no-role@group.calendar.google.com",
+                summary: "No role",
+                // accessRole omitted — Google's schema marks it optional
+              },
+            ],
+          }),
+      });
+
+      const response = await GET();
+      const { status, data } = await parseResponse<CalendarsResponse>(response);
+
+      expect(status).toBe(200);
+      expect(data.calendars[0].accessRole).toBe("reader");
     });
 
     it("returns empty array when no calendars exist", async () => {
--- a/src/app/api/calendar/calendars/route.ts
+++ b/src/app/api/calendar/calendars/route.ts
@@ -10,6 +10,16 @@ import { NextResponse } from "next/server";
 const GOOGLE_CALENDAR_API = "https://www.googleapis.com/calendar/v3";
 
 /**
+ * Access role exposed by `calendarList.list`. Determines whether the calendar
+ * accepts writes (used by the EventCreateDialog calendar picker — issue #268).
+ */
+export type CalendarAccessRole =
+  | "freeBusyReader"
+  | "reader"
+  | "writer"
+  | "owner";
+
+/**
  * Calendar information returned by this endpoint
  */
 export interface CalendarInfo {
@@ -20,6 +30,7 @@ export interface CalendarInfo {
   foregroundColor: string;
   primary: boolean;
   selected: boolean;
+  accessRole: CalendarAccessRole;
 }
 
 /**
@@ -101,6 +112,9 @@ export async function GET() {
       foregroundColor: item.foregroundColor || "#ffffff",
       primary: item.primary || false,
       selected: item.selected || false,
+      // Default to "reader" when Google omits accessRole — fail-closed so the
+      // event-create picker never offers a calendar we can't actually write to.
+      accessRole: item.accessRole ?? "reader",
     }));
 
     logger.log("Calendar list fetched", {
--- a/src/components/calendar/AddEventButton.tsx
+++ b/src/components/calendar/AddEventButton.tsx
@@ -3,6 +3,8 @@
 import { useCalendar } from "@/components/providers/CalendarProvider";
 import { Button } from "@/components/ui/button";
 import { useEventCreate } from "@/hooks/useEventCreate";
+import { useLocalStorage } from "@/hooks/useLocalStorage";
+import { useWritableCalendars } from "@/hooks/useWritableCalendars";
 import type { IEvent, IUser } from "@/types/calendar";
 import { useState } from "react";
 import { Plus } from "lucide-react";
@@ -20,6 +22,7 @@ const LOCAL_USER: IUser = {
 };
 
 const DEFAULT_CALENDAR_ID = "primary";
+const PERSISTED_CALENDAR_KEY = "calendar.lastUsedCalendarId";
 
 function generateId(): string {
   return crypto.randomUUID();
@@ -27,12 +30,22 @@ function generateId(): string {
 
 export function AddEventButton() {
   const { selectedDate } = useCalendar();
+  const { calendars } = useWritableCalendars();
+  const [persistedCalendarId, setPersistedCalendarId] = useLocalStorage<string>(
+    PERSISTED_CALENDAR_KEY,
+    DEFAULT_CALENDAR_ID
+  );
   const createEvent = useEventCreate({
     defaultCalendarId: DEFAULT_CALENDAR_ID,
   });
   const [open, setOpen] = useState(false);
 
   const handleCreate = (input: EventCreateInput) => {
+    // Remember the user's choice for next time. The dialog already validated
+    // it against the writable list (or fell back to primary), so anything
+    // arriving here is safe to store as-is.
+    setPersistedCalendarId(input.calendarId);
+
     const optimistic: IEvent = {
       id: generateId(),
       title: input.title,
@@ -42,7 +55,7 @@ export function AddEventButton() {
       startDate: input.startDate,
       endDate: input.endDate,
       user: LOCAL_USER,
-      calendarId: DEFAULT_CALENDAR_ID,
+      calendarId: input.calendarId,
     };
 
     // Fire-and-forget against the Google write path. The hook owns the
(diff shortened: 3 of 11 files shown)


BenSeymourODB:
CI failure on the "Validate Prisma migrations are up to date" step is **not introduced by this PR**:

- This branch's diff vs `origin/main` is 4 files, none touching `prisma/`.
- Locally, `npx prisma migrate diff --from-migrations prisma/migrations --to-schema prisma/schema.prisma --exit-code` returns 0 (no drift).
- This matches the symptom tracked in #263 ("chore(ci): provision shadow database for 'Validate Prisma migrations' step") — the step is fragile in CI without a shadow DB.

Continuing with Phase 2 (picker UI) and Phase 3 (persistence + wiring).

---

BenSeymourODB (commented):
Manual deep review (fallback from `/ultrareview`). Four substantive findings below — three bugs/gaps, one question. CLAUDE.md compliance checks all clear: no manual memoization (`useMemo`/`useCallback`/`React.memo`) in the new code, strict TypeScript throughout, no `any`, standard Tailwind palette used correctly.

---

BenSeymourODB:
## Review findings (manual deep review)

### Finding 1 — `state.calendarId` stale after calendars load mid-session [correctness]

**File:** `src/components/calendar/EventCreateDialog.tsx`, line 227–234

`initialCalendarId` is computed eagerly on every render from the `calendars` prop. The `useState` initializer only runs once (on mount), so if `AddEventButton` renders before `useWritableCalendars` has resolved — which it always does, since `isLoading` starts `false` and flips to `true` only after the first effect — the dialog's `state.calendarId` is initialised to `"primary"` with an empty calendar list.

When the fetch resolves, `calendars` is updated (triggering a re-render), `initialCalendarId` now resolves correctly, but `state.calendarId` is **not updated** — it stays `"primary"` from the initialiser. The dialog's picker will then render the full list but its `value` is already `"primary"`, which happens to be correct only because `"primary"` IS in the writable list.

The failure case: a user who has no `primary` writable calendar (e.g. signed in with a GSuite account where the primary is `reader`-only). They open the dialog immediately before the fetch completes. `state.calendarId` is set to `"primary"`. The fetch returns, the picker shows the correct writable list (e.g. `"work@company.com"`), but `state.calendarId` still holds `"primary"`. Since `"primary"` is not in the list, the Radix `Select` will render with an empty/placeholder selected value, and on submit the event goes to `"primary"` — a calendar they can't write to — causing a 403 from the Google API.

**Suggested fix:** There are two clean approaches:
1. In `AddEventButton`, don't pass `calendars` to `EventCreateDialog` until `isLoading` is `false` (i.e. pass `calendars={isLoading ? [] : calendars}` and keep the dialog disabled or loading-indicated during that window). Then the dialog initializes correctly every time.
2. Add a `useEffect` (or the existing "reset on open" guard) that syncs `state.calendarId` when `calendars` changes and the current `state.calendarId` is no longer in the list.

There is currently no test that exercises `isLoading: true` → `isLoading: false` with the dialog open during the transition.

---

### Finding 2 — `useWritableCalendars` non-OK error path leaks `isLoading: true` [correctness]

**File:** `src/hooks/useWritableCalendars.ts`, lines 66–71

```ts
if (!response.ok) {
  logger.log("useWritableCalendars: API returned non-OK", {
    status: response.status,
  });
  setCalendars([]);
  return; // ← early return, bypasses `finally`
}
```

An early `return` inside an `async` IIFE **does** trigger the `finally` block (this is standard JS semantics), so `setIsLoading(false)` will be called. This is actually safe. However, it's non-obvious and is inconsistent with the `catch` path which also explicitly calls through to `finally`. Worth adding a comment to avoid future confusion, or restructure so the early-return path doesn't rely on `finally` implicitly.

The hook test for the error path (`"returns empty list when API errors out"`) does assert `isLoading: false`, which passes — so the behaviour is correct. This is a minor clarity issue only.

---

### Finding 3 — Missing test: dialog open during calendar list loading [test coverage]

**File:** `src/components/calendar/__tests__/AddEventButton.test.tsx`

The `mockCalendars()` helper always sets `isLoading: false`. There is no test that verifies what happens when the user opens the dialog while `isLoading: true` and then `isLoading` transitions to `false` with a populated calendar list. This is exactly the race described in Finding 1.

A test to add:
```ts
it("updates selected calendar when list arrives after dialog opens", async () => {
  // Start with isLoading=true, empty list
  vi.mocked(useWritableCalendars).mockReturnValue({ calendars: [], isLoading: true });
  const { ctx, rerender } = renderWithContext();

  await user.click(screen.getByRole("button", { name: /add event/i }));
  expect(screen.queryByLabelText(/^calendar/i)).not.toBeInTheDocument(); // picker hidden

  // Simulate fetch completing
  vi.mocked(useWritableCalendars).mockReturnValue({
    calendars: [
      makeCalendar({ id: "work@company.com", summary: "Work", primary: false, accessRole: "writer" }),
    ],
    isLoading: false,
  });
  rerender(...); // trigger re-render

  // With only one writable calendar, picker still hidden — but calendarId should be "work@company.com"
  await user.type(screen.getByLabelText(/title/i), "Meeting");
  await user.click(screen.getByRole("button", { name: /create event/i }));
  const [, input] = vi.mocked(ctx.createEvent).mock.calls[0];
  expect(input.calendarId).toBe("work@company.com"); // would currently be "primary"
});
```

---

### Finding 4 — `resolveInitialCalendarId` branch: empty `calendars` returns `defaultCalendarId` even if it's an arbitrary string [contract question]

**File:** `src/components/calendar/EventCreateDialog.tsx`, lines 97–99

```ts
if (calendars.length === 0) {
  return defaultCalendarId ?? FALLBACK_CALENDAR_ID;
}
```

When `calendars` is empty (loading or not yet fetched), the function returns `defaultCalendarId` — which may be any persisted string from localStorage — without validating it against any list. The comment in `AddEventButton` says "The dialog already validated it against the writable list", but that validation only happens after the list is available. During the loading window, `defaultCalendarId` passes through unchecked.

This is mostly benign today (worst case: an event goes to a stale `calendarId` that the API will reject with a 403), but it breaks the "fail-closed" contract the `accessRole: ?? "reader"` default in `route.ts` was specifically designed to enforce. The PR description calls out the fail-closed approach as intentional.

If Finding 1's fix delays passing `calendars` until `isLoading: false`, this is automatically resolved. If not, worth documenting the gap explicitly in the JSDoc — or adding a server-side validation that rejects `POST /api/calendar/events` for non-writable `calendarId`s (which is a separate concern but the real enforcement layer).

---

BenSeymourODB:
Re-posting four review findings as a plain comment because the inline review comments did not attach correctly.

---

**Finding 1 — Bug: `resolveInitialCalendarId` is called unconditionally on every render, so a mid-render calendar-list update can silently clobber the user's in-progress picker selection**

`src/components/calendar/EventCreateDialog.tsx`, around line 227–244

`resolveInitialCalendarId(calendars, defaultCalendarId)` runs on every render to produce `initialCalendarId`, and that value is then fed into `buildInitialState` whenever the dialog transitions from closed → open (the `open !== prevOpen` guard). However, `initialCalendarId` is also used as the `useState` initializer argument — and because it is re-derived each render, if the `calendars` prop finishes loading (changes from `[]` to a real list) while the dialog is already open, the next render produces a different `initialCalendarId` but the guarded `setState` call is never reached (because `open === prevOpen`). The effect is subtle but real: the selected calendar displayed to the user and the one stored in `state.calendarId` can diverge if loading completes after the dialog opens. The fix is to derive `initialCalendarId` only once (inside the `useState` initializer) or to move the resolution inside `buildInitialState` so the re-open guard fully controls when it is applied.

---

**Finding 2 — Bug: `useWritableCalendars` does not re-fetch when the session transitions from `loading` → `authenticated`; it only ever fires once**

`src/hooks/useWritableCalendars.ts`, lines 54–93

The `useEffect` depends only on `[status]`, which is correct in principle, but the initial render almost always sees `status === "loading"` (NextAuth has not resolved yet), so the effect returns early and sets nothing. When the session resolves to `"authenticated"` the status changes and the effect re-runs — that part works. The bug is that if the component unmounts and remounts while the session is already `"authenticated"` (e.g. navigating away and back), the `cancelled` flag from the previous effect run is irrelevant but a brand-new fetch is issued correctly. The actual gap is that there is no retry or re-fetch on a 4xx/5xx response: `setCalendars([])` is called and `isLoading` drops to `false`, but the hook never attempts the call again unless `status` changes. If the Google API returns a transient 500 at mount time, the calendar list stays empty for the lifetime of the component with no recovery path. Consider adding an error state that the caller can act on, or a single retry with back-off.

---

**Finding 3 — Gap: the route test's `CalendarInfo` interface is re-declared locally and has drifted — it was missing `accessRole` before this PR, and the fix only partially closes the gap**

`src/app/api/calendar/calendars/__tests__/route.test.ts`, lines 45–58

The test file re-declares its own `CalendarInfo` interface rather than importing the one exported from `../route`. Before this PR the local interface lacked `accessRole` entirely, which meant type-checking could not catch a mismatch between what the route actually returns and what the tests assert against. This PR adds `accessRole` to the local interface (fixing the immediate gap), but the duplication remains: if the canonical `CalendarInfo` in `route.ts` gains or loses a field in the future, the test interface will silently fall behind again. The tests should import `CalendarInfo` and `CalendarsResponse` directly from `../route` (both are already exported) and delete the local re-declarations to keep the type contract single-sourced.

---

**Finding 4 — Question: why does `AddEventButton` pass `DEFAULT_CALENDAR_ID` (`"primary"`) as `defaultCalendarId` to `useEventCreate` but `input.calendarId` to the optimistic event and wire call?**

`src/components/calendar/AddEventButton.tsx`, lines 38–72

`useEventCreate` is initialized with `defaultCalendarId: DEFAULT_CALENDAR_ID` (hardcoded `"primary"`), but the `handleCreate` callback passes `input.calendarId` — which comes from the dialog and may be any writable calendar — to both the optimistic `IEvent` and the `CreateEventInput` wire payload. If `useEventCreate` internally falls back to its own `defaultCalendarId` in any code path (e.g. when the hook's argument is used as a fallback rather than the passed-in input field), events submitted to a non-primary calendar could silently be written to `"primary"` instead. The intent appears to be that the wire input's `calendarId` always wins, but the hook's initialization argument being hardcoded to `"primary"` is at minimum confusing and at worst a latent bug depending on `useEventCreate`'s internal logic. Please confirm whether `useEventCreate`'s `defaultCalendarId` option is used as a true fallback or is overridden entirely by the wire input, and document that contract.


---

BenSeymourODB:
Round-2 follow-up to the review findings. All four have been addressed or answered in commit `1712b90`:

### Finding 1 — race condition (the real bug) — **fixed**
`src/components/calendar/EventCreateDialog.tsx`

Confirmed and fixed. Concrete repro: localStorage holds `"revoked@cal"` (calendar the user lost write access to), dialog mounts during the `useWritableCalendars` loading window with `calendars=[]`. `resolveInitialCalendarId([], "revoked@cal")` returned the persisted id verbatim (the empty-list path bypassed the writable-list check), seeding `state.calendarId="revoked@cal"`. When the writable list resolved after the dialog opened, the late arrival did nothing — submission then sent `"revoked@cal"` and Google returned 403.

Fix is a prev-render guard mirroring the existing `prevOpen` pattern: when the `calendars` reference changes and the current `state.calendarId` isn't in the new list (and the new list is non-empty), reconcile via `resolveInitialCalendarId`. To make the guard's reference-equality check stable, the `calendars = []` default was replaced with a module-scope `EMPTY_CALENDARS` singleton — without that, the literal would mint a fresh array every render and trip an infinite re-render. Two new tests cover both directions: stale id → reconciled to primary, still-valid id → preserved.

### Finding 2 — test re-declares `CalendarsResponse` (single-sourcing) — **fixed**
`src/app/api/calendar/calendars/__tests__/route.test.ts:16`

Right call. The local re-declaration is exactly the pattern that hid the missing-`accessRole` gap before this PR. The test now imports `CalendarsResponse` from `../route`, so any future field added to the response type lands in the test annotations automatically.

### Finding 3 — early-return paths in `useWritableCalendars` rely on `finally` — **clarified**
`src/hooks/useWritableCalendars.ts:62`

Added an inline comment naming the invariant: the non-OK and `catch` branches both `return` before `setIsLoading(false)`, and the single `finally` block owns the loading flag so the responsibility doesn't get split across the three exit paths. No behaviour change.

### Finding 4 (and the variant in the second comment) — `useEventCreate({ defaultCalendarId: "primary" })` while passing `input.calendarId` — **answered, no change**
`src/components/calendar/AddEventButton.tsx:38`

This is intentional belt-and-braces, not dead code. Looking at `useEventCreate` (`src/hooks/useEventCreate.ts`):

```ts
const created = await createEvent(optimistic, {
  ...input,
  calendarId: input.calendarId ?? defaultCalendarId,
});
```

`defaultCalendarId` is the **fallback** when `input.calendarId` is absent — the call always wins when set. After this PR `EventCreateInput.calendarId` is required, so the fallback is currently unreachable from `AddEventButton`, but:
- `useEventCreate` is exposed as a general hook (used elsewhere in the planned PATCH path per `docs/calendar-event-database.md`), so the option needs to stay.
- Keeping the explicit `defaultCalendarId: DEFAULT_CALENDAR_ID` at the call site documents the contract and means a future code path that ever loses `input.calendarId` doesn't silently flip to whatever `useEventCreate`'s internal default is. Removing it would just make that hidden coupling brittle.

The optimistic `IEvent.calendarId` and the wire `CreateEventInput.calendarId` both use `input.calendarId` (line 56 / 70) — they cannot diverge from what the dialog committed to.

### Finding 5 (from the second comment) — `useWritableCalendars` doesn't retry on transient errors — **deferred**
`src/hooks/useWritableCalendars.ts:66`

Real gap, but out of scope for this PR. The current behaviour matches the rest of the calendar fetch surface (e.g. `useUserSettings` is also single-shot) and a refresh fully reloads the hook. Tracking this with the broader cache/retry hardening in #290 / #218 rather than landing a one-off retry here.

---

Full suite green: 1930 passed (was 1928 pre-fix; +2 new race tests). `pnpm lint:fix && pnpm format:fix && pnpm check-types` all clean.

---

PR #323, opened 2026-05-06
feat(theme): Wall-Projector theme + scoped light emphasis primitive (#319)

### LINKED ISSUE #319: feat(theme): add Wall-Projector theme — dark-by-default with scoped light emphasis primitive (Clock face as first consumer)
## User story

> As an operator of a digital wall calendar mounted on a wall (or projected onto one), I want a "Wall-Projector" theme that's dark-themed by default — to reduce light pollution at night and lower power on always-on displays — but allows me to opt specific elements into light theming so they pop against the dark surround.
>
> The first concrete use case: in the Clock view, I want the clock **face** rendered with a white background (light theming) while the rest of the page (header, sidebar, all-day list, controls) stays dark. The clock face is the dominant visual element and benefits from high black-on-white contrast for at-a-glance time reading from across the room.

## Why this is a third theme, not just a checkbox in dark mode

The behavior is specific to a wall-display use case:

- "Light" theme — pure light, looks washed out from across a room.
- "Dark" theme — uniformly dark, content can disappear at glance distance, but is right for ambient nighttime use.
- **"Wall-Projector"** — dark **chrome** (header, sidebar, controls) so the room stays dim, with **scoped light islands** for the high-density-information surfaces (clock face today; potentially Year-view month thumbnails, Day-view time grid, etc., later).

Folding this into "Dark" via per-element checkboxes muddies what "dark" means. A separate theme name keeps each theme self-consistent.

### Part 1 — `wall-projector` theme

Extend `next-themes` to recognize a third theme name `wall-projector` alongside `light`, `dark`, `system`. The theme applies the `.wall-projector` class on `<html>` (via the existing `attribute="class"` config in `src/components/providers/ThemeProvider.tsx:13`).

CSS-token-wise, `.wall-projector` inherits the **dark** token set by default. New CSS rules in `src/app/globals.css` after the existing `.dark` block:

```css
.wall-projector {
  /* Default to dark token values — same as .dark */
  --background: oklch(0.145 0.016 285.823);
  --foreground: oklch(0.985 0.002 285.823);
  /* …rest of the dark block, identical for now… */
}
```

Add the third item to the dropdown in `src/components/theme/theme-toggle.tsx:35-46` ("Wall-Projector" with a projector / monitor icon).

### Part 2 — scoped light/dark primitive

A reusable component that re-declares the opposite theme's CSS variables for everything inside it, no matter what the outer theme is:

```tsx
// src/components/theme/theme-scope.tsx
type ThemeScopeProps = { mode: "light" | "dark"; children: ReactNode; className?: string };

export function ThemeScope({ mode, children, className }: ThemeScopeProps) {
  return (
    <div data-theme-scope={mode} className={className}>
      {children}
    </div>
  );
}
```

CSS:

```css
[data-theme-scope="light"] {
  /* Re-declare the :root (light) token values */
  --background: oklch(1 0 0);
  --foreground: oklch(0.145 0.016 285.823);
  /* …rest of the :root block… */
}

[data-theme-scope="dark"] {
  /* Re-declare the .dark token values */
  --background: oklch(0.145 0.016 285.823);
  /* …rest of the .dark block… */
}
```

Because every UI component in this codebase already styles via the semantic tokens (`bg-background`, `text-foreground`, `bg-card`, etc.), wrapping a subtree in `<ThemeScope mode="light">` causes every Tailwind utility inside to resolve to light values, no per-component changes needed. This is the key reason this approach is cheap: the existing `dark:` classes don't need to be rewritten, because the variables they ultimately read from are scoped.

Caveats:
- Tailwind's `dark:` variant is gated by `@custom-variant dark (&:where(.dark, .dark *))` (`globals.css:4`). Inside a `<ThemeScope mode="light">` that's nested under `.dark` or `.wall-projector`, **the `dark:` variant is still active** because the ancestor still has `.dark` / `.wall-projector`. So a class like `bg-white dark:bg-black` would resolve to black inside a light scope under a dark page — wrong. Two options:
  - **(a)** Audit and remove `dark:` overrides in components that we want to wrap with `ThemeScope`, relying purely on semantic tokens. The Clock face (`src/components/calendar/analog-clock/clock-face.tsx`) is the first audit target.
  - **(b)** Tighten the dark variant to `&:where(.dark:not([data-theme-scope]), .dark:not([data-theme-scope]) *)` style — more precise but harder to reason about.
  - Recommend **(a)** as the long-term direction. The codebase already uses semantic tokens consistently (per `docs/styling.md`); `dark:` overrides should be exceptions, and removing them improves token discipline overall.

### Part 3 — Clock view consumer (the example from the user story)

In `src/components/calendar/AnalogClockView.tsx`, when the active theme is `wall-projector`:

- Render an additional toggle button (next to the existing controls) labeled something like "Emphasize clock face" or just an icon (e.g. circle with light/dark halves).
- When toggled on, wrap the `<AnalogClock>` SVG in `<ThemeScope mode="light">`. The SVG's clock face background, hour numbers, and clock hands resolve to light-theme values; the surrounding card / aside / header stay dark.
- Persist the toggle in `localStorage` (`calendar_clock_face_emphasis`) so it survives reloads on a wall display that never gets manually re-toggled.
- Toggle is **only visible** when the active theme is `wall-projector` — in light or dark themes the emphasis primitive doesn't make sense.

The toggle does not affect event arc colors — those should remain readable regardless. Only the clock-face background / numerals / hands flip.

The Clock subtree may need some adjustment to fully consume semantic tokens — particularly anywhere the existing implementation hard-codes white/black. A spot check before implementation:

- `src/components/calendar/analog-clock/clock-face.tsx` — verify all fills/strokes use semantic tokens, not literal `white`/`black`.
- `src/components/calendar/analog-clock/event-arc.tsx:111` — `stroke="white"` is hard-coded. This already breaks in light theme; should become a token.
- `src/components/calendar/analog-clock/event-arc.tsx:141` — `fill="white"` for title text — same issue.

Some of those are worth a small companion clean-up before this lands; otherwise the light-island will look great but the arcs around it will have wrong stroke colors.

## Future extension points (out of scope here, but consider while designing)

- Year view: each month thumbnail could opt into light scope, giving a calendar-grid feel against a dark page.
- Day view time grid: light scope on the grid background to make hour labels and event blocks read clearer at distance.
- Adding more emphasis modes (e.g. `mode="auto-contrast"` that picks based on something) — but the current `light` / `dark` API leaves room without changing the primitive.

## Acceptance criteria

- [ ] `theme-toggle.tsx` dropdown includes "Wall-Projector" as a third option, with an appropriate icon.
- [ ] Selecting Wall-Projector applies the `.wall-projector` class to `<html>` and the page renders identically to the existing dark theme.
- [ ] `<ThemeScope mode="light">` and `<ThemeScope mode="dark">` components exist, can be nested, and produce the correct local token values regardless of outer theme.
- [ ] In Wall-Projector theme, the Clock view shows a toggle that flips the clock face into a light scope. The clock face uses semantic tokens (no hard-coded white/black) so the flip works correctly.
- [ ] Toggle state persists across reloads (localStorage).
- [ ] Toggle is hidden in light/dark themes.
- [ ] Event arc strokes/text on the clock face read correctly in the light scope (i.e. dark text on light face, not white-on-light).
- [ ] All `pnpm lint:fix && pnpm format:fix && pnpm check-types` pass; new tests pass.

## Tests (TDD per CLAUDE.md)

- Unit / component tests on `<ThemeScope>`:
  - `mode="light"` mounted under `.dark` ancestor: queried CSS custom properties on the wrapper resolve to the light values.
  - `mode="dark"` mounted under `:root` ancestor: same in reverse.
  - Nested scopes work correctly (innermost wins).
- Component test on `AnalogClockView`:
  - Toggle is hidden when theme is `light`.
  - Toggle is hidden when theme is `dark`.
  - Toggle is visible when theme is `wall-projector`.
  - Toggling on wraps the clock SVG in a `[data-theme-scope="light"]`; toggling off removes it.
  - Toggle state survives a remount when localStorage is populated.
- E2E (optional): Playwright snapshot of `/calendar?view=clock` in `wall-projector` theme, both with and without the emphasis toggle.

## Files

- `src/components/providers/ThemeProvider.tsx` — extend allowed themes if next-themes config restricts them.
- `src/components/theme/theme-toggle.tsx` — third dropdown item.
- `src/components/theme/theme-scope.tsx` — new component (small, pure).
- `src/app/globals.css` — `.wall-projector` block + `[data-theme-scope]` blocks.
- `src/components/calendar/AnalogClockView.tsx` — the toggle + ThemeScope wrap.
- `src/components/calendar/analog-clock/clock-face.tsx` and `event-arc.tsx` — replace hard-coded `white`/`black` with semantic tokens (small companion clean-up).
- New tests in `src/components/theme/__tests__/` and `src/components/calendar/__tests__/AnalogClockView.test.tsx`.

## Related

- #90 (closed) — added the original light/dark/system toggle. This builds on its `next-themes` foundation and adds a third theme.
- Future: any new view that wants to opt into emphasis just consumes `<ThemeScope>` — no further global plumbing needed.

BenSeymourODB:
Closes #319.

## Summary

Adds a third "Wall-Projector" theme — dark-by-default chrome with **scoped light "islands"** for high-density information surfaces. First consumer is the AnalogClock face, which on a wall display benefits from high black-on-white contrast for at-a-glance reading.

Implementation follows the spec on the issue:

1. **`wall-projector` theme** — registered in `next-themes`, inherits dark tokens, fourth dropdown item in `ThemeToggle`.
2. **`ThemeScope` primitive** — `[data-theme-scope="light" | "dark"]` re-declares all token vars; nestable, opposite of the outer theme.
3. **Clock view consumer** — replaces hard-coded white/grey/red SVG fills with semantic tokens, then adds an "Emphasize clock face" toggle visible only in `wall-projector` theme that wraps the SVG in `<ThemeScope mode="light">`. Toggle persists in `localStorage`.

## Phases

- **Phase 1** (this commit) — theme registration + dropdown option.
- **Phase 2** — `ThemeScope` primitive + token-var blocks.
- **Phase 3** — Clock face/event-arc semantic-token migration + AnalogClockView toggle + persistence.

## Tests

- `theme-toggle.test.tsx` — fourth Wall-Projector menu item + click handler.
- `ThemeProvider.test.tsx` — `themes` prop registers `wall-projector` alongside `light`/`dark`.
- New `theme-scope.test.tsx` (Phase 2) — DOM/data-attribute assertions.
- `AnalogClockView.test.tsx` (Phase 3) — toggle visibility per theme + persistence + ThemeScope wrap.

## CLAUDE.md compliance

- TDD: tests written before implementation each phase.
- Standard Tailwind colors only — semantic tokens (`bg-background`, `text-foreground`, `var(--card)` for SVG fills).
- React Compiler — no manual memoization.
- Strict TS, no `any`.
- pnpm only.

https://claude.ai/code/session_0119gRWfEnFrZmXwwdYTdhse

---
_Generated by [Claude Code](https://claude.ai/code/session_0119gRWfEnFrZmXwwdYTdhse)_


BenSeymourODB:
CI run [25413229869](https://github.com/BenSeymourODB/next-digital-wall-calendar/actions/runs/25413229869) failed at the "Validate Prisma migrations are up to date" step. This is the pre-existing CI gap tracked in #263 — Prisma 7's `migrate diff --from-migrations --to-schema` requires a shadow database, but the workflow doesn't provision one, so the step errors out regardless of the PR's contents.

This PR does not touch `prisma/` (verified via `git diff main..HEAD -- prisma/` shows no changes from this branch). All other CI signals — lint, type check, migration naming, test (1924 tests), and `pnpm build` — pass locally. The build step reproduction was the genuine fix from the first CI run; it's good now.

Recommend merging is blocked only by #263, not by this PR.

---

src/components/calendar/AnalogClockView.tsx
```diff
@@ -34,6 +38,25 @@ export function AnalogClockView() {
...
+
+  // Emphasize the clock face with a light scope (issue #319). Only meaningful
+  // in `wall-projector` theme; persisted in localStorage so the toggle survives
+  // reloads on always-on wall displays. Initialize synchronously from
+  // localStorage when running in the browser to avoid a one-frame flicker
+  // between the default state and the restored value.
+  const [emphasizeFace, setEmphasizeFace] = useState<boolean>(() => {
+    if (typeof window === "undefined") return false;
+    return window.localStorage.getItem(EMPHASIS_STORAGE_KEY) === "true";
+  });
```
BenSeymourODB: **BLOCKER — hydration mismatch on SSR**

`useState` lazy-initialiser reads `window.localStorage` synchronously:

```tsx
const [emphasizeFace, setEmphasizeFace] = useState<boolean>(() => {
  if (typeof window === "undefined") return false;
  return window.localStorage.getItem(EMPHASIS_STORAGE_KEY) === "true";
});
```

`AnalogClockView` is a `"use client"` component, but Next.js still server-renders it for the initial HTML. On the server `typeof window === "undefined"` is `true` so the initialiser returns `false`. On the client the initialiser runs again _before_ hydration commits and may return `true` (if the key is set). React then sees a mismatch between the server-rendered `false` state (no `ThemeScope` div) and the client-rendered `true` state (with `ThemeScope` div), producing a hydration error in production (`#319` emphasis active on a returning wall-display user).

**Fix:** Initialise to `false` always, then restore from localStorage in a `useEffect`. Accept the one-frame flicker — on an always-on wall display mounted once, it is imperceptible. Alternatively, use `suppressHydrationWarning` on the wrapper div, but the `useEffect` approach is cleaner:

```tsx
const [emphasizeFace, setEmphasizeFace] = useState(false);

useEffect(() => {
  const stored = localStorage.getItem(EMPHASIS_STORAGE_KEY);
  if (stored === "true") setEmphasizeFace(true);
}, []);
```

The comment claiming "Initialize synchronously … to avoid a one-frame flicker" is the wrong trade-off: swapping a visual flicker for a hydration error is a regression in reliability.

---

src/components/calendar/AnalogClockView.tsx
```diff
@@ -49,6 +72,17 @@ export function AnalogClockView() {
     setSelectedEvent(match);
   };
 
+  const isWallProjector = theme === "wall-projector";
```
BenSeymourODB: **MAJOR — `theme === "wall-projector"` never true when `defaultTheme="system"`**

`useTheme()` returns the raw `theme` value, which is `"system"` when next-themes is resolving the system preference — even if the resolved colour scheme happens to be dark. During server render and on first paint, `theme` can be `"system"` or `undefined`, so `isWallProjector` will be `false` and the toggle/scope branch is entirely suppressed even after a user has explicitly selected "wall-projector" and reloaded.

You should use `resolvedTheme` here:

```tsx
const { theme, resolvedTheme } = useTheme();
const isWallProjector = resolvedTheme === "wall-projector";
```

`resolvedTheme` is what next-themes emits after resolving the `"system"` alias. Because `"wall-projector"` is an explicit selection (not a system default), `theme` will equal `"wall-projector"` when the user picks it — but relying on `resolvedTheme` is the correct and future-proof pattern and avoids subtle breakage if the logic ever needs to handle `"system"` resolving to wall-projector.

---

src/components/calendar/AnalogClockView.tsx
```diff
@@ -58,18 +92,31 @@ export function AnalogClockView() {
...
+            onClick={() => setEmphasizeFace((prev) => !prev)}
+            className="border-border bg-card text-card-foreground hover:bg-accent focus:ring-ring inline-flex items-center gap-2 self-end rounded-md border px-3 py-1.5 text-xs transition-colors focus:ring-2 focus:ring-offset-1 focus:outline-none"
+          >
+            <Sparkles className="h-3.5 w-3.5" aria-hidden="true" />
+            {emphasizeFace ? "Dim clock face" : "Emphasize clock face"}
+          </button>
+        )}
         <div
           data-testid="analog-clock-wrapper"
           className="aspect-square w-full [&>svg]:h-full [&>svg]:w-full"
```
BenSeymourODB: **MAJOR — `[&>svg]` child selector broken when ThemeScope div is inserted**

The wrapper div uses `[&>svg]:h-full [&>svg]:w-full` to force the AnalogClock SVG to fill its parent:

```tsx
className="aspect-square w-full [&>svg]:h-full [&>svg]:w-full"
```

When `emphasizeFace && isWallProjector`, the DOM becomes:

```
div.aspect-square          ← has [&>svg]
  └─ div[data-theme-scope] ← className="h-full w-full"   ← ThemeScope
       └─ svg              ← the AnalogClock SVG
```

The `[&>svg]` selector targets _direct_ SVG children, but the SVG is now a grandchild. The SVG will fall back to its intrinsic `width="720" height="720"` attributes and overflow on narrow screens, breaking the responsive layout.

The ThemeScope div receives `h-full w-full`, which helps, but only if the ThemeScope div itself fills the aspect-square container — which it will here since it is the sole child. The deeper issue is that `[&>svg]` must become `[&_svg]` (descendant), or the ThemeScope div needs `[&>svg]:h-full [&>svg]:w-full` forwarded to it as well, or the aspect-square constraint needs restructuring.

Easiest targeted fix — change the wrapper's child selector to descendant on the wrapper itself:

```tsx
className="aspect-square w-full [&_svg]:h-full [&_svg]:w-full"
```

Add a test asserting the SVG's effective size with ThemeScope active.

---

src/components/calendar/AnalogClockView.tsx
```diff
@@ -49,6 +72,17 @@ export function AnalogClockView() {
...
+  const isWallProjector = theme === "wall-projector";
+
+  const clockSvg = (
+    <AnalogClock
+      size={CLOCK_MAX_PX}
+      rawEvents={events}
+      arcThickness={CLOCK_MAX_PX * ARC_THICKNESS_RATIO}
+      onEventClick={(eventId, trigger) => openEventById(eventId, trigger)}
+    />
+  );
```
BenSeymourODB: **MAJOR — React Compiler anti-pattern: JSX element stored in a `const` variable**

```tsx
const clockSvg = (
  <AnalogClock ... />
);
```

Assigning a JSX element to a plain `const` inside the render function is semantically equivalent to calling `React.createElement(...)` on every render — it does not cache the element. The React Compiler cannot automatically memoize this because the inline arrow `(eventId, trigger) => openEventById(eventId, trigger)` creates a new function reference each render, which means `AnalogClock` will always re-render when its parent does.

More importantly, storing JSX in a variable rather than composing it inline in JSX makes the code harder for the React Compiler to analyse and optimise. Per CLAUDE.md: "Write clean, simple React code." The canonical approach is to inline the JSX directly at each call-site:

```tsx
{emphasizeFace && isWallProjector ? (
  <ThemeScope mode="light" className="h-full w-full">
    <AnalogClock size={CLOCK_MAX_PX} rawEvents={events} ... />
  </ThemeScope>
) : (
  <AnalogClock size={CLOCK_MAX_PX} rawEvents={events} ... />
)}
```

Yes, this repeats the props — but it is idiomatic React and allows the Compiler to reason about each branch independently.

---

src/components/calendar/AnalogClockView.tsx
```diff
@@ -34,6 +38,25 @@ export function AnalogClockView() {
...
+    return window.localStorage.getItem(EMPHASIS_STORAGE_KEY) === "true";
+  });
+
+  useEffect(() => {
+    if (typeof window === "undefined") return;
+    window.localStorage.setItem(
+      EMPHASIS_STORAGE_KEY,
+      emphasizeFace ? "true" : "false"
+    );
+  }, [emphasizeFace]);
```
BenSeymourODB: **Minor — localStorage write fires on initial mount with `false` even when key is already stored as `"false"`**

The `useEffect` writes `"false"` to localStorage on every mount (including the first), even when `emphasizeFace` never changed from its initial value. For an always-on wall display this is harmless, but it means the effect runs storage I/O unnecessarily and could mask the absence of an initial read. Consider skipping the write when the value has not changed from the stored value, or initialising the effect with a `mounted` guard.

---

src/components/calendar/__tests__/AnalogClockView.test.tsx
```diff
@@ -258,4 +272,88 @@ describe("AnalogClockView", () => {
...
+    it("persists the toggle state to localStorage", () => {
+      mockTheme = "wall-projector";
+      renderView();
+
+      fireEvent.click(screen.getByTestId(TOGGLE_TESTID));
+
+      expect(window.localStorage.getItem(STORAGE_KEY)).toBe("true");
+    });
+
+    it("restores the toggle state from localStorage on remount", () => {
```
BenSeymourODB, line 310: **Major — missing test: corrupted / unexpected localStorage value**

The tests cover `"true"` and the default (`null`/empty), but not a corrupted value like `"1"`, `"yes"`, `"TRUE"`, or any arbitrary string. The production code uses strict equality `=== "true"`, so anything else collapses to `false` — that is the right behaviour, but it is not asserted. Add a test:

```tsx
it("treats an unrecognised localStorage value as false (not emphasised)", () => {
  window.localStorage.setItem(STORAGE_KEY, "yes");
  mockTheme = "wall-projector";
  renderView();

  const wrapper = screen.getByTestId("analog-clock-wrapper");
  expect(wrapper.querySelector('[data-theme-scope="light"]')).toBeNull();
});
```

---

src/components/calendar/__tests__/AnalogClockView.test.tsx
```diff
@@ -13,6 +13,18 @@ import { fireEvent, render, screen } from "@testing-library/react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { AnalogClockView } from "../AnalogClockView";
 
+// next-themes mock — `mockTheme` is mutated per-test to drive the
+// emphasis-toggle visibility branch in AnalogClockView. Mock has to be
+// declared before the SUT import so it is hoisted by vitest's mock loader.
+let mockTheme: string = "light";
+vi.mock("next-themes", () => ({
```
BenSeymourODB, line 17: **Major — mock does not exercise `theme === "system"` + `resolvedTheme === "wall-projector"` path**

The mock hard-codes:

```tsx
resolvedTheme: mockTheme === "system" ? "light" : mockTheme,
```

This means the case where a user selects `"system"` and the OS preference resolves to `"wall-projector"` (or vice versa — a user selects `"wall-projector"` but `theme` is briefly `"system"` during SSR) is never tested. Given the BLOCKER comment on `theme` vs `resolvedTheme` in the production code, a test that sets `mockTheme = "system"` and `resolvedTheme = "wall-projector"` would immediately expose the bug and prevent regression once it is fixed.

---

src/app/globals.css
```diff
@@ -1,7 +1,18 @@
...
+ * which inherits the dark token set by default. Components nested inside a
+ * `[data-theme-scope="light"]` subtree should use semantic tokens
+ * (bg-background, text-foreground, etc.) rather than `dark:` overrides; the
+ * scope re-declares the tokens but cannot reverse the `dark:` variant on
+ * descendants. See issue #319 and src/components/theme/theme-scope.tsx.
+ *
+ * NB: Tailwind requires `@custom-variant` on a single line — do not break it.
+ */
+/* prettier-ignore */
+@custom-variant dark (&:where(.dark, .dark *, .wall-projector, .wall-projector *));
```
BenSeymourODB, line 16: **Major — `dark:` variant now fires inside any existing component in a `wall-projector` page, including future ThemeScope consumers**

```css
@custom-variant dark (&:where(.dark, .dark *, .wall-projector, .wall-projector *));
```

This is architecturally correct for the stated goal, and the PR correctly documents the caveat in `ThemeScope`'s JSDoc and the CSS comment. However, the `button.tsx` shadcn component in this repo already uses `dark:` overrides for its `outline` variant:

```
dark:bg-input/30 dark:border-input dark:hover:bg-input/50
```

Any `Button` rendered inside a future `ThemeScope mode="light"` island (e.g., a modal or settings panel inside the clock view) will still receive the dark-variant styles because `.wall-projector` is on the `<html>` ancestor, even though the ThemeScope has re-declared the tokens to light values. The token re-declaration fixes `bg-background` etc., but `dark:bg-input/30` hard-codes dark behaviour that the scope cannot reverse.

This is acknowledged in the component's JSDoc, but there is no tracking issue or TODO comment in `globals.css` flagging the components that will need an audit before they can be safely wrapped. Consider adding a CSS comment enumerating the known affected components, or opening a follow-up issue so the footgun is tracked.

---

src/app/globals.css
```diff
@@ -74,6 +86,70 @@
...
+  --ring: oklch(0.87 0.012 285.823);
+  --sidebar-background: oklch(0.985 0.002 285.823);
+  --sidebar-foreground: oklch(0.145 0.016 285.823);
+  --sidebar-primary: oklch(0.205 0.019 285.823);
+  --sidebar-primary-foreground: oklch(0.985 0.002 285.823);
+  --sidebar-accent: oklch(0.965 0.005 285.823);
+  --sidebar-accent-foreground: oklch(0.205 0.019 285.823);
+  --sidebar-border: oklch(0.921 0.007 285.823);
+  --sidebar-ring: oklch(0.87 0.012 285.823);
+}
```
BenSeymourODB, line 129: **Minor — `[data-theme-scope]` blocks omit `--chart-*` and `--radius` tokens**

The `:root` block declares `--chart-1` through `--chart-5` (light values) and `--radius`. The `.dark` / `.wall-projector` block declares different `--chart-*` values. The `[data-theme-scope="light"]` and `[data-theme-scope="dark"]` blocks omit both:

- `--chart-1` … `--chart-5`
- `--radius`

If a chart component (or any component depending on `--radius` border-radius tokens) is ever nested inside a ThemeScope, it will fall through to the inherited value from the outer `.wall-projector` ancestor, which means a light-scoped chart on a wall-projector page will show dark chart colours. This is likely not a problem today since the clock SVG doesn't use chart tokens, but the comment says "Values mirror the :root (light) and .dark blocks above; keep them in sync" — and right now they do not. Either add the missing tokens or amend the comment to document the deliberate exclusions.

---

src/components/theme/theme-toggle.tsx
```diff
@@ -40,6 +40,10 @@ export function ThemeToggle({ className }: ThemeToggleProps) {
           <Moon className="mr-2 h-4 w-4" />
           Dark
         </DropdownMenuItem>
+        <DropdownMenuItem onClick={() => setTheme("wall-projector")}>
```
BenSeymourODB, line 43: **Major — settings API `VALID_THEMES` does not include `"wall-projector"`**

This PR adds `"wall-projector"` to the next-themes theme list and to the UI dropdown, but `src/app/api/settings/route.ts` still has:

```ts
const VALID_THEMES = ["light", "dark", "auto", "system"];
```

And `src/components/settings/display-section.tsx` still only offers `["light", "dark", "system"]` in its radio group. If the settings form is ever wired to persist the theme via that API endpoint, a user selecting Wall-Projector here and then hitting "Save" will receive a `400`. Additionally the wall-projector setting would be silently lost on next page load.

Even if the settings persistence path is out of scope for this PR, the mismatch should be tracked (follow-up issue or TODO) so it does not accidentally reach production as a partially-wired feature.

---

src/components/calendar/analog-clock/clock-face.tsx
```diff
@@ -81,8 +82,9 @@ export function ClockFace({
             y1={outerY}
             x2={innerX}
             y2={innerY}
-            stroke="#d1d5db"
+            stroke="var(--muted-foreground)"
             strokeWidth={0.75}
+            opacity={0.4}
```
BenSeymourODB: **Minor — minute-tick `opacity={0.4}` is a new visual change not covered by any test**

The minute-tick marks were previously `stroke="#d1d5db"` with no explicit opacity; they now use `var(--muted-foreground)` + `opacity={0.4}`. This changes the rendering in both light and dark themes (not just under wall-projector). The existing `clock-face.test.tsx` has no assertion on tick stroke or opacity, so this visual regression risk is undetected. Consider adding a snapshot or attribute assertion for the minute-tick opacity.

---

src/components/calendar/analog-clock/clock-face.tsx
```diff
@@ -168,21 +170,22 @@ export function ClockFace({
...
         <line
           data-testid="second-hand"
           x1={cx}
           y1={roundCoord(cy + faceRadius * 0.12)}
           x2={cx}
           y2={roundCoord(cy - secondHandLength)}
-          stroke="#ef4444"
+          stroke="var(--destructive)"
           strokeWidth={1.5}
           strokeLinecap="round"
```
BenSeymourODB, line 189: **Minor — second hand uses `var(--destructive)` which is the same value in both light and dark scope**

The PR's stated intent is that the second hand "stays visually distinct against any face/foreground combination." In the light scope, `--destructive` is `oklch(0.577 0.245 27.325)` — a medium-saturation red-orange against near-white `--card`. In the dark scope the same value sits against `--card` at `oklch(0.205 …)`. The contrast ratio is approximately 3.5:1 — above the WCAG AA threshold for graphical objects (3:1) but only just. Worth a note for future audit; not a blocker.

---

src/components/theme/__tests__/theme-scope.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+          <ThemeScope mode="dark">
+            <div data-testid="dark-probe" />
+          </ThemeScope>
+          <ThemeScope mode="light">
+            <div data-testid="light-probe" />
+          </ThemeScope>
+        </div>
+      );
+
+      const darkVar = getComputedStyle(
```
BenSeymourODB, line 100: **Minor — CSS custom property test relies on jsdom `getComputedStyle` behaviour that varies across jsdom versions**

The injected `<style>` tag approach is creative and well-commented. However, jsdom's `getComputedStyle` support for custom properties (`--background`) was unreliable before jsdom 20 / happy-dom. If the test environment upgrades or switches JSDOM, this test could silently start returning empty strings for the custom property, making the assertion `expect(darkVar).toBe("rgb(20, 20, 20)")` fail even though the production CSS is fine.

Consider pinning the jsdom version expectation in a comment, or using a simpler assertion that checks the attribute value on the element rather than the computed style. Alternatively, note explicitly that this test is an approximation and the real validation is a visual regression test or E2E.

---

src/components/providers/ThemeProvider.tsx
```diff
@@ -7,13 +7,19 @@ interface ThemeProviderProps {
...
+const THEMES = ["light", "dark", "wall-projector"] as const;
+
 export function ThemeProvider({ children }: ThemeProviderProps) {
   return (
     <NextThemesProvider
       attribute="class"
       defaultTheme="system"
       enableSystem
       disableTransitionOnChange
+      themes={[...THEMES]}
```
BenSeymourODB, line 22: **Nit — `as const` on `THEMES` produces a `readonly` tuple; spreading into `themes={[...THEMES]}` is correct but the type widens to `string[]`**

`NextThemesProvider` expects `themes?: string[]`, so the spread is necessary and correct. This is fine. Just noting: if a future version of `next-themes` accepts a readonly array, the spread can be dropped. No action required.

---

BenSeymourODB (commented):
## Review: Wall-Projector theme + ThemeScope primitive (#319)

The overall architecture is sound — CSS-variable re-declaration via `[data-theme-scope]` is the correct approach and the token migration in the clock SVG is thorough. TDD structure is present, TS is strict, no manual memoisation. However there are **three blockers and two majors** that need resolving before merge.

---

### Blockers

1. **Hydration mismatch** (`AnalogClockView.tsx` line 51) — the `useState` lazy-initialiser reads `window.localStorage` synchronously. The server returns `false` (SSR guard fires); the client initialiser runs again and may return `true`, causing a React hydration error. Fix: initialise to `false`, restore in `useEffect`.

2. **SVG layout breaks when ThemeScope div is inserted** (`AnalogClockView.tsx` line 110) — the wrapper uses `[&>svg]:h-full [&>svg]:w-full` (direct-child selector). When the `ThemeScope` `<div>` is inserted between the wrapper and the SVG, the SVG falls back to its intrinsic 720×720 size and overflows on narrow screens. Fix: change `[&>svg]` to `[&_svg]` (descendant selector).

3. **`theme` vs `resolvedTheme`** (`AnalogClockView.tsx` line 75) — `useTheme().theme` can be `"system"` on first render even when the user has selected wall-projector, suppressing the toggle entirely. Use `resolvedTheme` instead.

### Majors

4. **JSX-in-`const` variable anti-pattern** (`AnalogClockView.tsx` line 77) — storing a JSX element in `const clockSvg` is not idiomatic React and impedes the React Compiler. Inline both branches directly.

5. **Settings API / display-section not updated** (`theme-toggle.tsx` line 43) — `VALID_THEMES` in `settings/route.ts` and the radio group in `display-section.tsx` still exclude `"wall-projector"`. If a user selects it and the settings are saved, the API returns 400. Also, `sonner.tsx` passes the raw theme string (including `"wall-projector"`) to a Sonner prop that only accepts `"light" | "dark" | "system"`.

6. **Missing test cases** — corrupted localStorage value; `theme="system"` + `resolvedTheme="wall-projector"` path.

### Minors / Nits

- `[data-theme-scope]` blocks omit `--chart-*` and `--radius` tokens (inconsistent with the "keep in sync" comment).
- `dark:` leakage into ThemeScope islands should be tracked in a follow-up issue, not just documented in JSDoc.
- `useEffect` writes localStorage on every mount even when the value has not changed.
- Minute-tick `opacity={0.4}` change affects all themes but has no test coverage.
- CSS custom-property test via `getComputedStyle` is jsdom-version-sensitive; worth a comment.

---

src/components/calendar/AnalogClockView.tsx
BenSeymourODB: Fixed in b1d9c41. Refactored to `useSyncExternalStore` rather than `useState` + `useEffect`: the hook returns the server snapshot (always `false`) during SSR and the first client paint, then commits the live `localStorage` value post-hydration, so there is no mismatch when a returning user has emphasis active. As a side benefit it also satisfies the `react-hooks/set-state-in-effect` lint rule that the simpler `useEffect → setState` form trips.

---

src/components/calendar/AnalogClockView.tsx
BenSeymourODB: Fixed in b1d9c41 — `isWallProjector` now keys off `resolvedTheme === "wall-projector"`. Added test coverage for the `theme="system"` + `resolvedTheme="wall-projector"` path (toggle visible) and `resolvedTheme="dark"` (hidden) so the regression is locked in.

---

src/components/calendar/AnalogClockView.tsx
BenSeymourODB: Fixed in b1d9c41 — wrapper now uses `[&_svg]:h-full [&_svg]:w-full` (descendant). Added a regression test in `AnalogClockView.test.tsx` asserting both that the wrapper class no longer contains `[&>svg]` and that the SVG is reachable as a grandchild of the wrapper through the `ThemeScope` div when emphasis is on.

---

src/components/calendar/AnalogClockView.tsx
BenSeymourODB: Fixed in b1d9c41 — the `const clockSvg = (...)` extraction is gone; both ternary branches inline `<AnalogClock />` directly with their own props. Yes the props repeat, but each branch is now independent for the React Compiler and follows CLAUDE.md's "clean, simple React code" guidance.

---

src/components/calendar/AnalogClockView.tsx
BenSeymourODB: Resolved by the `useSyncExternalStore` refactor in b1d9c41. There is no more `useEffect` writing back to localStorage on every mount — writes only happen inside `toggleEmphasis`, which is invoked from the click handler. Initial mount with stored `"false"` performs zero localStorage writes.

---

src/components/calendar/__tests__/AnalogClockView.test.tsx
BenSeymourODB: Added in b1d9c41:

```tsx
it("treats an unrecognised localStorage value as false (not emphasised)", () => {
  window.localStorage.setItem(STORAGE_KEY, "yes");
  mockTheme = "wall-projector";
  renderView();

  const wrapper = screen.getByTestId("analog-clock-wrapper");
  expect(wrapper.querySelector('[data-theme-scope="light"]')).toBeNull();
});
```

Locks in the strict `=== "true"` semantics.

---

src/components/calendar/__tests__/AnalogClockView.test.tsx
BenSeymourODB: Mock now exposes a separate `mockResolvedTheme` override (b1d9c41). Two new tests exercise the path:

- `mockTheme = "system"` + `mockResolvedTheme = "wall-projector"` → toggle visible.
- `mockTheme = "system"` + `mockResolvedTheme = "dark"` → toggle hidden.

---

src/app/globals.css
BenSeymourODB: Tracked in #324 (audit `dark:` overrides on components that may sit inside ThemeScope islands), opened in this round. Updated the JSDoc in `theme-scope.tsx` and the comment in `globals.css` to point at it instead of leaving the caveat as a dangling note. Did not migrate `Button`'s `outline` variant in this PR — that audit is the right scope for the follow-up issue rather than expanding this one.

---

src/app/globals.css
BenSeymourODB: Fixed in b1d9c41 — `--chart-1` through `--chart-5` are now declared in both `[data-theme-scope]` blocks with the matching light / dark values from the root blocks. `--radius` is intentionally omitted because it is the same value across light and dark, so re-declaring it adds no information; clarified that in the comment.

---

src/components/theme/theme-toggle.tsx
BenSeymourODB: Fixed inline in this PR rather than deferred (b1d9c41):

- `VALID_THEMES` in `src/app/api/settings/route.ts` now includes `"wall-projector"`.
- `THEME_OPTIONS` in `src/components/settings/display-section.tsx` now lists `["light", "dark", "wall-projector", "system"]` and a new `THEME_LABELS` map renders "Wall-Projector" with proper casing instead of relying on `capitalize`.
- `src/components/ui/sonner.tsx` maps `theme === "wall-projector"` → `"dark"` for the Sonner `theme` prop (which only accepts `"light" | "dark" | "system"`).

Added a docblock above each list pointing at the other two so they don't drift.

---

src/components/calendar/analog-clock/clock-face.tsx
BenSeymourODB: Reverted the `opacity={0.4}` hack in b1d9c41 and switched the minute-tick stroke to `var(--border)` instead of `var(--muted-foreground)`. `--border` is the existing subtle-line token and produces an appropriately muted line in both themes without the explicit opacity. Also added a `clock-face.test.tsx` assertion locking in the migrated stroke values for face background, hour markers/numbers, hands, center dot, period indicator, and second hand so any future regression is caught.

---

src/components/calendar/analog-clock/clock-face.tsx
BenSeymourODB: Acknowledged. Filing this as future-audit material rather than changing in this PR — the second hand has been `#ef4444` (a more saturated red, similar luminance) on this codebase since the AnalogClock landed, so adopting `--destructive` is contrast-neutral, not a regression. If we want a richer red on a light face we can introduce a dedicated `--clock-second-hand` token in a follow-up; happy to open one if you'd like.

---

src/components/theme/__tests__/theme-scope.test.tsx
BenSeymourODB: Documented the limitation in the test (b1d9c41) — the inline comment now explicitly notes that this jsdom-based assertion is an approximation and that the full visual contract (Tailwind classes flipping under the scope) is meant to be covered by Playwright/visual regression. If a future jsdom upgrade silently breaks the custom-property read, the comment should make the failure mode obvious to the next reader.

---

src/components/providers/ThemeProvider.tsx
BenSeymourODB: Acknowledged — no change. The `as const` + spread is intentional: it lets us reference `THEMES` as a readonly tuple type (e.g. for the matching `THEME_OPTIONS` / `THEME_LABELS` in `display-section.tsx`) while still producing the `string[]` that `next-themes` wants.

---

PR #325, opened 2026-05-06
feat(calendar): wire Zod validation into all server-side Google Calendar routes (#277)

BenSeymourODB:
Closes #277.

## Summary

PR #233 introduced `GoogleEventSchema`, `GoogleEventsListResponseSchema`, `GoogleCalendarListResponseSchema`, and the `parseGoogleResponse` helper, but no route actually wired them in. This PR closes that gap so every server-side Google Calendar response gets validated at the route boundary, with malformed payloads surfacing as a 502 plus a structured `logger.error` carrying the `GoogleApiValidationError`, endpoint name, calendar id, and the first five Zod issues.

## Status — ready for review

All three phases below have landed on this branch. Latest scheduled-routine run (2026-06-06) merged `origin/main`, resolved a conflict in `/api/calendar/calendars/route.ts` introduced by #266/#268's `accessRole` field, added a `narrowAccessRole(string | undefined) → TCalendarAccessRole` helper at the trust boundary, and added 7 unit tests for it.

Locally on the merged tip: **146 test files / 2339 tests passing**; `pnpm lint:fix` (0 errors, 6 pre-existing warnings on unrelated files); `pnpm check-types` clean; `pnpm format:fix` clean.

## Phases

- **Phase A** — `fetchEventsFromCalendar` validates with `GoogleEventsListResponseSchema`; `POST /api/calendar/events` validates the insert response with `GoogleEventSchema`. Per-route 502 + structured-log tests.
- **Phase B** — same treatment in `GET /api/calendar/calendars` and `GET /api/calendar/colors`.
- **Phase C** — document the trust boundary for client-side gapi calls in `src/lib/google-calendar.ts`.

## Merge with main (2026-06-06)

Main added `summaryOverride` and the required `accessRole: TCalendarAccessRole` field to `CalendarInfo` (#266/#268). The PR's typed Zod callback (`(item: GoogleCalendarListEntry) => ...`) types `item.accessRole` as `string | undefined`, which is not assignable to the literal union. Resolution:

- **New helper** `narrowAccessRole(value: string | undefined): TCalendarAccessRole` in `src/types/calendar.ts` — fails closed to `"reader"` for missing / unrecognised / case-variant values. Future Google additions still pass Zod validation (the schema stays permissive) but get a safe default at the route boundary.
- **7 unit tests** in `src/types/__tests__/narrow-access-role.test.ts` pinning down the failure-mode contract.
- **Merge commit** preserves the published SHAs of the original 5 commits — no force-push.

## Out of scope (deferred)

- `events.patch` / `events.get` — neither route exists at HEAD. The future PATCH PR (issue #265) should reuse `GoogleEventSchema` + `parseGoogleResponse` directly.
- `events.delete` — Google returns 204 No Content, no body to validate.

## Tests added

`events.list` / `events.insert` / `calendarList.list` (twice — used by both `calendars` and `colors` routes):

- Item missing the required `id` → 502 + `logger.error` with the right `endpoint` and `calendarId`.
- `items` field the wrong type (string instead of array, number) → 502 + log.
- Multi-calendar partial failure: one well-formed + one malformed → 200 with the bad calendar surfaced in `errors[]` as status 502, and `logger.error` fires exactly once (the inner `GoogleApiValidationError` log; the outer "Calendar fetch error" envelope is correctly suppressed via the `logged?: boolean` flag on `CalendarFetchError`).
- `events.insert` response shaped as an array → 502.
- `narrowAccessRole`: four canonical roles pass through; `undefined` / unknown strings / case-variants all fall back to `"reader"`.

## Test plan

- [x] `pnpm test:run` → 2339 / 2339 passing on the merged tip
- [x] `pnpm lint:fix` → 0 errors (6 pre-existing warnings on unrelated files)
- [x] `mergeable_state: "clean"` against `main` @ `d6daba5`

https://claude.ai/code/session_0159PCerNUcqtcdZ5w8YW5VQ

---
_Generated by [Claude Code](https://claude.ai/code/session_0159PCerNUcqtcdZ5w8YW5VQ)_


BenSeymourODB (commented):
**Overall:** the validation wiring is solid — every Google response path now goes through `parseGoogleResponse`, status codes are consistently 502, and the test coverage is thorough. One double-log needs fixing before merge; one nit and one question are worth a look.

---

**[blocker] Double `logger.error` when `events.list` validation fails on a multi-calendar request**

`fetchEventsFromCalendar` already calls `logger.error(validationError, {...})` at line ~104 in `events/route.ts` before returning the error result. The outer `GET` aggregation loop (line 214) then calls `logger.error(new Error("Calendar fetch error"), {...})` for *every* error in `results`, which fires a second time for the same validation failure.

For a single-calendar request this isn't visible (it returns early before aggregation), but any multi-calendar call with a malformed second calendar logs twice: once with the rich `GoogleApiValidationError` + structured issues, once with a generic bare `Error`. The second log has no `validationIssues`, so it degrades the signal and doubles the noise.

Fix options:
1. Skip the outer `logger.error` when `result.error.status === 502` (validation errors self-log).
2. Add a boolean flag `result.error.logged?: true` and skip accordingly.
3. Move all error logging to the outer loop and remove it from `fetchEventsFromCalendar`.

The multi-calendar partial-failure test (`"treats a malformed response on one of two calendars as a per-calendar error"`) asserts `logger.error` was called but doesn't assert `toHaveBeenCalledTimes(1)`, which is why this slipped through. Consider tightening that assertion.

---

**[question] `let parsed: T` annotation — is this papering over a real issue?**

Commit 43cccef explains that CI's `tsc` (5.9 / Zod 4) inferred a wider type on the `.map()` callback and failed, while the local lockfile didn't reproduce it. The fix is sound defensive practice: explicit `let parsed: GoogleEventsListResponse` removes the ambiguity without any behaviour change, and it makes the flow-narrowing across the try/catch explicit rather than relying on control-flow inference.

The one thing worth checking: is the `tsc` version pinned in CI? If CI was running a newer tsc than local and surfaced a real inference gap, the fix is correct regardless, but it's worth making sure `.nvmrc` / the CI workflow's Node/tsc versions are aligned so this doesn't surface as a false-local-green again.

---

**[nit] `calendarId: error.calendarId ?? ""`**

`GoogleApiValidationContext.calendarId` is `string | undefined`. Substituting `""` keeps the `Properties` type happy (`Record<string, string|number|boolean>`), but an empty string in Application Insights is harder to filter than simply omitting the key. Consider only including the property when it's defined:

```ts
...(error.calendarId ? { calendarId: error.calendarId } : {}),
```

Same pattern applies in both the `events.list` catch (line ~106) and `events.insert` catch (line ~568).

---

**Deferred work confirmed:** `[id]/route.ts` only exports `DELETE`; no `PATCH` or `GET` exists at HEAD. The PR body's claim is accurate.

---

BenSeymourODB (commented):
🤖 In-depth review pass from `implement-issue` (scheduled routine, fallback PR-review mode).

## Verified locally on the merged tip (`d7a9974`)

- `pnpm check-types` — clean
- `pnpm lint` — 0 errors, 6 pre-existing warnings on unrelated files (no new ones introduced)
- `pnpm test:run` — **2347 / 2347 passing** (was 2346 on main; +1 from the new regression test below)
- `pnpm format:fix` — clean

## What I changed in this pass

Pushed one fix commit (`d7a9974`):

**fix(calendar): strip internal `logged` flag from partial-failure wire response**

`CalendarFetchError.logged` is the server-only dedupe guard introduced in `bbb557c` so the outer GET aggregator doesn't re-log a per-calendar validation failure that `fetchEventsFromCalendar` already logged with a rich `GoogleApiValidationError` entry. Before this commit it was being pushed straight into `errors[]` on the partial-failure 200 response, so clients saw an internal logging-strategy flag on the wire.

Fix splits the type:
- `InternalCalendarFetchError` (carries `logged`) is used inside the route.
- `CalendarFetchError` (no `logged`) is the wire shape.
- A small `toWireFetchError()` projector maps between them at serialisation time.

Added a regression test in `route.test.ts` that pins the wire shape to exactly `{ calendarId, error, status }` for a multi-calendar partial-failure scenario.

## Other findings — not addressed in this PR

I flagged these to the user as out-of-scope for the current slice. Filing here so they don't get lost:

1. **Multi-calendar all-fail returns 200 (pre-existing).** The guard at `src/app/api/calendar/events/route.ts:245` only short-circuits when `calendarIds.length === 1`. If two requested calendars both fail (validation, 5xx, etc.) the route falls through to a 200 with `{ events: [], errors: [...] }` and even logs `"Calendar events fetched"` at info level. This is unchanged-from-main behaviour, but the new Zod-validation path makes it more likely to fire silently (a Google-side schema regression could fail all calendars simultaneously). Worth filing a follow-up issue to widen the all-fail short-circuit to `errors.length === calendarIds.length`.

2. **Defensive conditional spread on `error.calendarId`** at `route.ts:122` (and the matching POST site). In every current call site `parseGoogleResponse` is invoked with `calendarId: <string-param>`, so `error.calendarId` is never undefined — the `...(error.calendarId ? {...} : {})` is dead code in practice. Keeping it for now since it guards against a future caller omitting the field, but it could be tightened in a follow-up by making `GoogleApiValidationContext.calendarId` required on the events-route call sites only.

3. **Nit — redundant inline type annotations** on the `.map()` callbacks across all three routes (`events`/`calendars`/`colors`). Zod's `z.infer` already types `parsed.items`, so `(item: GoogleCalendarListEntry) => ...` is redundant. Per the commit message in `43cccef` ("pin Zod-validated types to avoid tsc inference quirk"), this was added defensively; if that quirk has since been resolved on the current TypeScript version, the annotations could be dropped. Not a blocker.

## My assessment

PR is good to merge after the `d7a9974` fix lands. Findings #1–#3 are follow-up material, not regressions introduced by this PR.

---

BenSeymourODB:
🤖 `implement-issue` claiming this PR for an in-depth review session. The prior claim on 2026-06-06T15:01Z went stale (>12h, no review threads landed), and the PR is otherwise clean (CI green, `mergeable_state: clean`) with substantive scope — best target for the routine's PR-review fallback. Will run the full local quality sweep, read the diff thoroughly, and post findings as inline review comments.

---

BenSeymourODB (commented):
## In-depth review — `implement-issue` routine pass

**Verdict: ready for human merge** after one defensive cleanup pushed in `d3b5ee2`.

### Local verification on the rebased tip (`d3b5ee2`, was `d7a9974`)

- `pnpm test:run` → **2347 / 2347 passing** across 147 files (was 2339 / 2339 on `d7a9974`; new total reflects the existing `narrow-access-role` suite — no test added/removed by my pass)
- `pnpm lint` → 0 errors, 6 pre-existing warnings unrelated to this PR
- `pnpm check-types` → clean
- `pnpm format --check` → clean
- CI: green on the prior tip (`d7a9974`, run 27065945963); will re-run on `d3b5ee2`

### What I pushed (`d3b5ee2`)

**`src/types/calendar.ts` — make `narrowAccessRole` exhaustive over the `TCalendarAccessRole` union.** The previous runtime guard was a hand-maintained `Set` with no compile-time link to the union:

```ts
// Before
const VALID_ACCESS_ROLES: ReadonlySet<TCalendarAccessRole> = new Set([
  "freeBusyReader", "reader", "writer", "owner",
]);
```

`ReadonlySet<TCalendarAccessRole>` constrains what *values* can be inserted but does **not** force the Set to be *exhaustive*. If a future contributor adds a fifth role to the union (e.g. Google introduces `"delegate"` and we add it to `TCalendarAccessRole`), the compiler is silent — the Set just misses the new value and `narrowAccessRole` silently downgrades it to `"reader"`. That's the exact failure mode the PR's existing tests already pin: an unrecognised role becomes `"reader"`. Acceptable for an *unknown* role from Google, but unacceptable for one the codebase has declared as known.

Swapped to a `Record<TCalendarAccessRole, true>` lookup so a missing key is a compile error, and switched the membership test to `Object.prototype.hasOwnProperty.call` to keep the prototype chain out of the trust boundary:

```ts
// After
const VALID_ACCESS_ROLES: Record<TCalendarAccessRole, true> = {
  freeBusyReader: true, reader: true, writer: true, owner: true,
};
// hasOwnProperty.call below — so toString/constructor never match
```

All 7 existing `narrow-access-role` tests pass unchanged; the contract (canonical values pass through, everything else falls back to `"reader"`) is identical.

### Observations (no action — flagging for future cleanup, not blockers)

1. **Unused `interface CalendarInfo` in `route.test.ts:46-56`** — pre-existing per `git blame` (`8fc429e`, 2026-05-04), surfaced as the only unused-vars warning on a file this PR touches. The test file now imports `CalendarsResponse` from the route, so the local copy is dead code. Worth deleting but not in this PR's scope.

2. **`event as gapi.client.calendar.Event` casts (`events/route.ts:157`, `:622`)** — the in-line comment correctly justifies these (Zod loose schema + structural spread make them safe). A wider refactor that has `normalizeFetchedEvent` accept `GoogleEventPayload` directly would remove the casts, but that touches both the server route and the legacy `lib/google-calendar.ts` callers — out of scope.

3. **`validationIssues` slice length inconsistency** — `GoogleApiValidationError` constructor truncates issues to 3 in the message (`google-calendar-schemas.ts:266`), but every route logs `error.issues.slice(0, 5)`. Both are reasonable defaults; harmonising on one number would just be cosmetic.

4. **`GoogleApiErrorBodySchema` is exported + tested but unused in production paths** — both routes still use raw `errorData.error?.message` access on 4xx error bodies (`events/route.ts:115`, `:676`, `events/[id]/route.ts:136`). Pre-existing from PR #233; issue #277's acceptance criteria only cover success-body validation, so this is an orthogonal trust-boundary gap — worth a follow-up issue (or a YAGNI delete) but explicitly out of scope here.

5. **All-fail multi-calendar GET still returns 200** (`events/route.ts:263-268`) — the "all calendars failed → return error" branch only fires when `calendarIds.length === 1`. A 2-calendar request where both fail still returns 200 with `events: []` and a populated `errors[]`. Pre-existing semantics; the new validation-failure path correctly populates `errors[i].status: 502` so a caller can detect the all-fail multi-calendar case from the errors array. Worth a follow-up to tighten if any UI consumer would benefit from a clearer "everything broke" signal.

### CLAUDE.md compliance

- TDD: every new branch has a corresponding 502 + structured-log test
- No manual memoisation
- Strict TS; the only casts are the documented `gapi.client.calendar.Event` ones at the Zod/mapper boundary
- No new Tailwind, no new UI surface
- The `narrowAccessRole` helper is the single fail-closed trust boundary that the issue's acceptance criteria called for

The PR delivers exactly the issue #277 contract end-to-end across `events.list`, `events.insert`, `calendarList.list` (twice — `calendars` and `colors` both consume it), with the client-side `google-calendar.ts` trust boundary documented in line with the issue's third acceptance criterion.

---

PR #326, opened 2026-05-06
feat(rewards): add PointsBadge + PointsAnimation, wire to TaskItem and Settings (#173)

### LINKED ISSUE #173: feat(rewards): add PointsBadge + PointsAnimation, wire to TaskItem and Settings

BenSeymourODB:
Closes #173.

Adds the visible reward surface and wires it into task completion + the settings page. Schema (#171) and API + PointsContext (#172) already landed — this PR delivers the user-facing piece.

### `PointsBadge` (`src/components/rewards/points-badge.tsx`)
Persistent header badge — 🏆 + locale-formatted total + "pts". Rendered fixed top-right inside `AppShell`. Hidden when rewards are disabled, no profile is active, or no `PointsProvider` is mounted (defensive — keeps `AppShell` tests free of context coupling). Uses standard Tailwind colors only.

### `PointsAnimation` (`src/components/rewards/points-animation.tsx`)
Transient "+N points! 🎉" banner. Visibility is fully prop-driven — parent passes `show={true}`, the component fires `onComplete` after ~2s so the parent flips show back to false. Keeps the component free of internal visibility state and side-effectful `setState` calls. Respects `prefers-reduced-motion` (renders the polite live region but skips animation classes), built on the existing `useReducedMotion` hook.

### `MountedPointsProvider` (`src/components/rewards/mounted-points-provider.tsx`)
Small wrapper that pulls `profileId` from `ProfileProvider` and forwards it to `PointsProvider`. Mounted in the root layout inside `ProfileProvider`. Keeps `PointsProvider` itself free of any direct coupling to the profile context (and therefore trivially unit-testable).

### API + context extension
`GET /api/points` now also returns `defaultTaskPoints` and `showPointsOnCompletion` from `UserSettings` (with schema-default fallbacks of `10` / `true` when no settings row exists). `PointsContext` consumes these and exposes them on the value, with the same defaults so behaviour degrades gracefully if the server is older.

Added `usePointsOptional()` for callers that should fall through gracefully when the provider is absent — used by `PointsBadge`, `TaskItem`, and `RewardSection`.

### `TaskItem` integration
`TaskItem` becomes reward-aware via `usePointsOptional()`: on the `incomplete → complete` transition (and only that transition) it POSTs to `/api/points/award`. The `+N` animation only shows when the server reported a fresh award (`alreadyAwarded === false`) AND the user has `showPointsOnCompletion` enabled — re-completing the same task is a no-op visually as well as on the ledger (server unique index already prevents double-credit). All existing `TaskItem` tests still pass; six new integration tests cover the reward-aware branches plus the no-provider fallback.

### `RewardSection` total-points display
When rewards are enabled and a profile is active, the settings section now shows a "Total points" row driven by `usePointsOptional()`. Three new tests cover enabled/disabled and no-provider cases.

### Playwright E2E (`e2e/reward-points.spec.ts`)
Drives a new `/test/rewards` harness that mounts `PointsProvider` with a fixed profile id together with a `TaskList` and `PointsBadge`. Three cases:
1. badge starts at 50 → completing a task POSTs award → animation appears → badge updates to 60 (fresh credit)
2. re-completing an already-awarded task is a no-op (server returns `alreadyAwarded: true` → no animation, badge stays put)
3. badge is hidden when the user has rewards disabled

Video capture is on for this spec so the PR can attach a clip.

## Acceptance

- [x] Toggling tasks back to incomplete and re-completing does NOT award additional points (`alreadyAwarded` from the server's unique `(profileId, taskId, reason)` index suppresses both the credit and the animation).
- [x] Settings UI persists via the existing `userSettings` API route (unchanged).
- [x] Component test for `PointsBadge` (hidden when disabled, shown when enabled).
- [x] Playwright happy-path: enable rewards → complete task → see animation + badge increment.

## Test results

- `pnpm test:run` — **132 files, 2044 tests, all passing**
- `pnpm lint` — clean
- `pnpm check-types` — clean
- E2E was **not run locally** — Playwright browser binaries could not be downloaded in this sandbox; the spec is type-checked and lint-clean. Please run `pnpm test:e2e e2e/reward-points.spec.ts` on a developer machine before merging.

## Note on CI

The current `ci` failure on this PR is **pre-existing on main** and unrelated to this change: `prisma/migrations/` already has two `0005_*` directories merged from independently-merged PRs (#193 `0005_add_meal_planning` and #198 `0005_pointtransaction_unique_task_award` — the latter should have been `0006_*`). `pnpm db:migrate:check-names` rejects the duplicate prefix on every PR. See [my comment above](#issuecomment-4385066599). Worth its own chore PR — renaming the directory would also need a `_prisma_migrations` rename in any environment that already applied it, so it's not a one-line fix.

## Plan reference

`.claude/plans/reward-point-system.md` (sections "Points Badge", "Points Animation", "Updated TaskItem with Points", "Integration with App Layout").

---


BenSeymourODB:
CI is failing on the **"Validate Prisma migration naming convention"** step (`pnpm db:migrate:check-names`):

```
::error::Migration "0005_pointtransaction_unique_task_award" has prefix 0005, expected 0006. Prefixes must be strictly sequential starting at 0001 (no gaps, no duplicates).
```

This is **pre-existing on `main`** and not caused by anything in this PR — `prisma/migrations/` already contains two `0005_*` directories from independently-merged PRs:

- `0005_add_meal_planning` (#193, merged earlier)
- `0005_pointtransaction_unique_task_award` (#198, should have been `0006_*`)

`prisma migrate diff` itself exits clean — only the naming check is unhappy. The fix is a directory rename of `0005_pointtransaction_unique_task_award` → `0006_pointtransaction_unique_task_award`, but that's its own concern (touches the `_prisma_migrations` table on any environment that has already applied it) and shouldn't ride on top of unrelated rewards-UI work. Worth a separate PR / quick chore commit on `main`.

This PR's logic and tests are fine — running `pnpm test:run src/components/rewards` locally is green for all 27 cases.

---

src/components/rewards/points-animation.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  useEffect(() => {
+    if (!show) {
+      return;
+    }
+    const timer = setTimeout(() => {
+      onComplete?.();
+    }, DISMISS_AFTER_MS);
+    return () => clearTimeout(timer);
+  }, [show, onComplete]);
```
BenSeymourODB: **`onComplete` in the `useEffect` dep array will restart the 2-second dismiss timer on every parent re-render that passes a new function reference.**

`TaskItem` passes `onComplete={() => setAnimationPoints(null)}` — an inline arrow created fresh each render. Every time `TaskItem` re-renders while the banner is showing (e.g. during the POST fetch that updates `totalPoints`), `onComplete` changes, the effect re-fires, `clearTimeout` cancels the old timer, and a fresh 2000 ms window begins. In practice this means the banner can stay visible longer than intended whenever the parent re-renders mid-animation.

Because the React Compiler handles memoization, the correct fix is to use a ref so the callback identity doesn't matter:

```ts
const onCompleteRef = useRef(onComplete);
onCompleteRef.current = onComplete;   // keep in sync each render

useEffect(() => {
  if (!show) return;
  const timer = setTimeout(() => {
    onCompleteRef.current?.();
  }, DISMISS_AFTER_MS);
  return () => clearTimeout(timer);
}, [show]);   // ← onComplete no longer a dep
```

This pattern is idiomatic React (no useCallback needed) and the existing test "fires onComplete after the dismiss timer elapses" still passes because the ref is always current.

---

src/components/rewards/points-animation.tsx
```diff
@@ -0,0 +1,63 @@
...
+
+  if (!show) {
+    return null;
+  }
+
+  return (
+    <div
```
BenSeymourODB, line 60: **`aria-live="polite"` fires asynchronously — consider `"assertive"` for transient celebratory announcements.**

`polite` defers the announcement until the user is idle; if the user is typing or navigating immediately after completing a task the "+10 points!" message may never be read aloud. For a brief celebratory banner that disappears after 2 seconds, `aria-live="assertive"` is more appropriate — it interrupts immediately, matching how a sighted user experiences the visual pop-in. The existing test on line 99 already asserts `aria-live="polite"` and should be updated in the same commit.

(This is a subjective a11y call, not a hard rule, but worth an explicit decision either way.)

---

e2e/reward-points.spec.ts
```diff
@@ -0,0 +1,235 @@
...
+ *
+ * Drives the `/test/rewards` harness which mounts a `PointsProvider`
+ * with a fixed `profileId` plus a `TaskList` and the `PointsBadge`.
+ * `/api/points*` and `/api/tasks*` are intercepted via `page.route()`
+ * so the spec runs without a real session.
+ *
+ * Video capture is enabled on this whole spec so the PR description
+ * can include a clip of the badge increment + animation banner.
+ */
+test.use({ video: "on" });
```
BenSeymourODB, line 14: **`test.use({ video: "on" })` records video for every run, not just failures — this will accumulate large artifacts in CI.**

The base `playwright.config.ts` does not configure a `video` policy, so this override means every run of the spec — including re-runs and retries — writes a video file to `test-results/`. That directory is in `.gitignore` so the files won't be committed, but storage costs and CI time accumulate.

Prefer `video: "retain-on-failure"` which records but only keeps the file when the test actually fails, matching the existing `screenshot: "only-on-failure"` policy in the base config.

Also note: `blob-report/` is not currently in `.gitignore` (only `/playwright-report` and `/test-results` are). If the CI reporter is ever switched to `blob`, those files could leak into commits.

---

src/components/tasks/task-item.tsx
```diff
@@ -11,24 +11,68 @@
...
 
 export function TaskItem({ task, onToggle, disabled = false }: TaskItemProps) {
   const isCompleted = task.status === "completed";
   const assignees = task.assignments;
+  const points = usePointsOptional();
+
+  const [animationPoints, setAnimationPoints] = useState<number | null>(null);
+
+  const handleToggle = async () => {
+    const wasCompleted = isCompleted;
```
BenSeymourODB, line 49: **`wasCompleted` snapshot is correct — but the guard relies on the task prop staying stale between the await.**

`wasCompleted` captures `isCompleted` (derived from `task.status`) synchronously before `await Promise.resolve(onToggle())`. This is correct because the parent controls `task` and typically won't push a new prop into `TaskItem` during the microtask gap. However, if `onToggle` is async and triggers a state update that causes React to re-render `TaskItem` with the new `task.status` before the resumed code runs, `wasCompleted` still holds the pre-toggle value from the closure — so the guard is fine.

Worth a one-line code comment explaining this invariant so future maintainers don't "fix" it:

```ts
// wasCompleted is captured before the async toggle so we always know
// which direction the transition ran — React may re-render TaskItem
// with the updated task.status before this closure resumes.
const wasCompleted = isCompleted;
```

---

src/components/rewards/points-context.tsx
```diff
@@ -127,13 +148,16 @@ export function PointsProvider({ profileId, children }: PointsProviderProps) {
         setFetchedEnabled(false);
         return;
       }
-      const data = (await response.json()) as {
-        totalPoints: number;
-        enabled: boolean;
-      };
+      const data = (await response.json()) as PointsApiResponse;
```
BenSeymourODB, line 150: **CLAUDE.md violation: `useCallback` is manual memoization and is explicitly forbidden.**

CLAUDE.md says: "DO NOT use manual memoization (`useMemo`/`useCallback`/`React.memo`). Write clean, simple React code." The React Compiler (enabled in `next.config.ts`) handles memoization automatically.

Both `awardPoints` and `refreshPoints` (this function) are wrapped in `useCallback` — these were pre-existing on this file but since this PR is editing the same callbacks heavily it's a good time to clean them up. Both should be plain `async` functions inside the provider component body and `useCallback` removed from the React import.

```diff
-  const refreshPoints = useCallback(async () => {
+  const refreshPoints = async () => {
     ...
-  }, [profileId]);
+  };

-  const awardPoints = useCallback(
+  const awardPoints = async (
     ...
-  }, [profileId]);
+  );
```

The existing test suite exercises both functions thoroughly — no behaviour changes, only wrapping removal. Tracking issue: #271 (repo-wide audit).

---

BenSeymourODB (commented):
First-round review from automated subagent. Five findings — see inline comments. Will follow up after addressing.

---

src/components/rewards/points-animation.tsx
BenSeymourODB: Fixed in 77f6535. Adopted the ref pattern; the effect now depends only on `show`. ESLint's `react-hooks/refs` rule rejected the inline `onCompleteRef.current = onComplete;` during render, so the ref is synced via a separate `useEffect([onComplete])` instead — same outcome, idiomatic with the new rule. All animation tests still pass.

---

src/components/rewards/points-animation.tsx
BenSeymourODB: Keeping `polite`. The +N banner is celebratory feedback, not a system message — interrupting the user's screen-reader output mid-utterance for a non-urgent +10 points pop-in is more disruptive than the worst-case `polite` outcome (the announcement gets dropped entirely if they're typing). MDN's WAI-ARIA guidance specifically suggests `assertive` for "important, and usually time-sensitive" content (errors, alerts) — a reward animation doesn't clear that bar.

The badge total in the persistent header is also `role="status"` (implicitly polite) — assistive tech users still get an authoritative source of truth for their points without the celebratory interrupt. Open to revisiting if real users find this surprising.

---

e2e/reward-points.spec.ts
BenSeymourODB: Keeping `video: "on"` for this spec. The session prompt explicitly instructed "For animations or screen navigation: enable video capture in the Playwright config (`use: { video: 'on' }` for the test) and design tests that exercise the animation/navigation path" — the +N animation is exactly that case, and the captured clip is intended for the PR description. The artifact lives under `test-results/` (gitignored) and the spec runs as a single file, so the cost is bounded to one file per CI run of this spec.

Fair point on `blob-report/` — it's not in `.gitignore` today. Worth adding either way; will note as a separate chore (small standalone PR rather than pile on here).

---

src/components/tasks/task-item.tsx
BenSeymourODB: Done in 77f6535 — added the comment as suggested.

---

src/components/rewards/points-context.tsx
BenSeymourODB: Fixed in 77f6535. Removed `useCallback` from both `awardPoints` and `refreshPoints` and dropped it from the React import. All 27 rewards tests still pass and the full vitest suite is green (132 files, 2044 tests).

The rest of `points-context.tsx` (the `useEffect` for AbortController, the `useState`/`useRef` calls) is fine — those aren't manual memoization. Repo-wide audit still belongs under #271.

---

PR #348, opened 2026-05-07
fix(auth): accept AUTH_SECRET + classify transient/terminal refresh errors (#315)

BenSeymourODB:
## Summary

Closes #315 (Parts A + B).

The "Session expired. Please sign in again." toast firing every ~1hr on a healthy device traces back to a **deployment misconfig that silently degraded into a per-request 401**:

- `token-cipher.resolveKey()` reads only `NEXTAUTH_SECRET`, but a NextAuth v5 project's `.env.local` sets `AUTH_SECRET`.
- The dev fallback never engaged → `encryptToken()` threw inside the session callback's refresh branch → the broad `catch` set `RefreshTokenError` → every API route returned 401 with the user-visible "Session expired" string.
- A second class of bug: even with the key configured correctly, **any** transient error (network blip, rate limit, decrypt hiccup) followed the same path and forced re-auth.

## Scope

- **Part A** — fail-fast on missing/invalid encryption key; accept the NextAuth v5 idiom.
- **Part B** — distinguish transient from terminal refresh failures so the session is only invalidated on a genuine Google revocation.

### Deferred (filed as follow-up issues from this PR)

- **Part C** — `Account.consecutive_refresh_failures` + exponential backoff + offline grace window. Requires a Prisma migration; out of scope here.
- **Part D** — `/api/health/auth` operator endpoint + AccountManager admin banner.

## Phase 1 — `fix(auth): accept AUTH_SECRET fallback + fail fast on missing token-encryption key`

- `resolveKey()` reads `AUTH_SECRET` (v5 canonical) in addition to `NEXTAUTH_SECRET`. `AUTH_SECRET` wins when both are set so v5 projects don't keep deriving from a stale v4 value.
- Error messages enumerate `AUTH_SECRET` / `NEXTAUTH_SECRET` / `TOKEN_ENCRYPTION_KEY` so devs on either NextAuth idiom recognise the gap.
- New `validateEncryptionKey()` exported from `token-cipher`; `auth.ts` calls it at module init so a missing/invalid key surfaces as an immediate boot failure instead of degrading into the silent per-request encrypt failure mode.
- `.env.local.example` now documents `AUTH_SECRET` and `TOKEN_ENCRYPTION_KEY` (both were previously absent).

## Phase 2 — `fix(auth): classify transient vs terminal refresh errors + guard build-phase validation`

- New `classifyTokenRefreshError(err): 'terminal' | 'transient'` in `src/lib/auth/refresh-error-classifier.ts`.
- Terminal classification limited to a **closed allow-list** of Google OAuth error codes (`invalid_grant`, `invalid_client`, `unauthorized_client`). Every other error class — `rate_limit_exceeded`, 5xx, network failure, decrypt failure, unknown shape — is classified `transient` and defaults to `transient` for anything unrecognised.
- Session callback now uses the classifier: terminal → log `TokenRefreshFailed` + set `session.error = "RefreshTokenError"` (legacy behaviour, intended); transient → log `TokenRefreshTransientFailure` and leave the session untouched so the next callback retries. Upstream API routes continue to serve cached data via the IndexedDB fallback during transient outages.
- **CI build fix**: Phase 1's eager `validateEncryptionKey()` at module init crashed `next build` because page-data collection evaluates `auth.ts` without runtime env vars. Now guarded behind `NEXT_PHASE !== "phase-production-build"`, preserving fail-fast in production runtime while making the build pass.

## Test plan

- [x] New unit tests for the AUTH_SECRET fallback, error-message wording, `validateEncryptionKey()`.
- [x] New unit tests for `classifyTokenRefreshError` covering each Google error code, network errors, decrypt failures, non-Error thrown values, and the safe-default invariant (13 tests).
- [x] `pnpm test` — 132 files / 2037 tests green.
- [x] `pnpm build` — green locally (the previous CI failure on the build step is what motivated the `NEXT_PHASE` guard).
- [ ] Manual: `AUTH_SECRET` set, `TOKEN_ENCRYPTION_KEY` unset → boots, encrypt round-trips. Both unset in dev → `auth.ts` import throws a clear error mentioning `AUTH_SECRET`.

🤖 Authored by a scheduled Claude Code agent run.

---
_Generated by [Claude Code](https://claude.ai/code/session_01PYF6pG9pKvHBkoxRbeZ8K9)_


src/lib/auth/refresh-session-tokens.ts
```diff
@@ -0,0 +1,156 @@
...
+        context: "TokenRefreshFailed",
+        userId,
+      });
+      return { kind: "terminal-error", error: error as Error };
+    }
+    deps.logger.error(error as Error, {
+      context: "TokenRefreshTransientFailure",
+      userId,
+    });
+    return { kind: "transient-error", error: error as Error };
```
BenSeymourODB: **Nit (defense in depth):** `error as Error` here (and on L103 + L144) is a lie if a non-Error value ever propagates out of `decryptToken`, `refreshGoogleAccessToken`, or `prisma.account.update`. Every current path throws Error subclasses, so this is correct today — but `logger.error()` types `error: Error` and will silently read `.message`/`.stack` as `undefined` if a primitive sneaks in.

Cheap belt-and-braces fix: normalise once at the catch site:
```ts
const err = error instanceof Error ? error : new Error(String(error));
```
and pass `err` to both `classifyTokenRefreshError` and `logger.error`. The classifier already handles non-Error inputs (`returns "transient"`), but this guarantees the logger gets a real Error too. Skip if you'd rather not widen the scope of the PR.

---

src/lib/auth/refresh-session-tokens.ts
```diff
@@ -0,0 +1,156 @@
...
+      },
+      where: {
+        provider_providerAccountId: {
+          provider: "google",
+          providerAccountId: account.providerAccountId,
+        },
+      },
+    });
+
+    deps.logger.event("TokenRefreshed", { userId, success: true });
```
BenSeymourODB: **Nit:** `success: true` on the `TokenRefreshed` event is redundant — the event name already conveys success, and no failure path emits this event. Dropping it would keep the event payload minimal and slightly easier to query in App Insights. Trivial; ignore if the convention elsewhere in the codebase already uses a `success` flag.

---

src/lib/auth/refresh-error-classifier.ts
```diff
@@ -0,0 +1,86 @@
...
+// next session callback can retry. See #315 for the failure mode this is
+// defending against — any transient error silently forcing a 1-hour re-auth.
+const TERMINAL_GOOGLE_ERROR_CODES: ReadonlySet<string> = new Set([
+  "invalid_grant",
+  "invalid_client",
+  "unauthorized_client",
+  // Misconfiguration on our side — retrying with the same grant_type can never
+  // succeed, so force re-auth rather than silently never refreshing.
+  "unsupported_grant_type",
+]);
```
BenSeymourODB, line 46: **Discussion (non-blocking) — `unsupported_grant_type` classification:**

Treating `unsupported_grant_type` as terminal forces re-auth on what is always a code/config bug on our side (we control the `grant_type` we send). Re-auth can never fix it — the next refresh attempt will send the same wrong `grant_type` and fail the same way, so the user ends up in a re-auth loop until the operator deploys a fix.

Two safer options long-term:
1. Classify as **transient** (default behavior) so the failure is loud in logs but not user-facing. Operator sees the alert; user keeps cached UI.
2. Once Part D (`/api/health/auth` + AccountManager admin banner per #315) lands, add an **operator-terminal** classification that surfaces in the admin banner without invalidating user sessions.

Not a blocker — current behavior is no worse than the pre-PR baseline, and the comment makes the intent clear. Worth a follow-up issue alongside Parts C/D.

---

src/lib/auth/auth.ts
```diff
@@ -1,13 +1,32 @@
...
+// user sees as "Session expired. Please sign in again." every hour even
+// though the underlying problem is a server-side env-var gap (#315).
+//
+// Skipped during `next build`: page-data collection evaluates this module
+// without runtime env vars and a throw here aborts the build. Production
+// runtime still validates eagerly on server start (NEXT_PHASE is unset or
+// "phase-production-server").
+if (process.env.NEXT_PHASE !== "phase-production-build") {
+  validateEncryptionKey();
+}
```
BenSeymourODB, line 28: **Observation (not a blocker):** the `NEXT_PHASE` guard is the right escape hatch and I confirmed the build passes locally with no auth env vars set. One thing to keep in mind: `NEXT_PHASE` is set by Next.js's own build/start commands. If anything ever loads `auth.ts` outside of those (custom server, worker script, ad-hoc Node REPL using the compiled module), the eager validation will fire — which is the desired behavior for the misconfig case #315 fixes, but worth a one-line note in `.claude/plans/auth-session-expired-315.md` so future contributors don't see the guard and assume it gates **all** non-test paths.

---

src/lib/auth/auth.ts
```diff
@@ -90,80 +108,29 @@ export const { handlers, signIn, signOut, auth } = NextAuth({
...
+        user.id,
+        googleAccount,
+        {
+          prisma,
+          refreshGoogleAccessToken,
+          encryptToken,
+          decryptToken,
+          logger,
+          googleClientId: process.env.GOOGLE_CLIENT_ID!,
+          googleClientSecret: process.env.GOOGLE_CLIENT_SECRET!,
```
BenSeymourODB, line 126: **Out-of-scope but worth flagging:** `process.env.GOOGLE_CLIENT_ID!` / `GOOGLE_CLIENT_SECRET!` are non-null assertions. If either is missing at runtime, Google returns `invalid_client` → classifier marks it terminal → user is force-logged-out for what is actually a server misconfig. The new `validateEncryptionKey()` boot check rules out the encryption-key class of misconfig; it would be nice to add a parallel assertion for the OAuth client credentials behind the same `NEXT_PHASE` guard so both misconfigs surface at boot rather than mid-session. Pure follow-up — not a regression introduced by this PR.

---

BenSeymourODB (commented):
## Independent in-depth review

Verified locally on `fd06d52`:
- `pnpm test:run` — 133 files / 2054 tests green
- `pnpm lint` — 0 errors, 6 pre-existing warnings (unrelated)
- `pnpm format` — clean
- `pnpm check-types` — clean
- `pnpm build` — green **with all auth env vars unset**, confirming the `NEXT_PHASE === "phase-production-build"` guard at `auth.ts:26` correctly short-circuits the eager `validateEncryptionKey()` during page-data collection (the original Phase 2 motivation)

### Architecture & correctness

The two-phase split is clean and the discriminated `RefreshOutcome` lets `auth.ts` stay a thin shim. I traced every error class through the catch in `refresh-session-tokens.ts` and the classifier behavior matches both the plan and the issue's intent:

- **Terminal** → `MissingRefreshTokenError`, `RefreshTokenDecryptError`, and Google `invalid_grant` / `invalid_client` / `unauthorized_client` / `unsupported_grant_type`. All four Google codes are correct revocation/permanent-state signals; the two internal sentinels cover the "ciphertext can't be read" + "row never had a token" cases the Sonnet self-review caught.
- **Transient** → everything else: 5xx, `rate_limit_exceeded`, `slow_down`, `internal_error`, `temporarily_unavailable`, raw `TypeError` from `fetch`, non-Error thrown values, unknown shapes. Safe-default semantics are well-tested in `refresh-error-classifier.test.ts:120-137`.

The "no DB write on transient" invariant (`session-callback.test.ts:260-268`) preserves last-known-good state so a stutter doesn't permanently strand the user with a stale `expires_at` rewritten to a bogus value.

I also checked test isolation: every test file that imports `@/lib/auth` either calls `vi.mock("@/lib/auth", ...)` (route tests) or `vi.mock("@/lib/auth/auth", ...)` + `vi.stubEnv("TOKEN_ENCRYPTION_KEY", ...)` (`helpers.test.ts`), so the eager `validateEncryptionKey()` never fires in the test suite. That's why CI stays green without env stubbing at the suite level.

### `AUTH_SECRET` precedence test (token-cipher.test.ts:209-245)

The new test that addresses self-review finding #4 is correct and convincing: encrypt with `AUTH_SECRET=X`, swap to `NEXTAUTH_SECRET=Y`, prove the envelope **fails** to decrypt — i.e. `resolveKey()` did **not** silently fall back to `NEXTAUTH_SECRET` for the encrypt. The inverse case (`AUTH_SECRET=X` + `NEXTAUTH_SECRET=Y` set together, round-trip works) closes the loop. Much stronger than the prior same-value symmetry assertion.

### Cross-PR note (informational, not a blocker)

PR #252 (`feat(auth): singleflight Google OAuth token-refresh queue`) wraps refresh in a singleflight queue but doesn't touch the session callback. After both merge, the helper here (`refreshGoogleSessionTokensIfNeeded`) won't route through the queue — concurrent session callbacks across requests would still issue parallel refresh fetches under load. Whichever PR lands second, the integration is small: pass a `refreshFn` dep that's the queue-wrapped version. No conflict at the file level, just a coordination note.

### Polish-level inline comments

Five inline comments above. All non-blocking:
- (`refresh-session-tokens.ts:103/144/154`) `error as Error` cast — defense in depth
- (`refresh-session-tokens.ts:139`) redundant `success: true` field
- (`refresh-error-classifier.ts:46`) `unsupported_grant_type` semantics → re-auth loop on a code bug
- (`auth.ts:28`) `NEXT_PHASE` guard scope note for the plan doc
- (`auth.ts:126`) parallel boot validation for OAuth client creds (follow-up)

### Verdict

Ready to merge as-is. The PR closes a real user-visible bug, the test coverage is thorough, and the design correctly limits terminal classification to a closed allow-list (the exact failure mode #315 calls out). Approving review-wise; deferring the merge button to you since I'm not a code owner.

---
🤖 Independent review by `claude-opus-4-7` via `/implement-issue → PR-review fallback`

---

src/lib/auth/refresh-session-tokens.ts
BenSeymourODB: Done in a193f6b. Normalised once at each catch site — both the inner decrypt `catch` and the outer classify/log `catch` now do `const err = error instanceof Error ? error : new Error(String(error))` and pass `err` to `classifyTokenRefreshError` and `logger.error` (and into the returned `RefreshOutcome.error`). The classifier still safe-defaults non-Error inputs to `transient`; this just guarantees `logger.error` always gets a real Error with usable `.message`/`.stack`. Existing tests still pass since every test throws an `Error` subclass (so `err === error` and the `toHaveBeenCalledWith`/`toEqual` identity assertions hold).

---

src/lib/auth/refresh-session-tokens.ts
BenSeymourODB: Done in a193f6b. Dropped `success: true` — the `TokenRefreshed` event is only emitted on the success path, so the flag was redundant. Payload is now just `{ userId }`. Updated the corresponding assertion in `session-callback.test.ts`.

---

src/lib/auth/refresh-error-classifier.ts
BenSeymourODB: Agreed it's worth revisiting — the re-auth-loop reasoning is sound. Filed as a follow-up: #378. Leaving the current `terminal` classification in place for this PR (it was a deliberate Sonnet-self-review finding and is no worse than the pre-PR baseline), and tracking the transient-vs-operator-terminal decision alongside Parts C/D there.

---

src/lib/auth/auth.ts
BenSeymourODB: Done in a193f6b. Added an "Implementation notes" section to `.claude/plans/auth-session-expired-315.md` spelling out that the `NEXT_PHASE === "phase-production-build"` guard only skips validation during `next build` page-data collection, and that anything loading `auth.ts` outside Next.js's own build/start commands (custom server, worker, ad-hoc REPL on the compiled module) still fires the eager validation by design — so future contributors don't assume it gates all non-test paths.

---

src/lib/auth/auth.ts
BenSeymourODB: Good catch on the parallel misconfig path. Filed as a follow-up: #379. Keeping it out of this PR since it's a pure follow-up (not a regression here), but tracked so the OAuth client creds get the same boot-time fail-fast treatment as the encryption key, behind the same `NEXT_PHASE` guard.

---

PR #349, opened 2026-05-07
feat(tasks): add TasksProvider for shared client state (#331)

BenSeymourODB:
Closes #331.

## Summary

Adds `TasksProvider` — the Calendar-style shared client-state container the Tasks track has been waiting on. It is the foundation for #329 (`/tasks` page), #332 (TaskListConfig CRUD UI), and #333 (TasksSettingsPanel), all of which need a single source for "the currently rendered config" and the `useTasks` lifecycle.

## What it owns (ephemeral UI state only)

- **`activeConfig`** — resolved by id from a `configs: TaskListConfig[]` prop with a stale-id fallback (persisted id → first available → null). Avoids leaving the surface wedged when a config is deleted.
- **`viewMode`** — `"list" | "grouped-by-list"`, persisted to `localStorage` under `tasks.viewMode`.
- **`lists`** — enabled subset of the active config.
- **`useTasks(activeConfig)` pass-through** — `tasks`, `loading`, `error`, `refreshTasks`, `updateTask`. Descendants share one fetch lifecycle instead of each calling `useTasks` and racing on the same config.

## What it deliberately does not own

DB-backed settings (`taskSortOrder`, `showCompletedTasks`, `defaultTaskListId`) keep flowing through `useUserSettings` / `ProfileSettings` — the provider is for ephemeral UI state, not a second source of truth. This matches the issue spec.

## API

```tsx
<TasksProvider configs={configs} initialActiveConfigId={...} initialViewMode={...}>
  <TaskList />
</TasksProvider>

const {
  configs, activeConfig, setActiveConfigId,
  lists, viewMode, setViewMode,
  tasks, loading, error, refreshTasks, updateTask,
} = useTasksContext();
```

`useTasksContext()` throws when used outside the provider so misuse fails loudly.

## React Compiler compliance

No `useCallback` / `useMemo` / `React.memo` per `docs/react-compiler.md` — the issue spec calls this out and the file follows it.

## Tests

12 new unit tests in `src/components/providers/__tests__/TasksProvider.test.tsx`:
- `useTasksContext` throws outside provider
- empty-`configs` returns `null` activeConfig + empty `lists`
- defaults activeConfig to first when nothing is persisted
- rehydrates activeConfigId from localStorage on mount
- falls back to first when persisted id is stale
- computes `lists` as enabled subset of activeConfig
- `setActiveConfigId` switches and persists to localStorage
- defaults `viewMode` to `"list"`
- `setViewMode` toggles and persists
- rehydrates `viewMode` from localStorage on mount
- `useTasks` is called with the active config and its return values are passed through (incl. `null` when configs is empty)

Full repo suite: 2031 / 2031 passing. Lint, format, and check-types clean.

## Deferred to sibling sub-issues

- Wiring DB-backed `configs` source — #332 (TaskListConfig CRUD UI). Provider currently accepts `configs` as a prop so the eventual repository hook can drop straight in.
- Mounting on a real page — #329 (`/tasks` route).
- Settings popover consumer — #333.
- Persisting `taskSortOrder` / `showCompletedTasks` to `ProfileSettings` — #334.

## Test plan
- [x] `pnpm test` — 2031 / 2031
- [x] `pnpm lint:fix` — no new errors
- [x] `pnpm format:fix` — no diff

https://claude.ai/code/session_019aDuo26G6JMTEjwrgUR8vH

---
_Generated by [Claude Code](https://claude.ai/code/session_019aDuo26G6JMTEjwrgUR8vH)_

### DIFF
--- a/src/components/providers/TasksProvider.tsx
+++ b/src/components/providers/TasksProvider.tsx
@@ -31,7 +31,7 @@ import type {
 import { type UseTasksReturn, useTasks } from "@/components/tasks/use-tasks";
 import { useLocalStorage } from "@/hooks/useLocalStorage";
 import type React from "react";
-import { createContext, useContext } from "react";
+import { createContext, useContext, useEffect } from "react";
 
 /** Tasks UI view mode. Mirrors the Calendar `view` axis. */
 export type TasksViewMode = "list" | "grouped-by-list";
@@ -43,6 +43,14 @@ export const TASKS_ACTIVE_CONFIG_LS_KEY = "tasks.activeConfigId";
 export const TASKS_VIEW_MODE_LS_KEY = "tasks.viewMode";
 
 const DEFAULT_VIEW_MODE: TasksViewMode = "list";
+const VALID_VIEW_MODES: ReadonlySet<string> = new Set([
+  "list",
+  "grouped-by-list",
+]);
+
+function isValidViewMode(value: unknown): value is TasksViewMode {
+  return typeof value === "string" && VALID_VIEW_MODES.has(value);
+}
 
 export interface ITasksContext extends UseTasksReturn {
   /**
@@ -111,38 +119,86 @@ export function TasksProvider({
     TASKS_ACTIVE_CONFIG_LS_KEY,
     initialActiveConfigId ?? null
   );
-  const [viewMode, setStoredViewMode] = useLocalStorage<TasksViewMode>(
+  const [storedViewMode, setStoredViewMode] = useLocalStorage<TasksViewMode>(
     TASKS_VIEW_MODE_LS_KEY,
     initialViewMode ?? DEFAULT_VIEW_MODE
   );
+
+  // Apply override props on mount when they're provided. `useLocalStorage`
+  // is backed by `useSyncExternalStore`, so the `initialValue` argument is
+  // only used when the storage key is absent; for a returning user with a
+  // persisted value the override would be silently ignored without this
+  // mount-only write through. Mirrors `CalendarProvider.initialView` /
+  // `initialAgendaMode` (#150 / #238). Capturing the props once is
+  // intentional: subsequent prop changes are owned by the parent
+  // component (e.g. it remounts the provider on a URL change).
+  useEffect(() => {
+    if (initialActiveConfigId !== undefined) {
+      setStoredConfigId(initialActiveConfigId);
+    }
+    if (initialViewMode !== undefined) {
+      setStoredViewMode(initialViewMode);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
 
   // Resolve `activeConfig` with a stale-id fallback: if the persisted id
   // points at a config the user has since deleted (or to nothing on the
   // very first mount), fall through to the first available config rather
   // than rendering an empty surface that "should" have content.
-  const activeConfig =
-    configs.find((c) => c.id === storedConfigId) ?? configs[0] ?? null;
+  const resolvedConfig = configs.find((c) => c.id === storedConfigId);
+  const activeConfig = resolvedConfig ?? configs[0] ?? null;
 
-  const setActiveConfigId = (id: string) => {
-    setStoredConfigId(id);
-  };
+  // Reconcile a stale `storedConfigId` so the next reload starts on the
+  // correct config without re-firing the fallback every render. Skipped
+  // when `configs` is empty (no canonical id to write through yet).
+  useEffect(() => {
+    if (
+      storedConfigId !== null &&
+      !resolvedConfig &&
+      activeConfig !== null &&
+      activeConfig.id !== storedConfigId
+    ) {
+      setStoredConfigId(activeConfig.id);
+    }
+  }, [storedConfigId, resolvedConfig, activeConfig, setStoredConfigId]);
 
-  const setViewMode = (mode: TasksViewMode) => {
-    setStoredViewMode(mode);
-  };
+  // Validate the stored view mode against the current union. A
+  // corrupted/manually-edited value (or one written by a future build
+  // and read by an older one) is typed as `TasksViewMode` at compile
+  // time but might not match the union at runtime; fall back to the
+  // default rather than render with a value the rest of the codebase
+  // doesn't recognise.
+  const viewMode: TasksViewMode = isValidViewMode(storedViewMode)
+    ? storedViewMode
+    : DEFAULT_VIEW_MODE;
 
   const lists = activeConfig ? activeConfig.lists.filter((l) => l.enabled) : [];
 
-  const tasksApi = useTasks(activeConfig);
+  // Destructure the exact `useTasks` keys we forward instead of spreading,
+  // so a future addition to `UseTasksReturn` whose name collides with a
+  // provider-owned key (e.g. both expose `loading`) is caught by the type
+  // checker at the assignment below rather than silently shadowing the
+  // provider's value at runtime.
+  const { tasks, loading, error, refreshTasks, updateTask } =
+    useTasks(activeConfig);
 
+  // Setters are exposed directly — they don't add type narrowing,
+  // logging, or side effects, so wrapping them only adds a closure that
+  // the React Compiler can't reason about. The public types in
+  // `ITasksContext` constrain callers to the intended argument shape.
   const value: ITasksContext = {
-    ...tasksApi,
+    tasks,
+    loading,
+    error,
+    refreshTasks,
+    updateTask,
     configs,
     activeConfig,
-    setActiveConfigId,
+    setActiveConfigId: setStoredConfigId,
     lists,
     viewMode,
-    setViewMode,
+    setViewMode: setStoredViewMode,
   };
 
   return (
--- a/src/components/providers/TasksProvider.tsx
+++ b/src/components/providers/TasksProvider.tsx
@@ -1,0 +1,151 @@
+"use client";
+
+/**
+ * TasksProvider — shared client-state container for the Tasks surface
+ * (issue #331). Analogous to `CalendarProvider` for the calendar surface.
+ *
+ * Owns ephemeral UI state only:
+ *
+ *   - `activeConfig` — the `TaskListConfig` currently being rendered.
+ *     Resolved by id against the `configs` prop with a stale-id fallback,
+ *     so a removed config never leaves the surface in a wedged "no view"
+ *     state.
+ *   - `viewMode` — `"list"` or `"grouped-by-list"`, persisted across
+ *     reloads.
+ *
+ * DB-backed settings (`taskSortOrder`, `showCompletedTasks`,
+ * `defaultTaskListId`) deliberately stay with `useUserSettings` /
+ * `ProfileSettings` and are NOT mirrored here — the provider is for
+ * ephemeral UI state, not a second source of truth.
+ *
+ * The provider also calls `useTasks(activeConfig)` once and re-exposes
+ * its `{ tasks, loading, error, refreshTasks, updateTask }` so descendants
+ * (`TaskList`, `TasksSettingsPanel`, future `TasksPage`) read from one
+ * canonical fetch lifecycle instead of each calling `useTasks` and racing
+ * each other on the same config.
+ */
+import type {
+  TaskListConfig,
+  TaskListSelection,
+} from "@/components/tasks/types";
+import { type UseTasksReturn, useTasks } from "@/components/tasks/use-tasks";
+import { useLocalStorage } from "@/hooks/useLocalStorage";
+import type React from "react";
+import { createContext, useContext } from "react";
+
+/** Tasks UI view mode. Mirrors the Calendar `view` axis. */
+export type TasksViewMode = "list" | "grouped-by-list";
+
+/** localStorage key for the user's last-active `TaskListConfig` id. */
+export const TASKS_ACTIVE_CONFIG_LS_KEY = "tasks.activeConfigId";
+
+/** localStorage key for the user's last-selected view mode. */
+export const TASKS_VIEW_MODE_LS_KEY = "tasks.viewMode";
+
+const DEFAULT_VIEW_MODE: TasksViewMode = "list";
+
+export interface ITasksContext extends UseTasksReturn {
+  /**
+   * All known `TaskListConfig` rows the provider can pick from. The
+   * provider does not fetch these itself — the caller is expected to
+   * supply them (for now hard-coded; #332 will populate from the DB).
+   */
+  configs: TaskListConfig[];
+  /** The currently-rendered config. `null` when `configs` is empty. */
+  activeConfig: TaskListConfig | null;
+  /**
+   * Switch the active config. Persists the id to `localStorage` so the
+   * choice survives a reload.
+   */
+  setActiveConfigId: (id: string) => void;
+  /**
+   * Effective enabled lists from `activeConfig`. Empty when there is
+   * no active config or every list is disabled.
+   */
+  lists: TaskListSelection[];
+  /** Ephemeral UI view mode. */
+  viewMode: TasksViewMode;
+  /** Switch the view mode. Persists to `localStorage`. */
+  setViewMode: (mode: TasksViewMode) => void;
+}
+
+const TasksContext = createContext<ITasksContext | null>(null);
+
+/**
+ * Read the Tasks context. Throws if called outside `TasksProvider` so
+ * mis-mounted descendants fail loudly instead of silently no-op'ing.
+ */
+export function useTasksContext(): ITasksContext {
+  const ctx = useContext(TasksContext);
+  if (!ctx) {
+    throw new Error("useTasksContext must be used within TasksProvider");
+  }
+  return ctx;
+}
+
+export interface TasksProviderProps {
+  children: React.ReactNode;
+  /**
+   * The set of `TaskListConfig` rows the user can pick from. Defaults
+   * to `[]` so the provider can mount before any configs exist (e.g.
+   * the empty-state path on `/tasks`).
+   */
+  configs?: TaskListConfig[];
+  /**
+   * Optional override for the initial active config id. Used by deep
+   * links (e.g. `/tasks?listConfig=...`) to win over the persisted
+   * value on first mount.
+   */
+  initialActiveConfigId?: string;
+  /** Optional override for the initial view mode. Same semantics. */
+  initialViewMode?: TasksViewMode;
+}
+
+export function TasksProvider({
+  children,
+  configs = [],
+  initialActiveConfigId,
+  initialViewMode,
+}: TasksProviderProps) {
+  const [storedConfigId, setStoredConfigId] = useLocalStorage<string | null>(
+    TASKS_ACTIVE_CONFIG_LS_KEY,
+    initialActiveConfigId ?? null
+  );
+  const [viewMode, setStoredViewMode] = useLocalStorage<TasksViewMode>(
+    TASKS_VIEW_MODE_LS_KEY,
+    initialViewMode ?? DEFAULT_VIEW_MODE
+  );
+
+  // Resolve `activeConfig` with a stale-id fallback: if the persisted id
+  // points at a config the user has since deleted (or to nothing on the
+  // very first mount), fall through to the first available config rather
+  // than rendering an empty surface that "should" have content.
+  const activeConfig =
+    configs.find((c) => c.id === storedConfigId) ?? configs[0] ?? null;
+
+  const setActiveConfigId = (id: string) => {
+    setStoredConfigId(id);
+  };
+
+  const setViewMode = (mode: TasksViewMode) => {
+    setStoredViewMode(mode);
+  };
+
+  const lists = activeConfig ? activeConfig.lists.filter((l) => l.enabled) : [];
+
+  const tasksApi = useTasks(activeConfig);
+
+  const value: ITasksContext = {
+    ...tasksApi,
+    configs,
+    activeConfig,
+    setActiveConfigId,
+    lists,
+    viewMode,
+    setViewMode,
+  };
+
+  return (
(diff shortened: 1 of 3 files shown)


src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+  /** Optional override for the initial view mode. Same semantics. */
+  initialViewMode?: TasksViewMode;
+}
+
+export function TasksProvider({
+  children,
+  configs = [],
+  initialActiveConfigId,
+  initialViewMode,
+}: TasksProviderProps) {
```
BenSeymourODB, line 117: **Bug: `initialActiveConfigId` is silently ignored when localStorage already holds a value.**

`useLocalStorage` (backed by `useSyncExternalStore`) reads the stored value on the client, so the `initialValue` argument is only used when the key is absent. If the user has a previously-persisted config id, `initialActiveConfigId` is never applied.

`CalendarProvider` handles this with a mount-only `useEffect` that writes the override through to storage when `initialView` is defined. The same pattern is needed here — otherwise deep-links like `/tasks?listConfig=xxx` will work on a fresh browser but silently no-op on a returning visitor.

```tsx
useEffect(() => {
  if (initialActiveConfigId !== undefined) {
    setStoredConfigId(initialActiveConfigId);
  }
}, []);
```

---

src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const [storedConfigId, setStoredConfigId] = useLocalStorage<string | null>(
+    TASKS_ACTIVE_CONFIG_LS_KEY,
+    initialActiveConfigId ?? null
+  );
+  const [viewMode, setStoredViewMode] = useLocalStorage<TasksViewMode>(
+    TASKS_VIEW_MODE_LS_KEY,
+    initialViewMode ?? DEFAULT_VIEW_MODE
+  );
```
BenSeymourODB, line 125: **Redundant wrapper functions recreated on every render — no-value wrappers.**

`setActiveConfigId` and `setViewMode` (lines 117-123) are inline lambdas that do nothing but delegate to the setter from `useLocalStorage`. They are recreated on every render (React Compiler won't memoize closures that capture nothing special here), and they don't add type narrowing or side-effects. Either expose the setters directly in the context value, or only introduce wrappers if/when they gain additional logic (e.g. logging, validation).

```tsx
// Instead of:
const setActiveConfigId = (id: string) => { setStoredConfigId(id); };

// Just use:
// setActiveConfigId: setStoredConfigId,  // in the value object
```

This also applies to `setViewMode`.

---

src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+
+  // Resolve `activeConfig` with a stale-id fallback: if the persisted id
+  // points at a config the user has since deleted (or to nothing on the
+  // very first mount), fall through to the first available config rather
+  // than rendering an empty surface that "should" have content.
+  const activeConfig =
+    configs.find((c) => c.id === storedConfigId) ?? configs[0] ?? null;
+
+  const setActiveConfigId = (id: string) => {
+    setStoredConfigId(id);
```
BenSeymourODB: **Stale `storedConfigId` after the fallback fires is never reconciled.**

When the persisted id no longer matches any config (the stale-id fallback), `activeConfig` correctly resolves to `configs[0]`, but `storedConfigId` in localStorage still holds the old id. On every subsequent render the `find` fails and the fallback fires again, so the user always starts on the first config regardless of what they later switch to — **their subsequent `setActiveConfigId` calls _will_ persist correctly, but only after they actively switch**. 

A one-time reconciliation write when the fallback fires would remove the inconsistency:
```tsx
const resolvedConfig = configs.find((c) => c.id === storedConfigId);
const activeConfig = resolvedConfig ?? configs[0] ?? null;

// Reconcile stale id so the next reload starts on the correct config
if (!resolvedConfig && activeConfig && storedConfigId !== null) {
  setStoredConfigId(activeConfig.id);
}
```
Worth deciding intentionally either way and adding a comment.

---

src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+  children,
+  configs = [],
+  initialActiveConfigId,
+  initialViewMode,
+}: TasksProviderProps) {
+  const [storedConfigId, setStoredConfigId] = useLocalStorage<string | null>(
+    TASKS_ACTIVE_CONFIG_LS_KEY,
+    initialActiveConfigId ?? null
+  );
+  const [viewMode, setStoredViewMode] = useLocalStorage<TasksViewMode>(
```
BenSeymourODB: **No guard against a corrupted/future `viewMode` value in localStorage.**

`useLocalStorage` will return whatever string is stored verbatim. If a user has a stale value (e.g. a third mode added in a future PR is later removed, or a typo from manual editing), `viewMode` will be typed as `TasksViewMode` at compile time but hold an invalid value at runtime.

`CalendarProvider` uses `migrateLegacySettings()` for exactly this pattern. A simple validation narrowing (or a reset to default when the value doesn't match the union) would harden this:
```tsx
const VALID_VIEW_MODES = new Set<TasksViewMode>(["list", "grouped-by-list"]);
const viewMode: TasksViewMode = VALID_VIEW_MODES.has(rawViewMode)
  ? rawViewMode
  : DEFAULT_VIEW_MODE;
```

---

src/components/providers/__tests__/TasksProvider.integration.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+import type { TaskListConfig } from "@/components/tasks/types";
+import { act, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockFetch = vi.fn();
+global.fetch = mockFetch as unknown as typeof fetch;
+
+const configA: TaskListConfig = {
+  id: "config-a",
```
BenSeymourODB, line 34: **Fetch mock leaks across test files — use `vi.stubGlobal` with teardown.**

```ts
global.fetch = mockFetch as unknown as typeof fetch;
```

This mutates the global permanently for the lifetime of the test process. The `CalendarProvider` test suite uses the correct pattern:
```ts
vi.stubGlobal("fetch", fetchMock);  // in beforeEach
vi.unstubAllGlobals();               // in afterEach
```

Without the teardown, any test file that runs after this one and doesn't set up its own fetch mock will silently use this file's `mockFetch` (which has been `.mockReset()`'d, so it returns `undefined`), causing confusing failures in unrelated tests. Please replace with `vi.stubGlobal` / `vi.unstubAllGlobals()`.

---

src/components/providers/__tests__/TasksProvider.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+describe("TasksProvider — activeConfig resolution", () => {
+  it("returns null activeConfig and empty lists when configs is empty", () => {
+    const { result } = renderHook(() => useTasksContext(), {
+      wrapper: wrap([]),
+    });
+
+    expect(result.current.activeConfig).toBeNull();
+    expect(result.current.lists).toEqual([]);
+    expect(result.current.configs).toEqual([]);
+  });
```
BenSeymourODB, line 100: **Missing test: `initialActiveConfigId` override wins over a pre-existing localStorage value.**

There are tests for rehydration from localStorage and for the stale-id fallback, but no test covers the case where `initialActiveConfigId` is passed and a different value is already persisted. Given the bug noted in `TasksProvider.tsx` (the override is currently silently ignored in that case), adding this test now will serve as both a regression net and a red test to drive the fix:

```ts
it("initialActiveConfigId wins over a pre-existing persisted id", () => {
  window.localStorage.setItem(
    TASKS_ACTIVE_CONFIG_LS_KEY,
    JSON.stringify("config-a")
  );
  const configs = [makeConfig({ id: "config-a" }), makeConfig({ id: "config-b" })];
  const { result } = renderHook(() => useTasksContext(), {
    wrapper: ({ children }) => (
      <TasksProvider configs={configs} initialActiveConfigId="config-b">
        {children}
      </TasksProvider>
    ),
  });
  expect(result.current.activeConfig?.id).toBe("config-b");
});
```

---

BenSeymourODB (commented):
First-pass automated review for PR #349 (`TasksProvider`, issue #331).

**Summary of findings:**

The implementation is clean, CLAUDE.md-compliant (no manual memoization, no `any`, strict types), and the spec's core contract (ephemeral state only, DB-backed settings stay in `useUserSettings`, `useTasks` pass-through, stale-id fallback) is correctly implemented. Test coverage is thorough and TDD structure is sound.

**Issues to address before merge:**

1. **`initialActiveConfigId` is silently ignored when localStorage already has a value** (see inline comment on line 109). The override must be written through via a mount-only `useEffect`, matching the `CalendarProvider` / `initialView` pattern. This is a correctness bug for the deep-link use case documented in the JSDoc.

2. **Stale `storedConfigId` after fallback fires is never reconciled** (line 127). The stale key persists in localStorage so every reload re-triggers the fallback. Decide intentionally and comment either way.

3. **No `viewMode` validation against unknown stored values** (line 114). A corrupted or future-removed value passes through typed as `TasksViewMode`. A one-liner guard matching `CalendarProvider`'s migration pattern would harden this.

4. **Integration test leaks `global.fetch` mutation** (integration test line 32). Replace with `vi.stubGlobal` / `vi.unstubAllGlobals()` to match the CalendarProvider test pattern and avoid cross-file mock contamination.

5. **Missing test for `initialActiveConfigId` override semantics** (unit test line ~100). The current suite doesn't cover the case where the prop must win over a pre-existing stored value — add a red test to drive the fix for issue #1 above.

The redundant `setActiveConfigId`/`setViewMode` wrapper functions (inline comments on line 117) are a minor code quality point, not a blocker.

---

src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+export function TasksProvider({
+  children,
+  configs = [],
+  initialActiveConfigId,
+  initialViewMode,
+}: TasksProviderProps) {
+  const [storedConfigId, setStoredConfigId] = useLocalStorage<string | null>(
+    TASKS_ACTIVE_CONFIG_LS_KEY,
+    initialActiveConfigId ?? null
+  );
```
BenSeymourODB, line 121: **Bug: `initialActiveConfigId` is silently ignored when localStorage already holds a value.**

`useLocalStorage` only uses `initialValue` as a fallback when the key is absent from storage. If a user already has `tasks.activeConfigId` persisted (e.g. from a previous session), passing `initialActiveConfigId` via a URL deep-link will have no effect — the persisted id wins unconditionally.

The jsdoc on the prop says it should "win over the persisted value on first mount", which matches how `CalendarProvider` handles `initialView` via a mount-only `useEffect` that writes through to storage when the override prop is defined.

Fix: add a mount-only effect that calls `setStoredConfigId(initialActiveConfigId)` when `initialActiveConfigId` is defined — exactly the pattern in `CalendarProvider` lines 253–274. The same bug affects `initialViewMode` on lines 114–117.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: **Same bug as above for `initialViewMode`.**

`initialViewMode ?? DEFAULT_VIEW_MODE` is passed as the `initialValue`, so it is only used when `tasks.viewMode` is absent from localStorage. A returning user with a persisted `"grouped-by-list"` can never be deep-linked back to `"list"` via this prop.

Same fix: mount-only effect that writes `initialViewMode` to storage when the prop is defined.

---

src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+    TASKS_VIEW_MODE_LS_KEY,
+    initialViewMode ?? DEFAULT_VIEW_MODE
+  );
+
+  // Resolve `activeConfig` with a stale-id fallback: if the persisted id
+  // points at a config the user has since deleted (or to nothing on the
+  // very first mount), fall through to the first available config rather
+  // than rendering an empty surface that "should" have content.
+  const activeConfig =
+    configs.find((c) => c.id === storedConfigId) ?? configs[0] ?? null;
```
BenSeymourODB: **Stale-id fallback doesn't update storage.**

When `configs.find(...)` misses (stale or absent id) and we fall through to `configs[0]`, `storedConfigId` still holds the stale value in localStorage. On the next mount the same fallback fires again — no harm functionally, but the storage stays forever dirty. Consider calling `setStoredConfigId(configs[0].id)` in a `useEffect` when `activeConfig !== null && storedConfigId !== activeConfig.id` so the stored value self-heals after a config is deleted.

---

src/components/providers/TasksProvider.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const setViewMode = (mode: TasksViewMode) => {
+    setStoredViewMode(mode);
+  };
+
+  const lists = activeConfig ? activeConfig.lists.filter((l) => l.enabled) : [];
+
+  const tasksApi = useTasks(activeConfig);
+
+  const value: ITasksContext = {
+    ...tasksApi,
```
BenSeymourODB: **`...tasksApi` spread can silently shadow future context fields.**

`UseTasksReturn` currently has `{ tasks, loading, error, refreshTasks, updateTask }` — no collision today. But if `useTasks` ever grows a field whose name matches a provider-owned key (e.g. both expose `loading`), the spread wins silently and the provider's own value disappears from context. Consider destructuring the exact keys you need instead of spreading:

```ts
const { tasks, loading, error, refreshTasks, updateTask } = tasksApi;
const value: ITasksContext = {
  tasks, loading, error, refreshTasks, updateTask,
  configs, activeConfig, setActiveConfigId, lists, viewMode, setViewMode,
};
```

This makes the boundary explicit and TypeScript will flag a new conflict at the assignment site.

---

src/components/providers/__tests__/TasksProvider.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+
+    const configs = [
+      makeConfig({ id: "config-a" }),
+      makeConfig({ id: "config-b" }),
+    ];
+    const { result } = renderHook(() => useTasksContext(), {
+      wrapper: wrap(configs),
+    });
+
+    expect(result.current.activeConfig?.id).toBe("config-b");
```
BenSeymourODB, line 128: **Missing test: `initialActiveConfigId` should win over a persisted value.**

There is a test for "rehydrates from localStorage" but no test asserting that passing `initialActiveConfigId` overrides a conflicting persisted id. This is the exact gap that hides the bug on lines 110–113 of the provider — the prop's override semantics go completely untested. Add:

```ts
it("initialActiveConfigId wins over a previously-persisted id", () => {
  window.localStorage.setItem(TASKS_ACTIVE_CONFIG_LS_KEY, JSON.stringify("config-a"));
  const configs = [makeConfig({ id: "config-a" }), makeConfig({ id: "config-b" })];
  const { result } = renderHook(() => useTasksContext(), {
    wrapper: ({ children }) => (
      <TasksProvider configs={configs} initialActiveConfigId="config-b">
        {children}
      </TasksProvider>
    ),
  });
  expect(result.current.activeConfig?.id).toBe("config-b");
});
```

Same gap exists for `initialViewMode`.

---

src/components/providers/__tests__/TasksProvider.integration.test.tsx
```diff
@@ -0,0 +1,173 @@
...
+  TasksProvider,
+  useTasksContext,
+} from "@/components/providers/TasksProvider";
+import type { TaskListConfig } from "@/components/tasks/types";
+import { act, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const mockFetch = vi.fn();
+global.fetch = mockFetch as unknown as typeof fetch;
```
BenSeymourODB: **`global.fetch` reassignment bleeds across test files.**

Assigning `global.fetch = mockFetch` at module scope means the real `fetch` is never restored — it stays patched for any test file that runs after this one in the same worker. Use Vitest's `vi.stubGlobal` instead, which is automatically undone after the suite:

```ts
beforeEach(() => {
  vi.stubGlobal("fetch", vi.fn());
  mockFetch = vi.mocked(globalThis.fetch);
  window.localStorage.clear();
});

afterEach(() => {
  vi.unstubAllGlobals();
});
```

Or at minimum wrap the current assignment in a `beforeAll` / `afterAll` pair that restores `originalFetch`.

---

BenSeymourODB (commented):
First-pass review of `TasksProvider` (#331). Overall the structure is clean — correct context null-init, no manual memoization, good separation of ephemeral vs DB-backed state, and the test coverage is solid for the happy paths. Two real bugs and three quality issues are called out inline.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Fixed in 1f5fb1e. Added a mount-only `useEffect` that writes both `initialActiveConfigId` and `initialViewMode` through to storage when the props are defined, mirroring `CalendarProvider`'s pattern. New unit test in `TasksProvider.test.tsx` (`initialActiveConfigId wins over a previously-persisted id`) locks in the override semantics.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Fixed in 1f5fb1e. The wrapper functions are gone — `setActiveConfigId: setStoredConfigId` and `setViewMode: setStoredViewMode` are exposed directly. The public `ITasksContext` types still constrain callers to the intended argument shape.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Fixed in 1f5fb1e. Added a reconciliation `useEffect` that writes `activeConfig.id` back to storage whenever `storedConfigId` is non-null but doesn't resolve to a config — so the stale key self-heals after one render and reload starts on the correct config without re-firing the fallback. The existing fallback test now also asserts the post-reconciliation localStorage value.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Fixed in 1f5fb1e. Added an `isValidViewMode` guard backed by a `VALID_VIEW_MODES` set; an unrecognised stored value falls through to `DEFAULT_VIEW_MODE`. New test (`falls back to default viewMode when the stored value is not in the union`) covers the corrupted-payload path.

---

src/components/providers/__tests__/TasksProvider.integration.test.tsx
BenSeymourODB: Fixed in 1f5fb1e. Replaced the module-scope `global.fetch =` assignment with `vi.stubGlobal("fetch", mockFetch)` in `beforeEach` and `vi.unstubAllGlobals()` in `afterEach`, matching the `CalendarProvider` test pattern.

---

src/components/providers/__tests__/TasksProvider.test.tsx
BenSeymourODB: Added in 1f5fb1e — `initialActiveConfigId wins over a previously-persisted id` and a sibling `initialViewMode wins over a previously-persisted value` both assert the override resolves correctly *and* gets written through to storage so reload preserves it.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Duplicate of the line-117 thread. Both `initialActiveConfigId` and `initialViewMode` are now written through on mount via a single `useEffect` (1f5fb1e), and the new override-wins test in `TasksProvider.test.tsx` locks the behaviour in.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Fixed in 1f5fb1e alongside the `initialActiveConfigId` fix — same mount-only `useEffect` calls `setStoredViewMode(initialViewMode)` when defined. New `initialViewMode wins over a previously-persisted value` test asserts a returning user with `"grouped-by-list"` persisted gets deep-linked back to `"list"` correctly.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Duplicate of the earlier stale-id thread. Reconciliation `useEffect` added in 1f5fb1e — writes `activeConfig.id` back to storage so the stored value self-heals after a config is deleted, exactly as suggested.

---

src/components/providers/TasksProvider.tsx
BenSeymourODB: Fixed in 1f5fb1e. The `value` object now destructures `{ tasks, loading, error, refreshTasks, updateTask }` from `useTasks(activeConfig)` explicitly, so any future name collision with provider-owned keys becomes a TS error at the assignment site rather than a silent runtime shadow.

---

src/components/providers/__tests__/TasksProvider.test.tsx
BenSeymourODB: Duplicate of the line-100 thread. Both override-wins tests (`initialActiveConfigId wins over a previously-persisted id` and `initialViewMode wins over a previously-persisted value`) added in 1f5fb1e — they assert resolution *and* that the override is written through to storage.

---

src/components/providers/__tests__/TasksProvider.integration.test.tsx
BenSeymourODB: Duplicate of the line-34 thread. `vi.stubGlobal("fetch", mockFetch)` / `vi.unstubAllGlobals()` are now used in `beforeEach` / `afterEach` (1f5fb1e), so the patched `fetch` is automatically restored after each test instead of leaking across files in the same Vitest worker.

---

PR #350, opened 2026-05-07
feat(calendar): Clock view — floating off-arc title labels with connector lines (#311)

BenSeymourODB:
Closes #311.

## Summary

When an event title in the Clock view overflows even the 2-line in-arc budget from #310, render the full title as a floating label outside the clock face with a thin connector line back to the arc midpoint. The arc keeps its colored fill and leading event emoji; the in-arc title is suppressed so the floating label is the sole title surface for that event.

## Approach (TDD, three phases)

1. **`clamp-label.ts`** — pure helper for the vertical clamp band `[clockTop − 0.10×h, clockBottom + 0.10×h]` so the AnalogClockView grid row height stays stable regardless of how many overflowing labels exist. **11 unit tests.**
2. **`floating-label.tsx`** — pure SVG primitive: connector `<line>` + bordered `<rect>` + `<text>`. Uses `clampLabelPosition` and a small `rectEdgeIntersection` helper to anchor the connector on the rect side facing the anchor. Optional `onClick` mirrors `EventArc` so the wiring layer can route into `EventDetailModal`. **17 component tests.**
3. **Wiring** — extracted `arc-title-layout.ts` so `EventArc` and `AnalogClock` agree on `didOverflow` without independent recomputation. `EventArc` gains a `forceHideTitle` prop that suppresses the in-arc title (preserving the leading emoji). `AnalogClock` renders a sibling `floating-labels-layer` for overflowing events; SVG carries `overflow="visible"` (presentation attribute, not inline `style`) so labels can paint outside the clock viewBox. **+8 integration tests + 4 EventArc tests.**

## Spec compliance checklist

- [x] Trigger: `fitTitleToArc(...).didOverflow === true` (and arc is wide enough to render at all)
- [x] Arc in overflow mode: keeps fill, keeps leading event emoji, drops in-arc title
- [x] Label sits on an invisible outer circle at `outerRadius + arcThickness × 0.6`
- [x] Connector at arc midpoint outer-radius, `strokeOpacity=0.6`
- [x] Vertical clamp keeps grid row height stable
- [x] Multiple labels: angular order, no collision avoidance for v1
- [x] Floating label is clickable when `onEventClick` is provided

## Test plan

- [x] `pnpm test` — 2059/2059 vitest tests pass (40 new across this PR)
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (only pre-existing warnings)
- [x] Playwright spec added at `e2e/analog-clock-floating-labels.spec.ts` exercising the production-style fixture page; **could not run locally** (browser binary download blocked in this session) — relying on CI

## Files

- `src/components/calendar/analog-clock/clamp-label.ts` (new)
- `src/components/calendar/analog-clock/floating-label.tsx` (new)
- `src/components/calendar/analog-clock/arc-title-layout.ts` (new — shared layout helper)
- `src/components/calendar/analog-clock/analog-clock.tsx` (wiring)
- `src/components/calendar/analog-clock/event-arc.tsx` (uses helper, accepts `forceHideTitle`)
- `src/components/calendar/analog-clock/types.ts` (`EventArcProps.forceHideTitle`)
- Tests: `__tests__/clamp-label.test.ts`, `__tests__/floating-label.test.tsx`, `__tests__/event-arc.test.tsx` (+4), `__tests__/analog-clock.test.tsx` (+8)
- `e2e/analog-clock-floating-labels.spec.ts` (new)

https://claude.ai/code/session_019iskjPEeRXWu1K8B22o9qN

### DIFF
--- a/e2e/analog-clock-floating-labels.spec.ts
+++ b/e2e/analog-clock-floating-labels.spec.ts
@@ -1,0 +1,84 @@
+import { expect, test } from "@playwright/test";
+
+/**
+ * Floating off-arc title labels (#311).
+ *
+ * Exercises the overflow path on top of #310's 2-line wrap: when a title
+ * cannot fit even on two curved lines (or sits on a sub-30° arc whose
+ * single-line budget is too tight), the in-arc title is suppressed and a
+ * sibling FloatingLabel renders the full text outside the clock face with
+ * a thin connector back to the arc midpoint.
+ *
+ * The default scenario is sufficient: "Team Standup" (12 chars) sits on a
+ * 15° arc whose single-line budget is well below 12 chars, so the title
+ * overflows and a floating label renders.
+ */
+test.describe("Analog Clock — floating off-arc labels (#311)", () => {
+  test("renders a floating label for an event whose title overflows the arc", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    await expect(page.getByTestId("analog-clock")).toBeVisible();
+
+    const layer = page.getByTestId("floating-labels-layer");
+    await expect(layer).toBeAttached();
+
+    // d3 ("Team Standup") sits on a 15° arc; its 12-char title overflows.
+    const label = page.getByTestId("floating-label-d3");
+    await expect(label).toBeVisible();
+
+    const text = page.getByTestId("floating-label-text-d3");
+    await expect(text).toHaveText("Team Standup");
+  });
+
+  test("suppresses the in-arc title for an overflowing event but keeps the colored arc", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    await expect(page.getByTestId("analog-clock")).toBeVisible();
+
+    // The arc fill is still painted.
+    await expect(page.getByTestId("event-arc-d3")).toBeVisible();
+    // …but no in-arc title text is rendered (would otherwise be event-title-d3).
+    await expect(page.getByTestId("event-title-d3")).toHaveCount(0);
+  });
+
+  test("renders a connector line in the event's color back to the arc midpoint", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    const connector = page.getByTestId("floating-label-connector-d3");
+    await expect(connector).toBeAttached();
+    // d3 is the blue Team Standup event.
+    await expect(connector).toHaveAttribute("stroke", "#3B82F6");
+  });
+
+  test("the analog clock SVG carries overflow='visible' so labels can paint outside the viewBox", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    const svg = page.getByTestId("analog-clock");
+    await expect(svg).toHaveAttribute("overflow", "visible");
+  });
+
+  test("does not render a floating label for an event whose title fits inside the arc", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    // d1 ("Family Game Night") sits on a 45° arc — fits across two lines.
+    await expect(page.getByTestId("event-title-d1")).toBeVisible();
+    await expect(page.getByTestId("floating-label-d1")).toHaveCount(0);
+  });
+
+  test("captures a screenshot of the default scenario showing the floating label layer", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    await expect(page.getByTestId("analog-clock")).toBeVisible();
+    await expect(page.getByTestId("floating-label-d3")).toBeVisible();
+    await page.screenshot({
+      path: "test-results/screenshots/clock-floating-labels-default.png",
+      fullPage: true,
+    });
+  });
+});
--- a/src/components/calendar/analog-clock/analog-clock.tsx
+++ b/src/components/calendar/analog-clock/analog-clock.tsx
@@ -10,6 +10,7 @@
  * Overlapping events are stacked at different radii (inner rings).
  */
 import { useEffect, useState } from "react";
+import { computeArcTitleLayout } from "./arc-title-layout";
 import { ClockFace } from "./clock-face";
 import {
   eventsToClockEvents,
@@ -17,6 +18,7 @@ import {
   getPeriodBounds,
 } from "./clock-utils";
 import { EventArc } from "./event-arc";
+import { FloatingLabel } from "./floating-label";
 import type { AnalogClockProps, ClockEvent } from "./types";
 
 /** Default clock diameter in pixels */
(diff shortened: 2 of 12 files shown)


src/components/calendar/analog-clock/analog-clock.tsx
```diff
@@ -129,28 +171,55 @@ export function AnalogClock({
...
+          )
+        )}
+      </g>
+
+      {/* Floating off-arc title labels for overflowing events (#311). Painted
+          above the arcs but below the clock face so the hands draw over any
+          label that bleeds toward the centre. */}
+      <g data-testid="floating-labels-layer">
+        {eventLayouts
+          .filter(({ layout }) => layout.fit.didOverflow)
```
BenSeymourODB: **[MEDIUM] Missing `arcSpan >= 10` guard — spec deviation causes spurious floating labels on sub-10° arcs**

The issue spec states the floating-label trigger is:
> `fitTitleToArc(...)` returns `didOverflow: true` **AND** the arc is wide enough to display an emoji (i.e. `arcSpan >= 10`)

The current filter only checks `didOverflow`:
```ts
.filter(({ layout }) => layout.fit.didOverflow)
```

For a sub-10° arc, `fitTitleToArc` will always return `didOverflow: true` (the char budget collapses to 0–2, any non-empty title overflows), so a `FloatingLabel` renders for an arc that has neither an emoji nor an in-arc title — a disconnected callout pointing at a nearly invisible sliver.

Fix: include the emoji threshold in the filter. Since `arcSpan` is not currently stored in `eventLayouts`, the cleanest fix is to add it there (computed on line 130 already) and then guard:
```ts
.filter(({ layout, arcSpan }) => layout.fit.didOverflow && arcSpan >= 10)
```

The corresponding `forceHideTitle` passed to `EventArc` on line 191 is harmless for these arcs (`showTitle` already short-circuits on `arcSpan < 20`) but should be guarded the same way for clarity:
```ts
forceHideTitle={layout.fit.didOverflow && arcSpan >= 10}
```

The E2E scenario (`arcSpan = 15°`) is unaffected since 15 ≥ 10.

---

src/components/calendar/analog-clock/event-arc.tsx
```diff
@@ -102,20 +99,16 @@ export function EventArc({
...
-  // on the arc's available circumference at titleRadius. Below
-  // TWO_LINE_MIN_SPAN_DEGREES we keep the single-line truncation behavior so
-  // narrow arcs don't end up with 2-char-per-line stubs.
-  const maxLines: 1 | 2 = arcSpan >= TWO_LINE_MIN_SPAN_DEGREES ? 2 : 1;
-  const fit = fitTitleToArc(
+
+  // Title layout (radius, font size, 1- vs 2-line, fit) is shared with
+  // AnalogClock so the floating-label overflow path (#311) agrees on
+  // didOverflow without recomputing.
+  const { titleRadius, titleFontSize, fit } = computeArcTitleLayout({
```
BenSeymourODB: **[LOW] `computeArcTitleLayout` is called twice with identical inputs — layout computed in `AnalogClock` is not passed down**

`AnalogClock` already calls `computeArcTitleLayout` for every event (to determine `didOverflow`) and stores the result in `eventLayouts`. `EventArc` then re-calls the same function with exactly the same `innerRadius`, `outerRadius`, `arcSpan`, and `cleanTitle`.

The results are always identical today, so this is not a correctness bug. However it undermines the module's own stated goal (the JSDoc says "keeps EventArc and AnalogClock agreeing on `didOverflow` without recomputing"). If someone later changes `EventArc` to pass slightly different radii, the two results can silently diverge, causing a title to be both suppressed by `forceHideTitle` (based on the `AnalogClock` computation) and simultaneously rendered (because the `EventArc`-local `fit.lines.length > 0`).

The straightforward fix is to add an optional `precomputedLayout?: ArcTitleLayout` prop to `EventArcProps` and use it when supplied, falling back to a local call only when absent (preserving the existing standalone-use contract). Alternatively, pass `layout` from `eventLayouts` directly into `EventArc` as a required prop and remove the internal call entirely.

---

src/components/calendar/analog-clock/arc-title-layout.ts
```diff
@@ -0,0 +1,58 @@
...
+  );
+  const maxLines: 1 | 2 = arcSpan >= TWO_LINE_MIN_SPAN_DEGREES ? 2 : 1;
+  const fit = fitTitleToArc(
+    cleanTitle,
+    arcSpan,
+    titleRadius,
+    titleFontSize,
+    maxLines
+  );
+  return { titleRadius, titleFontSize, maxLines, fit };
```
BenSeymourODB, line 57: **[LOW] `computeArcTitleLayout` has no dedicated unit test**

`arc-title-layout.ts` is the shared contract between `EventArc` and `AnalogClock` — the JSDoc explicitly flags what happens when they disagree. `clamp-label.ts` (a simpler helper) gets 13 focused unit tests; `arc-title-layout.ts` (the more critical invariant) is only exercised indirectly through the component tests.

Suggest adding a `__tests__/arc-title-layout.test.ts` covering at minimum:
- `arcSpan >= TWO_LINE_MIN_SPAN_DEGREES` → `maxLines === 2`
- `arcSpan < TWO_LINE_MIN_SPAN_DEGREES` → `maxLines === 1`
- Returned `titleRadius` equals `innerRadius + arcHeight * 0.68`
- `titleFontSize` is capped at `TITLE_FONT_SIZE_MAX`
- That `fit.didOverflow` is `true` for a title that clearly exceeds the arc budget

This is a test-coverage gap, not a runtime defect, but it's worth addressing given the spec's emphasis on the shared-source-of-truth invariant.

---

e2e/analog-clock-floating-labels.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    await expect(page.getByTestId("floating-label-d1")).toHaveCount(0);
+  });
+
+  test("captures a screenshot of the default scenario showing the floating label layer", async ({
+    page,
+  }) => {
+    await page.goto("/test/analog-clock?scenario=default&hour=10&min=10");
+    await expect(page.getByTestId("analog-clock")).toBeVisible();
+    await expect(page.getByTestId("floating-label-d3")).toBeVisible();
+    await page.screenshot({
```
BenSeymourODB, line 79: **[INFO] Screenshot test writes to `test-results/` — that's the right path, but worth noting**

The screenshot is saved to `test-results/screenshots/clock-floating-labels-default.png`, which is covered by the `.gitignore` (`/test-results`), so the artifact will not be committed. No action needed, just confirming the path is correct and the gitignore entry protects it as per CLAUDE.md's "never commit test output artifacts" rule.

---

BenSeymourODB (commented):
Overall this is a clean, well-structured implementation. The `arc-title-layout.ts` extraction to share the layout contract, the `clampLabelPosition` pure helper with excellent test coverage, the `rectEdgeIntersection` geometry for connector endpoints, and the `overflow="visible"` SVG attribute are all correct and spec-compliant. TDD coverage is present and strong for `FloatingLabel` and `clamp-label`. No `any` types, no manual memoization, no non-standard Tailwind colors.

Three issues found — one requires a fix before merge:

**Must fix (MEDIUM):** The floating-label filter on line 202 of `analog-clock.tsx` is missing the `arcSpan >= 10` guard required by the spec. Sub-10° arcs (too narrow to show an emoji) can still produce a `FloatingLabel`, rendering a dangling callout pointing at an invisible arc sliver. See inline comment on line 202.

**Should fix (LOW):** `EventArc` recomputes `computeArcTitleLayout` with the same inputs that `AnalogClock` already computed. The shared-source-of-truth goal of `arc-title-layout.ts` is partially undermined — a future divergence in inputs would cause `forceHideTitle` and `fit.lines` to silently disagree. See inline comment on `event-arc.tsx` line 106.

**Nice to have (LOW):** `arc-title-layout.ts` has no dedicated unit test despite being the critical shared invariant the feature depends on. `clamp-label.ts` — a simpler helper — gets 13 focused tests. See inline comment on `arc-title-layout.ts` line 57.

---

src/components/calendar/analog-clock/analog-clock.tsx
BenSeymourODB: Fixed in 632f1a7. Added `arcSpan` to `eventLayouts` and a derived `isOverflow = layout.fit.didOverflow && arcSpan >= EMOJI_MIN_SPAN_DEGREES` (=10) flag — used for both the floating-label filter and the `forceHideTitle` prop on `EventArc` for symmetry. Pulled the threshold into a named constant at the top of the file with a JSDoc cross-reference to `EventArc`'s `showEmoji` gate. New integration test in `analog-clock.test.tsx` asserts no `FloatingLabel` renders for a 5° sliver even when `didOverflow=true`.

---

src/components/calendar/analog-clock/event-arc.tsx
BenSeymourODB: Fixed in 632f1a7. Took the optional-prop approach: added `precomputedLayout?: ArcTitleLayout` to `EventArcProps`. `AnalogClock` now passes its `eventLayouts[i].layout` down so `EventArc` consumes the exact same object that drove `forceHideTitle` — no recomputation, no possibility of silent divergence. The helper still falls back to a local call when omitted so standalone `EventArc` usage (existing tests, ad-hoc renders) keeps working.

---

src/components/calendar/analog-clock/arc-title-layout.ts
BenSeymourODB: Fixed in 632f1a7. Added `__tests__/arc-title-layout.test.ts` (13 tests) covering each suggestion: `maxLines` branching at the threshold (incl. boundary), `titleRadius = innerRadius + arcHeight × TITLE_RADIUS_RATIO`, `titleFontSize` capped at `TITLE_FONT_SIZE_MAX`, `didOverflow` reported true/false on representative inputs, plus an explicit "determinism invariant" test that asserts two calls with identical inputs produce strictly identical output (the contract `EventArc` and `AnalogClock` depend on).

---

e2e/analog-clock-floating-labels.spec.ts
BenSeymourODB: Confirmed — `test-results/` is gitignored and the screenshot path is intentional (matches the convention used by the existing `e2e/analog-clock.spec.ts` screenshots, e.g. `test-results/screenshots/clock-default-am.png`). No action needed.

---

PR #353, opened 2026-05-07
chore(react-compiler): ban manual memoization + strip violations (#271)

BenSeymourODB:
## Summary

Closes #271.

CLAUDE.md forbids manual memoization (`useMemo`, `useCallback`, `React.memo`) under the React Compiler, but three prior PR reviews (#196, #197, #200) flagged repeat violations and asked for an audit + lint-rule follow-up. This PR delivers both, plus the cleanup.

### Phase 1 — ESLint enforcement (commit 1)

- **`no-restricted-imports`** flags any `useMemo`, `useCallback`, or `memo` named-imported from `react`.
- **`no-restricted-syntax`** flags `React.memo` / `React.useMemo` / `React.useCallback` accessed via default or namespace imports.
- Vendored shadcn under `src/components/ui/**` is exempt — `pnpm bump-ui` overwrites it from upstream and we don't want to fork those files.
- Integration tests (`__tests__/eslint-no-manual-memo.test.ts`) lint sample sources against the real flat config so regressions in the rule itself fail tests, not just `pnpm lint`.
- CLAUDE.md and `docs/react-compiler.md` updated to reference the rule and explain `"use no memo"` as the only sanctioned escape hatch.

### Phase 2 + 3 — Cleanup + RC in vitest (commit 2)

Stripped the 9 violations from 8 files:

- `src/components/providers/CalendarProvider.tsx` (5 sites)
- `src/components/tasks/use-tasks.ts` (2)
- `src/components/profiles/profile-context.tsx` (3)
- `src/components/recipe/recipe-display.tsx` (1)
- `src/components/recipe/use-zoom.ts` (3)
- `src/components/recipe/use-recipe-pagination.ts` (5)
- `src/components/rewards/points-context.tsx` (2)
- `src/hooks/useLocalStorage.ts` (3)

`MockCalendarProvider.tsx` (referenced in the issue body as having a `useMemo` on line 163) was already cleaned up — line 163 is now a comment about RC memoization.

#### Subtle find: vitest needed React Compiler too

Pure removal of the `useCallback` wrappers caused `Maximum update depth exceeded` infinite loops in `profile-context.test.tsx` (8/15 failing). Root cause: vitest doesn't run through Next's build pipeline, so RC's auto-memoization wasn't applied — every render produced a new inline-function identity, every `useEffect` whose deps referenced that function re-fired, every effect setState, and the cycle repeated. Production was fine because RC was active; tests were broken because RC was absent.

Fix: `babel-plugin-react-compiler` added to `vitest.config.ts`'s `plugin-react` config (already a project dependency). Tests now see the same memoization behaviour as production. Per CLAUDE.md TDD policy, no test was deleted or weakened.

Where `react-hooks/exhaustive-deps` would have demanded a banned `useCallback` re-wrap (4 sites), the rule is suppressed inline at the function declaration with a comment pointing back to `docs/react-compiler.md`.

## Test plan

- [x] `pnpm lint` — 0 errors. The 6 remaining warnings are pre-existing (unused vars, `` in `profile-avatar`/`account-section`) and not in scope.
- [x] `pnpm test` — **132 files / 2027 tests passing**, including the 8 new ESLint integration tests
- [x] No manual memoization remains in `src/` outside `src/components/ui/` (vendored shadcn)
- [x] Lint rule fires on each banned form (named import, namespace import, default import) — verified by integration tests

https://claude.ai/code/session_0153T9AjTeVeLbR4hu5dniQP

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -70,6 +70,8 @@ Application Insights is **not yet implemented**. The logger abstraction exists f
 
 Enabled by default. **DO NOT** use manual memoization (`useMemo`, `useCallback`, `React.memo`). Write clean, simple React code. See [docs/react-compiler.md](./docs/react-compiler.md).
 
+This is enforced by ESLint: `no-restricted-imports` bans `useMemo`/`useCallback`/`memo` named imports from `react`, and `no-restricted-syntax` bans `React.memo`/`React.useMemo`/`React.useCallback` member access. Vendored `src/components/ui/**` is exempt because `pnpm bump-ui` overwrites it from upstream shadcn.
+
 ### shadcn/ui
 
 Components are **copied into `src/components/ui/`**, not installed as a package. Warn the user if installing components would overwrite existing customizations. Use the Shadcn MCP server to browse and install.
--- a/__tests__/eslint-no-manual-memo.test.ts
+++ b/__tests__/eslint-no-manual-memo.test.ts
@@ -1,0 +1,142 @@
+import { ESLint } from "eslint";
+import path from "node:path";
+import { describe, expect, it } from "vitest";
+
+/**
+ * Integration tests for the project's ESLint config (#271).
+ *
+ * Manual memoization (`useMemo`, `useCallback`, `React.memo`) is forbidden
+ * by CLAUDE.md because the project uses the React Compiler. These tests
+ * lint sample sources against the real `eslint.config.mjs` and assert
+ * that the corresponding rules fire — and that legitimate React code is
+ * still accepted.
+ *
+ * Vendored shadcn components under `src/components/ui/**` are globally
+ * ignored and are NOT exercised here.
+ */
+
+const cwd = path.resolve(__dirname, "..");
+
+function eslint() {
+  // useEslintrc / overrideConfigFile defaults pick up the project's
+  // flat config (`eslint.config.mjs`) automatically.
+  return new ESLint({ cwd, errorOnUnmatchedPattern: false });
+}
+
+const fixtureFile = "src/__lint_fixture__.tsx";
+
+async function lint(source: string) {
+  const results = await eslint().lintText(source, { filePath: fixtureFile });
+  const messages = results.flatMap((r) => r.messages);
+  return messages;
+}
+
+describe("eslint config: manual memoization ban (#271)", () => {
+  it("flags `useMemo` named imported from react", async () => {
+    const messages = await lint(
+      `import { useMemo } from "react";\nexport const value = useMemo(() => 1, []);\n`
+    );
+    expect(
+      messages.some(
+        (m) =>
+          m.ruleId === "no-restricted-imports" &&
+          typeof m.message === "string" &&
+          /useMemo|memoization/i.test(m.message)
+      )
+    ).toBe(true);
+  });
+
+  it("flags `useCallback` named imported from react", async () => {
+    const messages = await lint(
+      `import { useCallback } from "react";\nexport const cb = useCallback(() => {}, []);\n`
+    );
+    expect(
+      messages.some(
+        (m) =>
+          m.ruleId === "no-restricted-imports" &&
+          typeof m.message === "string" &&
+          /useCallback|memoization/i.test(m.message)
+      )
+    ).toBe(true);
+  });
+
+  it("flags `memo` named imported from react", async () => {
+    const messages = await lint(
+      `import { memo } from "react";\nexport const Wrapped = memo(function X() { return null; });\n`
+    );
+    expect(
+      messages.some(
+        (m) =>
+          m.ruleId === "no-restricted-imports" &&
+          typeof m.message === "string" &&
+          /memo|memoization/i.test(m.message)
+      )
+    ).toBe(true);
+  });
+
+  it("flags `React.memo` member access via namespace import", async () => {
+    const messages = await lint(
+      `import * as React from "react";\nexport const Wrapped = React.memo(function X() { return null; });\n`
+    );
+    expect(
+      messages.some(
+        (m) =>
+          m.ruleId === "no-restricted-syntax" &&
+          typeof m.message === "string" &&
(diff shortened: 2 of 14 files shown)


BenSeymourODB (commented):
## Deep Review — chore(react-compiler): ban manual memoization + strip violations (#271)

Reviewed: diff, all 8 modified source files on disk, ESLint rule behavior (live linting via `node`), `vitest.config.ts` / `next.config.ts` parity, and all 4 `eslint-disable` comments. Here are my findings ranked by severity.

---

### Finding 1 — Medium: `no-restricted-syntax` only catches `React.*` by name; aliases slip through

**File:** `eslint.config.mjs` lines 41–54 / `docs/react-compiler.md` updated section

The selector `MemberExpression[object.name='React'][property.name=/^(memo|useMemo|useCallback)$/]` only fires when the namespace identifier is literally `React`. I confirmed live:

```ts
import * as Foo from "react";
export const W = Foo.memo(fn); // ← no lint error
```

```ts
import { default as R } from "react";
export const W = R.memo(fn); // ← no lint error
```

The docs say the rule covers "default or `* as React` imports", which is only true when the alias happens to be `React`. This is a gap in the rule, not in the current codebase (no such aliasing exists today), but it means the docs overstate the coverage and a future contributor could circumvent the ban unintentionally.

**Suggestion:** Either narrow the docs to say "catches `React.memo/useMemo/useCallback` when the import is named `React`", or add a test case (and note the gap). `no-restricted-imports` already correctly catches the named-import form regardless of aliasing, so the blast radius of this gap is limited to namespace/default import patterns.

---

### Finding 2 — Medium: `useSyncExternalStore` stability now depends entirely on RC, with no fallback documentation

**File:** `src/hooks/useLocalStorage.ts`

`useSyncExternalStore` has a contract: `subscribe` must be referentially stable (or the store will re-subscribe on every render), and `getSnapshot` must be pure and stable. Before this PR, `useCallback` enforced both; now RC is the only guarantee.

RC does memoize these — `subscribe` captures `key` (stable as long as the caller passes a stable string, which it does), and `getSnapshot` captures `key` too. In practice this is safe because all callers today pass string literals or state-derived strings that are themselves stable.

However, if a future caller passes a dynamically computed string (`useLocalStorage(\`pref-${userId}\`, default)`), RC *will* still re-create `subscribe` on every render when `userId` changes, which is correct semantically. The risk is zero for current callers, but worth a comment in the file noting the RC dependency and why `useSyncExternalStore` does not need `useCallback` guards under RC — the current comment only says "RC memoizes based on captured dependencies" without explaining the `useSyncExternalStore`-specific stability requirement. Not a bug, but a maintainability hazard if someone reads this cold.

---

### Finding 3 — Low (confirmed correct, noting for visibility): `refreshEvents` eslint-disable placement

**File:** `src/components/providers/CalendarProvider.tsx` lines 556 and 905

There are two warnings that `react-hooks/exhaustive-deps` would have raised without the suppression:

1. **Line 556** (before `const refreshEvents = async () => {`): The rule fires on the *variable declarator* when the function is referenced in a `useEffect` dep array that changes on every render. Verified live — the warning fires at the declaration site, so the `eslint-disable-next-line` on the line above it is correctly placed.

2. **Line 905** (inside the `[status]` useEffect, before `}, [status]`): The rule fires on the dep array, warning that `refreshEvents` is a missing dep. Also correctly suppressed.

Both suppression comments are justified: the ref-then-interval pattern (`refreshEventsRef.current = refreshEvents; useEffect(..., [refreshEvents])`) is a legitimate RC idiom and adding `refreshEvents` to the `[status]` effect's dep array would trigger a re-fetch on every render since RC was only activated in vitest post-cleanup.

The comments explaining why each suppress is valid are clear and point to the right docs. No action needed here.

---

### Finding 4 — Low: Test suite lacks an alias-import negative/positive case

**File:** `__tests__/eslint-no-manual-memo.test.ts`

The tests cover:
- named import (`import { memo }`) ✓
- namespace import as `React` (`import * as React`) ✓
- default import as `React` (`import React`) ✓
- non-react `memo` from `lodash-es` ✓
- legitimate hooks not flagged ✓

Missing:
- `import * as Foo from "react"; Foo.memo(...)` — should document that this is a known gap (not caught)
- `import { memo as M } from "react"; M(...)` — this *is* caught by `no-restricted-imports` (confirmed), so it would make a useful positive test to prove it

These are `nice-to-have` additions, not blockers.

---

### Finding 5 — Low: `vitest.config.ts` passes `{}` options to `babel-plugin-react-compiler`; `next.config.ts` uses `reactCompiler: true`

**File:** `vitest.config.ts` lines 11–14

`next.config.ts` uses `reactCompiler: true` (no options, inherits Next.js defaults). `vitest.config.ts` passes `["babel-plugin-react-compiler", {}]` (empty options object, plugin defaults). These defaults are the same upstream — `compilationMode: "infer"`, no `target` override — so there is no actual divergence today. This is the correct pattern for `@vitejs/plugin-react`. The PR description correctly explains the rationale (vitest doesn't go through Next's pipeline). No action needed, but if the project ever sets RC options in `next.config.ts` (e.g. `panicThreshold`), `vitest.config.ts` would need a matching update.

---

### Summary

- Cleanup is complete: zero actual `useMemo`/`useCallback`/`React.memo` calls remain in `src/` outside `src/components/ui/` (only comments reference the banned APIs).
- The ESLint rule is sound for all realistic usage patterns in this codebase. The alias-via-non-`React`-name gap is a theoretical bypass, not a current risk.
- All 4 `eslint-disable` comments are justified and correctly positioned.
- The vitest RC setup is the right fix for the infinite-loop regression; empty options match Next.js defaults.
- TDD policy satisfied: 8 new tests, no tests deleted or weakened.

The one thing worth fixing before merge (or in a follow-up) is narrowing the `docs/react-compiler.md` claim about what `no-restricted-syntax` covers — it currently overstates coverage for non-`React`-named namespace imports. Everything else is either correct or an enhancement.

---

PR #354, opened 2026-05-07
feat(settings): promote weekStartDay to UserSettings (DB-backed, app-wide) (#338)

BenSeymourODB:
## Summary

Closes #338.

`weekStartDay` was previously only a `CalendarProvider` field, persisted to per-browser localStorage and editable only via `CalendarSettingsPanel`. This PR promotes it to a DB-backed app-wide setting on `UserSettings`, so:

- it survives a browser change (already user-scoped),
- the main Settings page can edit it next to time/date/zoom format, and
- downstream features (Tasks due-date pickers, Meal Planner grid, Mini-Calendar widgets) get a single source of truth.

Per the issue body, this PR makes the source-of-truth change only — it does **not** rewrite the calendar helpers / a11y wiring. #205, #231, #352 stay as the consumer-side follow-ups.

## Phases

**Phase 1 — Schema/migration** (`0007_add_user_settings_week_start_day`)
Adds `weekStartDay Int @default(0)` to `UserSettings`. Greenfield-friendly ALTER (no `CONCURRENTLY` needed pre-launch).

**Phase 2 — API validation**
Extends the field-by-field validator on `PUT /api/settings` to enforce `weekStartDay ∈ {0, 1}`. Three rejection paths and two happy-path values are covered.

**Phase 3 — Hook + UI + provider migration shim**
- `useUserSettings` exposes `weekStartDay` (default `0`); the pick filter discards rogue values so a manually-edited DB row can't poison `weekStartsOn` parameters.
- `DisplaySection` gets a "Week starts on" radio group. `SettingsPage` defensively clamps non-`{0, 1}` ints to `0`.
- `CalendarSettingsPanel` keeps its existing radio; the provider's `setWeekStartDay` now PUTs through `/api/settings` when authenticated, while still mirroring to localStorage so unauthenticated surfaces work.
- One-shot localStorage→server migration: on first authenticated mount, if the user had a non-default local value and no `calendar-week-start-day-migrated` flag, PUT the local value up and tag the flag. Server value sync-down keeps the local state in step with cross-browser changes.

## Test plan

- [x] `pnpm test` — 2033 passing (was 2029, +4 migration-shim tests + 3 hook tests + 3 DisplaySection tests + 3 API tests)
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — 0 errors
- [x] `useUserSettings` returns `weekStartDay` from server, falls back to `0` on miss / unauth, rejects rogue values
- [x] `DisplaySection` shows the new control and fires `onChange` with `weekStartDay`
- [x] `CalendarProvider`: migration fires once with the right body, no-ops when local matches server, no-ops when the flag is set, and toggle PUTs through

## Out of scope

- Rewriting `WeekCalendar` / `MiniCalendarSidebar` / `applyCalendarKeyboardAction` to read from the new app-wide source — tracked under #205, #231, #352.

https://claude.ai/code/session_01CQv3BcWTsd1nVwWy5B6xRZ


src/components/providers/CalendarProvider.tsx
```diff
@@ -331,11 +340,83 @@ export function CalendarProvider({
...
+      }).catch((error) => {
+        logger.error(error as Error, {
+          context: "weekStartDayMigration",
+        });
+      });
+    }
+
+    window.localStorage.setItem(FLAG_KEY, "1");
+    weekStartMigrationRef.current = true;
+  }, [status, settings.weekStartDay, userSettings.weekStartDay]);
```
BenSeymourODB: **Race condition — migration fires before `useUserSettings` resolves**

The migration effect's dependency array includes `userSettings.weekStartDay`, but `useUserSettings` starts with `DEFAULT_USER_CALENDAR_SETTINGS.weekStartDay` (which is `0`) and replaces it only after the `/api/settings` fetch resolves.

Timeline on a fresh authenticated load:
1. Render — `status = "authenticated"`, `userSettings.weekStartDay = 0` (default).
2. Migration effect fires. `localValue = 1` (Monday from localStorage), `serverValue = 0` (still default).
3. Condition `localValue !== serverValue` is true → PUT fires, flag is set.
4. `/api/settings` GET resolves → `userSettings.weekStartDay = 1` (already saved — no problem *this time*).

The functional impact is mostly a redundant PUT in the case where the server already had the user's preference (e.g. they migrated on another browser earlier). The comment "promote local value to server" is also misleading because the migration cannot distinguish "user explicitly chose Sunday" from "no prior preference" when the flag hasn't been set yet.

Recommendation: Gate the migration effect on `!isLoading` from `useUserSettings` so `serverValue` is the real server value, not the in-memory default.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -331,11 +340,83 @@ export function CalendarProvider({
...
+        method: "PUT",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({ weekStartDay: day }),
+      }).catch((error) => {
+        logger.error(error as Error, {
+          context: "setWeekStartDay",
+          weekStartDay: day,
+        });
+      });
+    }
```
BenSeymourODB, line 396: **Missing rollback on PUT failure for `setWeekStartDay`**

`setWeekStartDay` optimistically updates React state and localStorage immediately, then fires a fire-and-forget PUT. On network failure the `.catch` only logs — it does not call `setWeekStartDayState(previousDay)` or `updateSettings({ weekStartDay: previousDay })`.

Compare this with `createEvent` / `deleteEvent` later in the file, which restore the optimistic entry on failure. The same pattern should apply here: capture `const previousDay = weekStartDay` before the update and restore it in the catch handler. Without a rollback, after a transient network error the UI shows Monday but the server stores Sunday — and the sync-down effect won't fire again until `userSettings.weekStartDay` changes (which it won't, because the GET still returns the old server value).

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -331,11 +340,83 @@ export function CalendarProvider({
...
+
+  // Once the server value is known and the migration has run, treat the
+  // server as the source of truth: if it diverges from local state, sync
+  // local state down. This is what makes a change made on a different
+  // browser show up here without a manual reload.
+  useEffect(() => {
+    if (status !== "authenticated") return;
+    if (!weekStartMigrationRef.current) return;
+    if (userSettings.weekStartDay !== weekStartDay) {
+      setWeekStartDayState(userSettings.weekStartDay);
```
BenSeymourODB: **Defensive clamp on the sync-down assignment**

`userSettings.weekStartDay` is typed `TWeekStartDay` (i.e. `0 | 1`), but the type only holds because `pickCalendarFields` filters rogue server values. Adding the same clamp the page-level code uses keeps the intent explicit and consistent:

```ts
const safe = userSettings.weekStartDay === 1 ? 1 : 0;
setWeekStartDayState(safe);
```

Low priority — the hook already guards. Filing for symmetry with `src/app/settings/page.tsx`.

---

src/components/providers/__tests__/CalendarProvider.test.tsx
```diff
@@ -1640,3 +1640,159 @@ describe("CalendarProvider", () => {
...
+        typeof init === "object" &&
+        init !== null &&
+        (init as RequestInit).method === "PUT"
+    );
+    expect(settingsPuts).toHaveLength(1);
+    const body = JSON.parse(String(settingsPuts[0][1].body));
+    expect(body).toEqual({ weekStartDay: 1 });
+  });
+
+  it("does not migrate when local matches server (both default Sunday)", async () => {
```
BenSeymourODB, line 1735: **Test gap — missing case: local=1, server=1 (non-default match)**

The existing "does not migrate when local matches server" test only covers the default-default case (both `0`). There is no test for the case where the user has already set Monday on this browser *and* on the server (e.g. after a prior migration succeeded on another browser). In that case the effect should see `localValue === serverValue` and skip the PUT.

This matters because the bug in the race-condition comment can manifest here: if the migration effect fires before `useUserSettings` resolves (when `serverValue` is still the in-memory default `0`), local=1 looks different from server=0 and the PUT fires unnecessarily. A test that stubs the fetch mock to return `weekStartDay: 1` *and* sets `calendar-settings` with `weekStartDay: 1` and verifies no PUT is issued would catch that race once the `!isLoading` gate is added.

---

src/components/providers/__tests__/CalendarProvider.test.tsx
```diff
@@ -1640,3 +1640,159 @@ describe("CalendarProvider", () => {
...
+        init !== null &&
+        (init as RequestInit).method === "PUT"
+      ) {
+        return init as RequestInit;
+      }
+    }
+    return undefined;
+  }
+
+  it("migrates a non-default localStorage weekStartDay to /api/settings exactly once", async () => {
```
BenSeymourODB, line 1700: **Test gap — no test for migration while `useUserSettings` is loading**

None of the four migration tests verify behaviour when `status === "authenticated"` but `useUserSettings` is still in-flight (i.e. `isLoading === true`). This is the window described in the race-condition comment where `userSettings.weekStartDay` is still the in-memory default. A test that delays the `/api/settings` GET response (e.g. using a deferred Promise) and asserts that no PUT is issued until the GET resolves would cover this path and prevent regressions if the migration dependency ordering is changed later.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: **Migration effect dependency on `settings.weekStartDay` causes re-runs**

The dependency array includes `settings.weekStartDay` (the localStorage value). Once the effect sets the flag and short-circuits on subsequent renders, the value is irrelevant — but its presence in the deps is surprising. Consider documenting why it's there (e.g. "covers the rare case where a separate tab updates the localStorage value before this tab finishes the migration") or removing it with an explanatory comment.

---

src/components/settings/display-section.tsx
```diff
@@ -88,6 +90,37 @@ export function DisplaySection({ values, onChange }: DisplaySectionProps) {
...
+              <input
+                type="radio"
+                name="weekStartDay"
+                value="1"
+                checked={values.weekStartDay === 1}
+                onChange={() => onChange({ weekStartDay: 1 })}
+                className="text-blue-600"
+              />
+              Monday
+            </Label>
```
BenSeymourODB, line 120: **Styling — raw `<input type="radio">` instead of shadcn `RadioGroup`**

The new "Week starts on" fieldset uses raw `<input type="radio">` with `text-blue-600`. `text-blue-600` has no visual effect on a native radio (would need `accent-blue-600`), and there's a precedent for using shadcn `RadioGroup`/`RadioGroupItem` elsewhere in the codebase (e.g. `CalendarSettingsPanel`).

That said, the existing Theme and Time-Format fields in this same component also use raw `<input>` radios with the same `text-blue-600` class, so this PR matches the file-local pattern. A separate cleanup PR could convert all three fields together. Flagging here so it's not lost.

---

BenSeymourODB (commented):
First-pass review by Sonnet sub-agent. Highlights:

1. **Race condition** in the migration effect — fires before `useUserSettings` resolves, comparing local against the in-memory default rather than the real server value. Mostly causes a redundant PUT, but the rationale comment is misleading. **Fixing.**
2. **No rollback** in `setWeekStartDay` PUT failure path — diverges from the `createEvent`/`deleteEvent` pattern in the same file. **Fixing.**
3. **Test gaps** for the non-default-match case and the still-loading case. **Fixing.**
4. **Styling note** on raw `<input>` radios — matches the existing file-local pattern; flagged for a separate cleanup PR.

Will push fixes for 1–3 and reply to 4–6 with rationale.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in cb2fc55. `useUserSettings` now exposes `hasLoadedFromServer` (set only after a successful `/api/settings` GET) and the migration effect short-circuits unless that flag is true. Two new regression tests cover the deferred-then-PUT and the defer-while-loading windows.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in cb2fc55. `setWeekStartDay` now captures the prior value, optimistically updates state + localStorage, and reverts both in the `.catch` handler if the PUT rejects. New regression test holds a deferred-rejection PUT open so the optimistic flip is observable, then asserts the post-rollback state.

While writing the test I also caught a related correctness bug: the sync-down effect's deps included `weekStartDay`, so every optimistic flip retriggered it and immediately reverted the local value back to the stale server value. Removed `weekStartDay` from the deps and made the effect a one-way "copy server → local on server change". The defensive `=== 1 ? 1 : 0` clamp comment 3 asked for is folded into the same fix.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Done in cb2fc55 — the sync-down effect now applies the `=== 1 ? 1 : 0` clamp before assignment. Same shape as the page-level guard.

---

src/components/providers/__tests__/CalendarProvider.test.tsx
BenSeymourODB: Added in cb2fc55 — see the new `does not migrate when local matches server (both Monday)` test. Stubs the fetch mock to return `weekStartDay: 1`, sets the localStorage `calendar-settings` to `weekStartDay: 1`, waits for the migration flag to flip, then asserts no PUT was issued. With the new `userSettingsLoaded` gate, this passes deterministically — without it, the test would have caught the race.

---

src/components/providers/__tests__/CalendarProvider.test.tsx
BenSeymourODB: Added in cb2fc55 — see `does not migrate while useUserSettings is still loading` and `defers migration until the server value is known, then PUTs`. Both hold the `/api/settings` GET open via a deferred `Promise`, assert no PUT is issued in the loading window, then resolve the GET and assert the expected outcome (no-op when local matches server, single PUT when it doesn't). Together they pin the `hasLoadedFromServer` gate against future regressions.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Removed in cb2fc55. `settings.weekStartDay` is no longer in the dep array; replaced with an inline `eslint-disable-next-line react-hooks/exhaustive-deps` plus a comment explaining why (the effect is mount-only after the flag is set, and reads the localStorage value imperatively). Effect now re-runs only on auth-state and server-value changes — both of which can legitimately bring new information.

---

src/components/settings/display-section.tsx
BenSeymourODB: Acknowledged but deliberately left as-is. The new fieldset matches the file-local pattern (Theme + Time Format also use raw `<input type="radio">` with `text-blue-600`). Converting all three to shadcn `RadioGroup`/`RadioGroupItem` is the right cleanup but expands the diff beyond the scope of #338.

Filing a follow-up issue would be the right next step — happy to do that as a separate one-line ticket if you want, otherwise leaving the trail here for someone working on a broader Settings UI polish PR.

---

BenSeymourODB: Fixed the in-scope half in a5cc29b: the new "Week starts on" radios now use `accent-blue-600` instead of the no-op `text-blue-600` (native radios take their color from `accent-color`, not `color`).

The broader cleanup — converting all three `DisplaySection` radio fields (Theme, Time Format, Week starts on) to shadcn `RadioGroup`/`RadioGroupItem` and dropping the residual `text-blue-600` on the other two — is filed as #373 to keep this PR scoped to #338.

---

PR #366, opened 2026-06-02
fix(settings): rollback only updated keys to preserve concurrent PUTs (#363)

BenSeymourODB:
## Summary

- Replace closure-snapshot rollback in `SettingsForm.updateSettings` with a partial-key revert sourced from the truly-current state via the functional setter — so an overlapping successful PUT to a different field is no longer wiped out when an earlier PUT fails.
- New `settings-form.test.tsx` reproduces the bug (theme PUT fails while timeFormat PUT succeeds) and pins the fix.
- Scoped: `updateTaskSettings` is introduced by PR #362 (not yet on `main`) and has the same shape — best fixed inside #362's review or as an immediate follow-up after it lands.

## Plan

No prior plan in `.claude/plans/`. Issue body laid out the repro and acceptance criteria; this PR implements the *partial-merge* variant of the proposed fix so the second acceptance criterion (concurrent successful PUT preserved) actually holds — the literal "snapshot the whole settings object" variant from the issue still wipes the concurrent change.

## Test plan

- [x] `pnpm test:run src/components/settings/__tests__/settings-form.test.tsx` — three new specs cover: single-PUT rollback, overlapping-PUTs preservation (the bug under fix), and the no-rollback-on-success sanity check.
- [x] `pnpm test:run` — full suite, 2022 tests passing.

## Notes for the reviewer

- Project board `Status` could not be flipped from this session (no GraphQL/projects MCP tool available); please update manually if you rely on board state.
- A follow-up issue (or a quick PR-#362 review comment) should apply the same `setSettings((curr) => { previousPartial = ...; return ... })` pattern to `updateTaskSettings` when it lands.

Closes #363

---
_Generated by [Claude Code](https://claude.ai/code/session_014eSrseECt2QtEyWXFhZxta)_

### DIFF
--- a/src/components/settings/__tests__/settings-form.test.tsx
+++ b/src/components/settings/__tests__/settings-form.test.tsx
@@ -363,3 +363,156 @@ describe("SettingsForm — Tasks settings wiring (#334)", () => {
     );
   });
 });
+
+/**
+ * Tests for SettingsForm optimistic-update rollback semantics (#363).
+ *
+ * The rollback path must restore the pre-call value of *only* the fields
+ * being updated, read from the truly-current state via the functional
+ * setter — not a stale closure of the entire settings object. Without
+ * this, an overlapping successful PUT to a different key loses its
+ * optimistic state when an earlier PUT to a different key fails.
+ */
+function deferred<T>() {
+  let resolve!: (value: T) => void;
+  let reject!: (reason: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+}
+
+describe("SettingsForm — updateSettings rollback", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockActiveProfile(ACTIVE_PROFILE_ID);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  it("rolls back to the pre-call value when a single PUT fails", async () => {
+    const user = userEvent.setup();
+    vi.stubGlobal(
+      "fetch",
+      vi.fn((url: string) => {
+        if (typeof url === "string" && url.startsWith("/api/profiles/")) {
+          return Promise.resolve({
+            ok: true,
+            json: async () => ({
+              taskSortOrder: "dueDate",
+              showCompletedTasks: false,
+            }),
+          } as unknown as Response);
+        }
+        return Promise.resolve({ ok: false } as Response);
+      })
+    );
+
+    renderSettings();
+
+    const lightRadio = screen.getByLabelText(/^light$/i);
+    const darkRadio = screen.getByLabelText(/^dark$/i);
+    expect(lightRadio).toBeChecked();
+
+    await user.click(darkRadio);
+
+    await waitFor(() => {
+      expect(toast.error).toHaveBeenCalledWith("Failed to save settings");
+    });
+
+    expect(lightRadio).toBeChecked();
+    expect(darkRadio).not.toBeChecked();
+  });
+
+  it("preserves a concurrent successful PUT when an earlier PUT fails", async () => {
+    const user = userEvent.setup();
+    const first = deferred<Response>();
+    const second = deferred<Response>();
+    let userSettingsCallIndex = 0;
+    vi.stubGlobal(
+      "fetch",
+      vi.fn((url: string) => {
+        // The profile-settings GET fires on mount; let it resolve immediately
+        // so the form is fully wired before we click. Only the two /api/settings
+        // PUTs are the ones under test.
+        if (typeof url === "string" && url.startsWith("/api/profiles/")) {
+          return Promise.resolve({
+            ok: true,
+            json: async () => ({
+              taskSortOrder: "dueDate",
+              showCompletedTasks: false,
+            }),
+          } as unknown as Response);
+        }
+        userSettingsCallIndex += 1;
+        return userSettingsCallIndex === 1 ? first.promise : second.promise;
+      })
+    );
+
+    renderSettings();
+
+    const lightRadio = screen.getByLabelText(/^light$/i);
+    const darkRadio = screen.getByLabelText(/^dark$/i);
+    const twelveHourRadio = screen.getByLabelText(/12-hour/i);
+    const twentyFourHourRadio = screen.getByLabelText(/24-hour/i);
+
+    await user.click(darkRadio);
+    await user.click(twentyFourHourRadio);
+
+    expect(darkRadio).toBeChecked();
+    expect(twentyFourHourRadio).toBeChecked();
+
+    second.resolve({ ok: true } as Response);
+    first.reject(new Error("network failure"));
+
+    await waitFor(() => {
+      expect(toast.error).toHaveBeenCalledWith("Failed to save settings");
+    });
+
+    // Theme reverts (its PUT failed)…
+    expect(lightRadio).toBeChecked();
+    expect(darkRadio).not.toBeChecked();
+    // …but the time-format change must NOT be wiped by the failed theme PUT's
+    // rollback — its own PUT succeeded.
+    expect(twentyFourHourRadio).toBeChecked();
+    expect(twelveHourRadio).not.toBeChecked();
+  });
+
+  it("does not roll back when the PUT succeeds", async () => {
+    const user = userEvent.setup();
+    const fetchMock = vi.fn((url: string) => {
+      if (typeof url === "string" && url.startsWith("/api/profiles/")) {
+        return Promise.resolve({
+          ok: true,
+          json: async () => ({
+            taskSortOrder: "dueDate",
+            showCompletedTasks: false,
+          }),
+        } as unknown as Response);
+      }
+      return Promise.resolve({ ok: true } as Response);
+    });
+    vi.stubGlobal("fetch", fetchMock);
+
+    renderSettings();
+
+    const darkRadio = screen.getByLabelText(/^dark$/i);
+    await user.click(darkRadio);
+
+    await waitFor(() => {
+      expect(fetchMock).toHaveBeenCalledWith(
+        "/api/settings",
+        expect.objectContaining({
+          method: "PUT",
+          body: JSON.stringify({ theme: "dark" }),
+        })
+      );
+    });
+
+    expect(darkRadio).toBeChecked();
+    expect(toast.error).not.toHaveBeenCalled();
+  });
+});
--- a/src/components/settings/settings-form.tsx
+++ b/src/components/settings/settings-form.tsx
@@ -123,6 +123,9 @@ export function SettingsForm({
     // to other keys is preserved (#363).
     let previousPartial: Partial<UserSettingsData> | undefined;
     setSettings((curr) => {
+      // The cast is sound because `Object.keys(partial)` is bounded by
+      // `partial`'s `Partial<UserSettingsData>` type — no extraneous keys
+      // can enter the snapshot.
       previousPartial = Object.fromEntries(
         (Object.keys(partial) as Array<keyof UserSettingsData>).map((key) => [
           key,
--- a/src/components/settings/settings-form.tsx
+++ b/src/components/settings/settings-form.tsx
@@ -117,8 +117,20 @@ export function SettingsForm({
   }, [activeProfileId]);
 
   const updateSettings = async (partial: Partial<UserSettingsData>) => {
-    const updated = { ...settings, ...partial };
-    setSettings(updated);
+    // Capture the pre-call value of *only* the keys we're about to overwrite,
+    // read from the functional setter's current state (not a stale closure).
+    // On rollback, merge those keys back so any concurrent successful PUT
+    // to other keys is preserved (#363).
+    let previousPartial: Partial<UserSettingsData> | undefined;
+    setSettings((curr) => {
+      previousPartial = Object.fromEntries(
+        (Object.keys(partial) as Array<keyof UserSettingsData>).map((key) => [
+          key,
+          curr[key],
+        ])
+      ) as Partial<UserSettingsData>;
+      return { ...curr, ...partial };
+    });
 
     try {
       const response = await fetch("/api/settings", {
@@ -137,7 +149,10 @@ export function SettingsForm({
       emitUserSettingsChange(partial);
     } catch {
       toast.error("Failed to save settings");
-      setSettings(settings);
+      const revert = previousPartial;
+      if (revert) {
+        setSettings((curr) => ({ ...curr, ...revert }));
+      }
     }
   };
 


src/components/settings/settings-form.tsx
```diff
@@ -66,8 +66,20 @@ export function SettingsForm({
   }, []);
 
   const updateSettings = async (partial: Partial<UserSettingsData>) => {
-    const updated = { ...settings, ...partial };
-    setSettings(updated);
+    // Capture the pre-call value of *only* the keys we're about to overwrite,
+    // read from the functional setter's current state (not a stale closure).
+    // On rollback, merge those keys back so any concurrent successful PUT
+    // to other keys is preserved (#363).
+    let previousPartial: Partial<UserSettingsData> | undefined;
```
BenSeymourODB, line 124: **`previousPartial` timing: is it reliably populated before `catch` runs? Yes.**

The assignment `previousPartial = Object.fromEntries(...)` happens synchronously inside the `setSettings` updater function. React calls updater functions synchronously before the `setState` call itself returns — this is true in both React 18 and React 19 (the updater is invoked eagerly during the batched-update flush that happens before the next microtask tick). The `await fetch(...)` that follows is a macrotask boundary, so by the time any rejection can propagate to `catch`, the updater has already run and `previousPartial` is set.

The `let previousPartial: ... | undefined` + `if (revert)` guard is safe: the only path to `catch` is after `setSettings(...)` has been called, so `revert` is always truthy at that point. The guard is defensive but harmless.

---

src/components/settings/settings-form.tsx
```diff
@@ -81,7 +93,10 @@ export function SettingsForm({
       }
     } catch {
       toast.error("Failed to save settings");
-      setSettings(settings);
+      const revert = previousPartial;
+      if (revert) {
+        setSettings((curr) => ({ ...curr, ...revert }));
```
BenSeymourODB, line 157: **Concurrent PUT to the same key: rollback ordering is last-write-wins, which is correct.**

If two PUTs race on the *same* key (e.g. the user clicks "Dark" then immediately clicks "System" before the first fetch resolves):

- PUT-1 saves `previousPartial = { theme: "light" }` and optimistically sets `theme → "dark"`.
- PUT-2 saves `previousPartial = { theme: "dark" }` (read from the functional setter which sees the already-optimistic "dark") and optimistically sets `theme → "system"`.

If PUT-1 fails and PUT-2 succeeds: PUT-1's catch reverts to `"light"` via `setSettings((curr) => ({ ...curr, theme: "light" }))`. That wipes the successful "system". This is an existing race condition that the partial-key approach does not fully solve for same-key overlap.

However: (a) this scenario requires two concurrent mutations of the same field, which is nearly impossible through normal UI interaction since each section only exposes one radio group / input, and (b) the issue's acceptance criteria only required preserving *different-key* concurrent PUTs — same-key is out of scope. The PR correctly does not over-engineer this. Just documenting it so it's a known, accepted limitation.

---

src/components/settings/__tests__/settings-form.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+
+    await waitFor(() => {
+      expect(mockToastError).toHaveBeenCalledWith("Failed to save settings");
+    });
+
+    expect(lightRadio).toBeChecked();
+    expect(darkRadio).not.toBeChecked();
+  });
+
+  it("preserves a concurrent successful PUT when an earlier PUT fails", async () => {
```
BenSeymourODB, line 430: **Does this test actually fail on the old code? Yes — deterministically.**

Old rollback: `setSettings(settings)` where `settings` is the stale closure captured at the time `updateSettings` was defined (i.e., the initial render's `settings` value).

Walking through the old code with this test:
1. `updateSettings({ theme: "dark" })` — optimistic set. `settings` in the closure still equals `baseSettings` (the render hasn't flushed with new values yet from React's perspective when the closure was created — or more precisely, the closure captured the `settings` value from the render in which `updateSettings` was defined, which is the initial render: `{ theme: "light", timeFormat: "12h", ... }`).
2. `updateSettings({ timeFormat: "24h" })` — optimistic set to `"24h"`. The closure's `settings` still references the initial render value `{ ..., timeFormat: "12h", ... }`.
3. Second PUT resolves OK — no rollback.
4. First PUT rejects — old code calls `setSettings(settings)` where `settings` = initial `baseSettings` snapshot = `{ theme: "light", timeFormat: "12h", ... }`. This wipes `timeFormat` back to `"12h"`, and the test assertion `expect(twentyFourHourRadio).toBeChecked()` fails.

The test is a true discriminator.

---

src/components/settings/__tests__/settings-form.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    const darkRadio = screen.getByLabelText(/^dark$/i);
+    const twelveHourRadio = screen.getByLabelText(/12-hour/i);
+    const twentyFourHourRadio = screen.getByLabelText(/24-hour/i);
+
+    await user.click(darkRadio);
+    await user.click(twentyFourHourRadio);
+
+    expect(darkRadio).toBeChecked();
+    expect(twentyFourHourRadio).toBeChecked();
+
```
BenSeymourODB, line 467: **Deferred-promise test: is it deterministic? Yes, with one caveat worth knowing.**

The `deferred()` helper correctly externalises the resolve/reject handles. The ordering `second.resolve(ok) → first.reject(error)` is deliberate — it settles the later PUT first so that by the time the first PUT's catch runs, the successful change is already in state. This is the exact scenario from issue #363.

One minor point: `userEvent.setup()` with no explicit `delay` option defaults to `delay: null` in recent versions (typing is instantaneous). The two `await user.click()` calls are sequential, so `callIndex` will be 1 then 2 in order. This is reliable.

The only theoretical flakiness would be if React batched both `setSettings` calls from the two `updateSettings` invocations into a single render before either `fetch` call ran, but that can't happen here because `userEvent.click` `await`s the full event cycle (including the microtask that dispatches `updateSettings`'s `setSettings` + starts `fetch`) before returning. Both fetches are started before either deferred resolves.

---

src/components/settings/__tests__/settings-form.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    // Theme reverts (its PUT failed)…
+    expect(lightRadio).toBeChecked();
+    expect(darkRadio).not.toBeChecked();
+    // …but the time-format change must NOT be wiped by the failed theme PUT's
+    // rollback — its own PUT succeeded.
+    expect(twentyFourHourRadio).toBeChecked();
+    expect(twelveHourRadio).not.toBeChecked();
+  });
+
+  it("does not roll back when the PUT succeeds", async () => {
```
BenSeymourODB, line 484: **Missing assertion: success path doesn't verify `fetchMock` was called with the right body.**

Test 3 waits for `fetchMock` to have been called but never asserts *what* it was called with. For a regression suite pinning a specific bug fix, it would be useful to at least assert:

```ts
expect(fetchMock).toHaveBeenCalledWith("/api/settings", expect.objectContaining({
  method: "PUT",
  body: JSON.stringify({ theme: "dark" }),
}));
```

Not a blocker — the test's primary job is "no rollback on success," which it does correctly. Just a suggestion to tighten it.

---

src/components/settings/__tests__/settings-form.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+  let reject!: (reason: unknown) => void;
+  const promise = new Promise<T>((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+  return { promise, resolve, reject };
+}
+
+describe("SettingsForm — updateSettings rollback", () => {
+  beforeEach(() => {
```
BenSeymourODB, line 387: **Mock hygiene: `vi.clearAllMocks()` in `beforeEach` + `vi.unstubAllGlobals()` in `afterEach` — correct, no leakage.**

`vi.clearAllMocks()` resets call counts/results on existing mocks; `vi.unstubAllGlobals()` tears down the `fetch` stub. The order (clear before each, unstub after each) means each test gets a clean `fetch` and clean call counts. No leakage between tests.

One micro-nit: `mockToastSuccess` is hoisted and mocked but never asserted. It doesn't cause any problem (it just prevents console noise if the component ever calls `toast.success`), but it could be removed if it's not needed. Leave it — it's future-proof.

---

src/components/settings/settings-form.tsx
```diff
@@ -66,8 +66,20 @@ export function SettingsForm({
   }, []);
 
   const updateSettings = async (partial: Partial<UserSettingsData>) => {
```
BenSeymourODB, line 119: **CLAUDE.md compliance check: all green.**

- **TDD ordering**: Tests are in the new file added in this commit; the production change is minimal and follows the test spec. The diff shows no code path that isn't covered by at least one of the three specs.
- **No manual memoization**: `updateSettings` is a plain `async` function inside the component body — no `useCallback`. Correct per React Compiler policy.
- **Strict TS / no `any`**: `Object.fromEntries(...) as Partial<UserSettingsData>` is the only cast, and it's necessary because `fromEntries` returns `{[k: string]: V}`. Clean.
- **No nested-object issue**: `UserSettingsData` is flat (all primitives). The `Object.keys` + `map` snapshot is correct for this shape. If the interface ever gains nested objects, a shallow clone would still be sufficient for a one-level partial — documenting this as a non-issue today.
- **No `undefined` value edge case**: `partial` keys are always present in `UserSettingsData` (the callers pass typed partials), so `curr[key]` is never `undefined` for a valid key. Safe.

---

BenSeymourODB (commented):
## Review: fix(settings) rollback only updated keys (#363)

**Verdict: LGTM — ship it** (one non-blocking suggestion below).

### Core fix correctness

The partial-key revert is the right choice over the issue's "snapshot the whole settings object" proposal. The proposed literal would still wipe any concurrent write that landed between the first `setSettings` and the failing PUT's catch, because it snapshots the *pre-optimistic* full state. The partial-merge approach correctly restricts the rollback to exactly the keys that were attempted, using a functional-setter read to get the truly-current value — not a stale closure.

**`previousPartial` timing:** Assignment happens synchronously inside the `setSettings` updater, which React calls before the call returns. The `await fetch(...)` that follows is a macrotask boundary, so `previousPartial` is always populated before any rejection reaches `catch`. The `if (revert)` guard is defensive but harmless.

**Same-key concurrent PUT:** Two concurrent mutations of the *same* key still race (last-revert wins). This is a known, accepted limitation; it's practically unreachable through normal UI (each section controls a disjoint set of fields) and is out of scope for issue #363's acceptance criteria.

**Nested objects / `undefined` values:** `UserSettingsData` is fully flat today. No issue.

### Tests

All three specs are deterministic and correctly discriminate the bug. The deferred-promise scenario is reliable because `userEvent.click` awaits the full event cycle before returning, ensuring both fetches are in-flight before either is settled. Mock hygiene is clean (`vi.clearAllMocks` + `vi.unstubAllGlobals`). The concurrent-key test provably fails on the old code: the old `setSettings(settings)` rollback used the initial render's closure value, which included `timeFormat: "12h"`, wiping the successful 24h change.

**One non-blocking suggestion:** Test 3 (success path) could assert `fetchMock` was called with the correct body — see inline comment on line 161.

### CLAUDE.md

All checks pass: no `useCallback`/`useMemo`, strict TS, standard Tailwind (no styling changes), TDD ordering visible in the diff, zero `any`.

### Follow-up noted

The PR description correctly calls out that the same pattern needs to be applied to `updateTaskSettings` introduced in #362. That can be handled in that PR's review or as an immediate follow-up.

---

src/components/settings/settings-form.tsx
BenSeymourODB: Acknowledged — no code change needed. The synchronous-updater + macrotask-boundary reasoning matches my model; the `if (revert)` stays as defensive code (cheap, clarifies intent for future readers).

---

src/components/settings/settings-form.tsx
BenSeymourODB: Accepted limitation, no action. Same-key concurrent PUTs are out of scope per #363 and require a per-field sequence-ID design (also called out in the issue body) that would be over-engineering today.

---

src/components/settings/__tests__/settings-form.test.tsx
BenSeymourODB: Confirmed — walking the old code through the test gives exactly the trace described; the spec is a real discriminator (it also failed in the TDD-red step before the fix was applied).

---

src/components/settings/__tests__/settings-form.test.tsx
BenSeymourODB: Acknowledged. The two sequential `await user.click()` calls guarantee both fetches are in flight before either deferred resolves; no batching can collapse the scenario.

---

src/components/settings/__tests__/settings-form.test.tsx
BenSeymourODB: Addressed in 365be8b — test 3 now asserts the PUT URL, method, and serialised partial body.

---

src/components/settings/__tests__/settings-form.test.tsx
BenSeymourODB: Acknowledged. Keeping `mockToastSuccess` for future-proofing as suggested — costs nothing and prevents accidental console noise if `toast.success` is wired in later.

---

src/components/settings/settings-form.tsx
BenSeymourODB: Acknowledged. The `Object.fromEntries(...) as Partial<UserSettingsData>` cast is the minimum needed; if the interface ever gains nested objects, a focused refactor of the partial-snapshot helper would be cleaner than embedding deep-clone logic here.

---

BenSeymourODB:
🤖 `implement-issue` follow-up: rebased onto current `main` to clear the `dirty` mergeable state. The PR was logically ready (CI green at 365be8b, the prior in-depth review LGTM'd it, all seven review threads have follow-up replies) — only blocker was the conflict against PR #354 (`makeUserSettings` fixture) and PR #362 (`useProfile` import + profile-scoped task settings) both having landed on `main` since this branch was cut.

### What changed in the rebase

**`src/components/settings/settings-form.tsx`** — auto-merged cleanly. The `updateSettings` partial-key revert is preserved verbatim; it slots in alongside main's new `useProfile` / `taskSettings` wiring without conflict.

**`src/components/settings/__tests__/settings-form.test.tsx`** — manual add/add resolution. Both branches independently created this file:
- Main (commit `dd9e995`, PR #362): 8 specs for profile-scoped Tasks settings wiring (#334), built on `makeUserSettings()` + `vi.mocked(useProfile)`.
- This PR: 3 specs for the `updateSettings` rollback semantics (#363).

Resolved by keeping main's full file as the base and appending the three rollback specs as a new `describe("SettingsForm — updateSettings rollback", …)` block, adapted to:
- Reuse `makeUserSettings()` / `renderSettings()` / `mockActiveProfile(ACTIVE_PROFILE_ID)` from the top of the file (matches the #371 factory-pattern direction).
- Use the existing `vi.mock("sonner", …)` + `import { toast } from "sonner"` instead of `vi.hoisted` toast vars.
- Let the profile-settings GET on mount short-circuit through the same `fetch` stub so the profile-scoped effect doesn't interfere with the `/api/settings` PUTs under test.

The second commit (`365be8b`, the test-3 PUT-body assertion from the prior review) is dropped automatically by `git rebase` — its patch is already upstream because the resolved test file includes the tightened form of test 3 directly. The pre-rebase two-commit history is preserved logically (review thread → fix on `365be8b`) and remains visible on the PR's "Commits" tab via the force-with-lease push notice.

### Verification on the rebased HEAD (`a196cc3`)

- `pnpm test:run src/components/settings/__tests__/settings-form.test.tsx` → **11/11 passing** (8 inherited from main + 3 new rollback specs)
- `pnpm test:run` → **2333/2333 passing** across 146 files
- `pnpm lint:fix` → 0 errors, 6 pre-existing warnings unrelated to this PR
- `pnpm format:fix` → clean
- `pnpm check-types` → clean

`mergeable_state` flipped from `dirty` → `unstable` post-push; should settle to `clean` once CI completes.

### Note on history rewrite

The rebase required `--force-with-lease` to update `claude/brave-maxwell-0TqrM`. No other branch shares this ref; `--force-with-lease` (not `--force`) refuses if anything pushed since my fetch. Flagging in case the project convention is strict no-rewrite — happy to redo as a merge commit instead.

---

PR #367, opened 2026-06-03
fix(scheduler): stability and correctness bugs from PR #50 self-review (#221)

BenSeymourODB:
## Summary

Addresses the stability and correctness bugs catalogued in #221 (the post-PR-#50 self-review). Bug 7 (the `ScreenTransition` AnimatedSwap-style refactor) is deferred to the follow-up issue #368.

| # | Bug | Status |
|---|---|---|
| 1 | Unstable `controls` object | **Fixed here** — stable identity via `useState` initializer + commit-phase `useRef` mirror. Compatible with React Compiler (no `useMemo`/`useCallback`). |
| 2 | Manual-unpause lockout while interaction detection active | **Fixed here** — `useInteractionDetector` gains an `ignoreRef` and a `reset()` callback; `ScreenScheduler` wraps the nav controls in a ref'd div and calls `resetInteractionPause()` on every manual resume (button + Space key). |
| 3 | Midnight wraparound in `isTimeMatch` | Already fixed on `main`; existing tests cover it. |
| 4 | Time-specific check skips initial run | Already fixed; **regression test added here**. |
| 5 | Module-level mutable `idCounter` | Already fixed (uses `crypto.randomUUID()`); **regression test added here**. |
| 6 | Countdown drift from separate `setInterval` | Already fixed (single 1s tick is the source of truth); **regression test added here**. |
| 7 | `ScreenTransition` geometry / timer-race | **Deferred** to #368. Substantial refactor mirroring PR #156's `AnimatedSwap` work; warrants its own implementation and review cycle. |

## Plan

Linked plan: `.claude/plans/scheduler-stability-bugs-221.md`

## Test plan

- [x] New regression tests for already-fixed bugs (4, 5, 6) pass.
- [x] New tests for bug 1 (stable `controls` identity, stable individual function identities, wrappers observe latest state) pass.
- [x] New tests for bug 2 (`ignoreRef` filtering, `reset()` callback, end-to-end ScreenScheduler resume-button behaviour) pass.
- [x] Full scheduler suite green (150 tests).
- [x] Full project suite green (2032 tests).

## Deferred

- Bug 7 → #368 (refactor `ScreenTransition` to mirror `AnimatedSwap`).

Closes #221 (bugs 1–6); bug 7 tracked separately in #368.

### DIFF
--- a/src/components/scheduler/__tests__/screen-scheduler.test.tsx
+++ b/src/components/scheduler/__tests__/screen-scheduler.test.tsx
@@ -184,6 +184,37 @@ describe("ScreenScheduler", () => {
         expect.stringMatching(/next screen in/i)
       );
     });
+
+    it("pressing ArrowRight while interaction-paused resumes rotation", () => {
+      render(
+        <ScreenScheduler config={defaultConfig} autoStart>
+          <div data-testid="content">Content</div>
+        </ScreenScheduler>
+      );
+
+      // Trigger an interaction-pause via a click outside the controls.
+      act(() => {
+        fireEvent.click(screen.getByTestId("content"));
+      });
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        "Screen rotation paused"
+      );
+
+      // Press ArrowRight to manually navigate. The keydown event fires
+      // on window — outside the controls container — and so without the
+      // explicit reset would extend the interaction-pause. The fix
+      // calls `resetInteractionPause()` in the arrow-key handlers, so
+      // navigation should resume rotation.
+      act(() => {
+        fireEvent.keyDown(window, { key: "ArrowRight" });
+      });
+
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        expect.stringMatching(/next screen in/i)
+      );
+    });
   });
 
   it("applies transition config from schedule config", () => {
--- a/src/components/scheduler/__tests__/screen-scheduler.test.tsx
+++ b/src/components/scheduler/__tests__/screen-scheduler.test.tsx
@@ -4,8 +4,8 @@
  * Tests rendering of children and navigation controls integration.
  */
 import type { ScheduleConfig } from "@/components/scheduler/types";
-import { render, screen } from "@testing-library/react";
-import { describe, expect, it, vi } from "vitest";
+import { act, fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import { ScreenScheduler } from "../screen-scheduler";
 
 // Mock next/navigation
@@ -112,6 +112,78 @@ describe("ScreenScheduler", () => {
 
     expect(screen.getByTestId("screen-transition")).toBeInTheDocument();
     expect(screen.getByTestId("child-content")).toBeInTheDocument();
+  });
+
+  // Regression test for issue #221, bug 2: clicking the scheduler's
+  // own pause/resume button used to immediately re-pause the scheduler
+  // because the window-level interaction listener treated the click as
+  // generic user activity. The fix exempts the navigation controls'
+  // container subtree from the interaction detector.
+  describe("manual unpause via toggle button (#221, bug 2)", () => {
+    beforeEach(() => {
+      vi.useFakeTimers();
+    });
+
+    afterEach(() => {
+      vi.useRealTimers();
+    });
+
+    it("a click outside the controls subtree pauses the scheduler", () => {
+      render(
+        <ScreenScheduler config={defaultConfig} autoStart>
+          <div data-testid="content">Content</div>
+        </ScreenScheduler>
+      );
+
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        expect.stringMatching(/next screen in/i)
+      );
+
+      // Click on page content (outside navigation controls subtree).
+      act(() => {
+        fireEvent.click(screen.getByTestId("content"));
+      });
+
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        "Screen rotation paused"
+      );
+    });
+
+    it("a click on the pause/resume button does NOT re-pause the scheduler", () => {
+      render(
+        <ScreenScheduler config={defaultConfig} autoStart>
+          <div data-testid="content">Content</div>
+        </ScreenScheduler>
+      );
+
+      // First, pause via a click outside the controls.
+      act(() => {
+        fireEvent.click(screen.getByTestId("content"));
+      });
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        "Screen rotation paused"
+      );
+
+      // Now click the toggle button to resume. The click happens inside
+      // the controls subtree, so the interaction detector must ignore
+      // it; the scheduler should be unpaused and STAY unpaused.
+      const toggleButton = screen.getByRole("button", {
+        name: /resume rotation/i,
+      });
+      act(() => {
+        fireEvent.click(toggleButton);
+      });
+
+      // Without the fix, the click would have re-set the
+      // interaction-pause and the status would still read paused.
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        expect.stringMatching(/next screen in/i)
+      );
+    });
   });
 
   it("applies transition config from schedule config", () => {
--- a/src/components/scheduler/__tests__/use-screen-scheduler.test.ts
+++ b/src/components/scheduler/__tests__/use-screen-scheduler.test.ts
@@ -333,33 +333,45 @@ describe("useScreenScheduler", () => {
     // its own setInterval that drifted from the navigation interval. The
     // fix unifies them into a single 1s tick — the countdown is derived
     // from the same timer that triggers navigation, eliminating drift by
-    // construction. This test pins that invariant: while rotation is
-    // running, exactly one 1000ms interval is registered.
+    // construction. This test pins that invariant via both the
+    // structural shape (one 1s interval) and the behavioural one (the
+    // navigation fires after `intervalSeconds` ticks).
     it("uses a single 1s interval as the shared source of truth for countdown and navigation", () => {
       mockPathname.mockReturnValue("/calendar");
       const setIntervalSpy = vi.spyOn(globalThis, "setInterval");
 
-      const { result } = renderHook(() => useScreenScheduler(defaultConfig));
+      try {
+        const { result } = renderHook(() => useScreenScheduler(defaultConfig));
 
-      act(() => {
-        result.current.controls.start();
-      });
+        act(() => {
+          result.current.controls.start();
+        });
 
-      // Drain the immediate time-specific check.
-      act(() => {
-        vi.advanceTimersByTime(0);
-      });
+        // Drain the immediate time-specific check.
+        act(() => {
+          vi.advanceTimersByTime(0);
+        });
 
-      const intervalDelays = setIntervalSpy.mock.calls.map((call) => call[1]);
-      const oneSecondIntervals = intervalDelays.filter((d) => d === 1000);
-      const sixtySecondIntervals = intervalDelays.filter((d) => d === 60_000);
+        const intervalDelays = setIntervalSpy.mock.calls.map((call) => call[1]);
+        const oneSecondIntervals = intervalDelays.filter((d) => d === 1000);
+        const sixtySecondIntervals = intervalDelays.filter((d) => d === 60_000);
 
-      // Exactly one 1s tick (drives countdown AND navigation).
-      expect(oneSecondIntervals).toHaveLength(1);
-      // Plus the unrelated 60s time-specific check interval.
-      expect(sixtySecondIntervals).toHaveLength(1);
+        // Exactly one 1s tick (drives countdown AND navigation).
+        expect(oneSecondIntervals).toHaveLength(1);
+        // Plus the unrelated 60s time-specific check interval.
+        expect(sixtySecondIntervals).toHaveLength(1);
 
-      setIntervalSpy.mockRestore();
+        // Behavioural pin: that single 1s interval drives navigation at
+        // exactly `intervalSeconds * 1000`ms — no separate timer fires
+        // navigation early, none fires late.
+        act(() => {
+          vi.advanceTimersByTime(60_000);
+        });
+        expect(mockPush).toHaveBeenCalledTimes(1);
+        expect(mockPush).toHaveBeenCalledWith("/recipe");
+      } finally {
+        setIntervalSpy.mockRestore();
+      }
     });
 
     it("does not auto-navigate when externalPaused is true", () => {
--- a/src/components/scheduler/__tests__/use-screen-scheduler.test.ts
+++ b/src/components/scheduler/__tests__/use-screen-scheduler.test.ts
@@ -462,6 +462,99 @@ describe("useScreenScheduler", () => {
     });
   });
 
+  // Regression tests for issue #221, bug 1: the `controls` object was
+  // recreated as a fresh literal on every render, so consumers that
+  // listed `controls` in their `useEffect` dependency array re-ran their
+  // effects on every internal state change — including the 1-second
+  // countdown tick.
+  describe("stable controls reference (#221, bug 1)", () => {
+    it("returns the same controls reference across renders triggered by state changes", () => {
+      mockPathname.mockReturnValue("/calendar");
+      const { result } = renderHook(() => useScreenScheduler(defaultConfig));
+
+      const initialControls = result.current.controls;
+
+      // Trigger an internal state change: start the scheduler.
+      act(() => {
+        result.current.controls.start();
+      });
+
+      expect(result.current.controls).toBe(initialControls);
+
+      // Trigger another internal state change: advance one countdown tick.
+      act(() => {
+        vi.advanceTimersByTime(1000);
+      });
+
+      expect(result.current.controls).toBe(initialControls);
+
+      // Pause / resume cycles must not break identity either.
+      act(() => {
+        result.current.controls.pause();
+      });
+      expect(result.current.controls).toBe(initialControls);
+
+      act(() => {
+        result.current.controls.resume();
+      });
+      expect(result.current.controls).toBe(initialControls);
+    });
+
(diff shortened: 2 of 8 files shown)


src/components/scheduler/use-screen-scheduler.ts
```diff
@@ -276,14 +276,41 @@ export function useScreenScheduler(
...
+  const liveControlsRef = useRef<SchedulerControls>({
     start,
     stop,
     pause,
     resume,
     navigateToNext,
     navigateToPrevious,
-  };
+  });
+  useEffect(() => {
```
BenSeymourODB, line 295: **[Medium – Correctness / Timing] First-render stale-ref window before first commit**

`liveControlsRef` is initialized with render-time closures at line 287–294, then refreshed by this `useEffect` (which runs *after* the commit). On the very first render the ref is correct. On every subsequent render, there is a window between re-render and this effect running where any synchronous call to a stable wrapper dispatches through stale closures from the *previous* render.

In practice the stable wrappers are only called from user events and other effects — both of which run post-commit — so the window is safe today. However the pattern is identical to the flawed "write ref during render" anti-pattern except it's deferred by one tick, making it subtly fragile if a future effect is added that calls a control synchronously during commit.

The canonical React pattern avoids this entirely by writing the ref **during render** (not during a commit-phase effect), which React Compiler's ESLint plugin allows when the write is unconditional and not in event handlers. If the ESLint rule is blocking a render-phase write here, a comment explaining why the effect was chosen over a render-phase write would clarify intent:

```ts
// Write during render is safe: unconditional, same component.
// React Compiler ESLint: if this triggers, add a suppression comment here.
liveControlsRef.current = { start, stop, pause, resume, navigateToNext, navigateToPrevious };
```

Not a blocker for this PR given the current call sites, but worth noting for future-proofing.

---

src/components/scheduler/use-interaction-detector.ts
```diff
@@ -42,9 +57,18 @@ const INTERACTION_EVENTS = [
...
+  const { pauseDurationMs, enabled, ignoreRef } = options;
   const [isPaused, setIsPaused] = useState(false);
   const timeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  // Mirror `ignoreRef` into an internal ref so the effect can read the
+  // latest value without re-subscribing listeners when the ref identity
+  // changes (refs are stable, but the prop slot is optional).
+  const ignoreRefMirror = useRef<RefObject<HTMLElement | null> | undefined>(
+    ignoreRef
+  );
+  useEffect(() => {
```
BenSeymourODB: **[Low – Complexity / Unnecessary indirection] `ignoreRefMirror` is over-engineering for this call site**

The goal of this mirror is to let `handleInteraction` read the current `ignoreRef` without being in the `useEffect` dep array (which would cause listeners to re-subscribe). That's a valid concern in the general case.

However, `controlsContainerRef` is created once with `useRef` in `ScreenScheduler` — the *object identity* of the `RefObject` never changes between renders. So `ignoreRef` (the prop passed in) is always the same object reference. The mirror adds a one-render lag (the `useEffect` keeping `ignoreRefMirror.current` fresh runs after the first render) and introduces a ref-of-a-ref that is harder to reason about.

Simpler equivalent that is safe for this specific caller and for any stable `RefObject`:

```ts
// ignoreRef is a stable useRef — its object identity never changes.
// Capture it directly in the closure; no mirror needed.
const ignoreEl = ignoreRef?.current;
```

If the hook needs to support dynamically-swapped `RefObject` instances (not the current use case), the mirror approach is justified — but that should be documented. As-is the comment says "refs are stable, but the prop slot is optional", which is the motivation, yet the opt-in prop slot never changes identity in this codebase. Consider simplifying or adding a `// future-proofing: ignoreRef identity could in theory change` comment.

---

src/components/scheduler/screen-scheduler.tsx
```diff
@@ -127,6 +134,10 @@ export function ScreenScheduler({
           e.preventDefault();
```
BenSeymourODB, line 140: **[Medium – UX Bug] ArrowLeft/ArrowRight don't clear the interaction-pause — same class of bug as toggle-button issue**

Space on resume calls `resetInteractionPause()` (line 140), but ArrowLeft (line 127) and ArrowRight (line 130) do not. This means:

1. User scrolls/clicks something → interaction-pause fires.
2. User presses ArrowRight to manually advance — navigation fires, but `isInteractionPaused` stays `true`, so `effectivelyPaused` remains `true` and rotation stays paused until the timer expires.

Worse: the ArrowRight keydown event itself is on `INTERACTION_EVENTS` and is dispatched on `window`, so its target is outside `controlsContainerRef`. That keydown event will immediately trigger a new interaction-pause, resetting the timer. The user is stuck in interaction-pause even after explicitly pressing an arrow key.

Same logic applies to the `SchedulerStatusIndicator` — if a user interacts with it (it's outside `controlsContainerRef`) it'll re-pause as expected, but arrow keys from keyboard should be treated as explicit commands like Space.

Suggested fix:

```ts
case "ArrowLeft":
  controls.navigateToPrevious();
  resetInteractionPause();
  break;
case "ArrowRight":
  controls.navigateToNext();
  resetInteractionPause();
  break;
```

Should be covered by a test: press ArrowRight while interaction-paused → verify `effectivelyPaused` becomes false.

---

src/components/scheduler/use-interaction-detector.ts
```diff
@@ -84,5 +117,13 @@ export function useInteractionDetector(
     };
   }, [enabled, pauseDurationMs]);
 
-  return { isPaused };
+  const [reset] = useState(() => () => {
```
BenSeymourODB, line 113: **[Low – Test Gap] `reset()` has no direct unit test**

The `reset()` function is tested only transitively through the `screen-scheduler.test.tsx` integration test. There's no focused unit test that directly exercises `reset()` in `use-interaction-detector.test.ts`.

Specifically missing:
- `reset()` while paused clears `isPaused` immediately.
- `reset()` cancels the pending resume timeout (i.e., calling `reset()` then advancing time by `pauseDurationMs` does not flip `isPaused` back to `true` via the old timeout).
- `reset()` when not paused is a no-op (doesn't throw).

The second case is important: without it, the existing code is verifiably correct but a future regression that calls `setIsPaused(true)` in the old timeout closure would go undetected.

Suggested addition to the `ignoreRef` describe block or its own block in `use-interaction-detector.test.ts`:

```ts
describe("reset()", () => {
  it("clears isPaused immediately", () => {
    const { result } = renderHook(() =>
      useInteractionDetector({ pauseDurationMs: 2000, enabled: true })
    );
    act(() => { window.dispatchEvent(new MouseEvent("click")); });
    expect(result.current.isPaused).toBe(true);
    act(() => { result.current.reset(); });
    expect(result.current.isPaused).toBe(false);
  });

  it("cancels the pending resume timeout", () => {
    const { result } = renderHook(() =>
      useInteractionDetector({ pauseDurationMs: 2000, enabled: true })
    );
    act(() => { window.dispatchEvent(new MouseEvent("click")); });
    act(() => { result.current.reset(); });
    act(() => { vi.advanceTimersByTime(2000); }); // old timeout fires
    expect(result.current.isPaused).toBe(false); // must still be false
  });
});
```

---

src/components/scheduler/__tests__/use-screen-scheduler.test.ts
```diff
@@ -282,6 +329,39 @@ describe("useScreenScheduler", () => {
...
+      const intervalDelays = setIntervalSpy.mock.calls.map((call) => call[1]);
+      const oneSecondIntervals = intervalDelays.filter((d) => d === 1000);
+      const sixtySecondIntervals = intervalDelays.filter((d) => d === 60_000);
+
+      // Exactly one 1s tick (drives countdown AND navigation).
+      expect(oneSecondIntervals).toHaveLength(1);
+      // Plus the unrelated 60s time-specific check interval.
+      expect(sixtySecondIntervals).toHaveLength(1);
+
+      setIntervalSpy.mockRestore();
```
BenSeymourODB: **[Low – Test Fragility] Bug 6 test: `setIntervalSpy.mockRestore()` is not in a `finally` block**

If either assertion on lines 357–360 fails, `setIntervalSpy.mockRestore()` on line 362 never runs. That leaks the spy into subsequent tests, which can cause confusing failures in tests that call `setInterval` and don't expect a spy.

```ts
// Wrap in try/finally:
try {
  // ... act + assertions ...
  expect(oneSecondIntervals).toHaveLength(1);
  expect(sixtySecondIntervals).toHaveLength(1);
} finally {
  setIntervalSpy.mockRestore();
}
```

Alternatively, move the spy to `beforeEach` and the restore to `afterEach` for the whole describe block.

---

src/components/scheduler/__tests__/use-screen-scheduler.test.ts
```diff
@@ -282,6 +329,39 @@ describe("useScreenScheduler", () => {
       expect(mockPush).not.toHaveBeenCalled();
     });
 
+    // Regression test for issue #221, bug 6: the countdown previously had
+    // its own setInterval that drifted from the navigation interval. The
+    // fix unifies them into a single 1s tick — the countdown is derived
+    // from the same timer that triggers navigation, eliminating drift by
+    // construction. This test pins that invariant: while rotation is
+    // running, exactly one 1000ms interval is registered.
+    it("uses a single 1s interval as the shared source of truth for countdown and navigation", () => {
```
BenSeymourODB, line 339: **[Low – Test Coverage Gap] Bug 6 test doesn't assert that navigation actually fires at the right time**

The test verifies the structural invariant (only one 1s interval registered), which is a good architectural pin. But it doesn't verify the *behavioral* invariant: that navigation fires exactly once after `intervalSeconds * 1000`ms, not before and not after. The existing "auto-navigates after interval when active and not paused" test covers this at 60s, but a dedicated assertion here would strengthen the regression test narrative — e.g., advance 59 seconds and assert no navigation, then advance 1 more and assert exactly one navigation, confirming the single-interval design drives both functions. Not blocking, just a suggestion to make the test more fully document the intent.

---

src/components/scheduler/use-screen-scheduler.ts
```diff
@@ -276,14 +276,41 @@ export function useScreenScheduler(
...
-  const controls: SchedulerControls = {
+  // Stable `controls` object identity is required so that consumers
+  // listing `controls` in a `useEffect` dependency array do not re-run
+  // their effects on every internal state change (e.g. the 1s countdown
+  // tick). The live implementations close over current state and are
+  // refreshed in a commit-phase effect; the exposed wrappers are
+  // created once via `useState`'s initializer and dispatch through the
+  // ref. This pattern is compatible with React Compiler — no
+  // `useMemo`/`useCallback` involved.
+  const liveControlsRef = useRef<SchedulerControls>({
```
BenSeymourODB, line 287: **[Low – Correctness, verified safe] Initial `liveControlsRef` value and first-render window**

The plan notes "wrappers called before the first effect commit see correct state — refs are stable so probably fine, but verify."

Verified: On the very first render, `liveControlsRef` is initialized with the render-time closures (`start`, `stop`, etc. defined just above). At that point no state has been set yet, so these are exactly the correct implementations for initial state — there is no stale-ref issue on first render.

However: the comment says the ref is "refreshed in a commit-phase effect" — this is accurate but could mislead a reader into thinking it's only refreshed once. The `useEffect` on line 295 has **no dependency array**, so it runs after **every** render, keeping `liveControlsRef.current` always up-to-date after commit. This is correct.

One nit: the `useEffect` on line 295 is missing its dependency array entirely. While "no deps = every render" is intentional here, React Compiler's exhaustive-deps ESLint rule will produce a lint warning (it expects either explicit deps or a suppression comment). Consider adding a comment:

```ts
useEffect(() => {
  liveControlsRef.current = { start, stop, pause, resume, navigateToNext, navigateToPrevious };
  // No dependency array: intentionally runs after every render to keep
  // liveControlsRef.current in sync with the latest render-time closures.
});
```

Or simply confirm this is not triggering a lint warning (the PR description says lint is clean, so this may already be handled by the config).

---

src/components/scheduler/__tests__/screen-scheduler.test.tsx
```diff
@@ -114,6 +114,78 @@ describe("ScreenScheduler", () => {
...
+        fireEvent.click(screen.getByTestId("content"));
+      });
+
+      expect(screen.getByRole("status")).toHaveAttribute(
+        "aria-label",
+        "Screen rotation paused"
+      );
+    });
+
+    it("a click on the pause/resume button does NOT re-pause the scheduler", () => {
```
BenSeymourODB, line 154: **[Low – Test Coverage] No test for ArrowLeft/ArrowRight not re-pausing when interaction-paused**

This test covers the toggle button correctly. But `ArrowLeft`/`ArrowRight` keyboard shortcuts (see `screen-scheduler.tsx:127-131`) do **not** call `resetInteractionPause()`. A user pressing ArrowRight while interaction-paused would find the keydown event itself re-triggers the interaction-pause (the event fires on `window`, which is outside `controlsContainerRef`, so the detector catches it). There is no integration test that catches this regression.

If the ArrowLeft/ArrowRight fix is applied (see comment on `screen-scheduler.tsx:134`), this test file is the right place for a matching test:

```ts
it("pressing ArrowRight while interaction-paused resumes rotation", () => {
  // 1. render + autoStart
  // 2. fireEvent.click(content) → interaction-paused
  // 3. fireEvent.keyDown(document, { key: 'ArrowRight' })
  // 4. expect status to show "next screen in..." not paused
});
```

---

BenSeymourODB (commented):
Deep review of PR #367 — scheduler stability and correctness bugs (#221).

**Overall:** The fixes for bugs 1 (stable controls reference) and 2 (manual-unpause lockout) are structurally correct and well-tested. The regression tests for bugs 4, 5, 6 are appropriately targeted. No `any` types, no manual memoization, no test artifact directories, CLAUDE.md conventions respected. Co-Authored-By lines present on all commits.

**One medium-severity bug found (not blocked, but should follow up):**

Arrow-key keyboard shortcuts (ArrowLeft/ArrowRight) don't call `resetInteractionPause()`, meaning a user who navigates by keyboard while interaction-paused will re-trigger the interaction-pause on every keypress (because the `keydown` event fires on `window`, outside the `controlsContainerRef` exemption zone). This is the same class of bug as the toggle-button lockout fixed in this PR, just not covered by the scope of bug 2's fix.

**Six inline comments posted** — see individual threads for details. Summary:
- 1× Medium (correctness): ArrowLeft/ArrowRight keyboard nav while interaction-paused re-triggers the interaction-pause — fix should mirror the Space-key treatment.
- 1× Medium (correctness/timing): `liveControlsRef` update via commit-phase `useEffect` vs render-phase write — safe today but fragile if future effects call controls synchronously during commit.
- 1× Low (complexity): `ignoreRefMirror` indirection is unnecessary since `controlsContainerRef` is always the same object; consider simplifying.
- 1× Low (test gap): `reset()` has no direct unit test in `use-interaction-detector.test.ts`.
- 1× Low (test fragility): Bug 6 `setIntervalSpy.mockRestore()` outside `finally` block — will leak spy on assertion failure.
- 1× Low (test coverage): No test for ArrowLeft/ArrowRight not re-pausing on keyboard navigation while interaction-paused.

---

src/components/scheduler/use-screen-scheduler.ts
BenSeymourODB: Acknowledged — leaving the commit-phase effect pattern in place for now (ESLint's "no ref writes during render" rule is on for this project, so the render-phase write would need a suppression). Documented the safety analysis in the new comment block on the effect itself in a follow-up commit. The window is empty for current call sites and the assumption is now explicit.

---

src/components/scheduler/use-interaction-detector.ts
BenSeymourODB: Good call — simplified in 0eb6257. Dropped the mirror, captured `ignoreRef` directly in the effect closure, and added `ignoreRef` to the dep array. For the current stable-`useRef` caller the effect still subscribes exactly once; if a caller ever passes an unstable ref the effect re-subscribes per render, which is loud and easy to spot. Comment on the read site now states the stable-RefObject assumption.

---

src/components/scheduler/screen-scheduler.tsx
BenSeymourODB: Fixed in 0eb6257 — both arrow-key cases now call `resetInteractionPause()`. Added a matching integration test in `screen-scheduler.test.tsx`: ArrowRight while interaction-paused resumes rotation. Good catch — this was the same class of bug as the toggle button, just on a different command path.

---

src/components/scheduler/use-interaction-detector.ts
BenSeymourODB: Added in 0eb6257. Four direct `reset()` tests: clears `isPaused` immediately; cancels the pending timeout (proved by a second pause window that the cancelled-then-fired old timeout cannot clear); no-op when not paused; identity stable across renders.

---

src/components/scheduler/__tests__/use-screen-scheduler.test.ts
BenSeymourODB: Fixed in 0eb6257 — wrapped the spy in try/finally so an assertion failure can't leak the spy into subsequent tests.

---

src/components/scheduler/__tests__/use-screen-scheduler.test.ts
BenSeymourODB: Added in 0eb6257 — the test now advances 60 seconds and asserts exactly one `router.push` to `/recipe`, alongside the structural one-1s-interval check. The single timer drives both behaviours; both are pinned.

---

src/components/scheduler/use-screen-scheduler.ts
BenSeymourODB: Thanks for the verification. The `useEffect` is in fact missing a deps array intentionally (commit-phase "run after every render"), and the project ESLint config does not flag that pattern — `pnpm lint` reports zero new warnings on this file. Confirmed.

---

src/components/scheduler/__tests__/screen-scheduler.test.tsx
BenSeymourODB: Added in 0eb6257 — `pressing ArrowRight while interaction-paused resumes rotation`, following the suggested shape (autostart → click content to interaction-pause → ArrowRight on window → assert status reads "next screen in" rather than "paused").

---

PR #369, opened 2026-06-03
chore(ci): add Playwright E2E coverage workflow (#281)

BenSeymourODB:
## Summary

- Adds `.github/workflows/e2e.yml` — Playwright E2E suite runs on PRs, on pushes to `main`, and via manual `workflow_dispatch`.
- Provisions a `postgres:17` service container, applies Prisma migrations with `pnpm db:migrate:deploy`, then runs `pnpm test:e2e --project=chromium`.
- Caches Playwright browsers keyed on the resolved `@playwright/test` version with a fast-path `install-deps` step on cache hit. Concurrency-cancels in-flight runs on the same ref.
- Uploads the Playwright HTML report and `test-results/` (traces, screenshots, videos) as artifacts on failure with 14-day retention.

## Plan

Linked plan: [`.claude/plans/playwright-e2e-ci-workflow.md`](./.claude/plans/playwright-e2e-ci-workflow.md)
Tracking: #281

## Scope choices

- **Chromium-only in CI** — the local `playwright.config.ts` matrix (firefox / webkit / mobile-chrome / tablet) stays intact for dev. Running all five projects with `workers: 1` blows past the 10-min runtime target from the issue. Cross-browser CI is a deliberate follow-up.
- **Auth-requiring specs auto-skip** — `calendar-integration.spec.ts` and friends call `getExistingUserSession()` and `test.skip(true)` when no session is in the DB, which is exactly what happens in CI. The shared signed-in fixture needed for non-trivial auth flows is tracked in #278; out of scope here.
- **No code changes** — pure CI configuration plus the plan file. No source files modified, so lint/format/types/Vitest remain unaffected.

## Test plan

- [x] `pnpm lint:fix` → 0 errors (6 pre-existing warnings unchanged)
- [x] `pnpm test:run` → 131 files / 2019 tests passing
- [ ] First workflow run on this PR completes within the 20-min job `timeout-minutes` (target ~10 min)
- [ ] Artifact upload paths verified on first failing run (will iterate if `playwright-report/` lands somewhere unexpected)

Closes #281

---
_Generated by [Claude Code](https://claude.ai/code/session_01Nu5h8iNT71rqQmUw1SbEEb)_


BenSeymourODB (commented):
First-pass review run via a Sonnet subagent. 5 findings, 4 actionable, 1 rejected. Summary so the trail is auditable:

**Applied:**

- **F2/F3 — Playwright install split.** Collapsed the cache-conditional `install --with-deps` / `install-deps` pair into one unconditional `pnpm exec playwright install --with-deps chromium` step. The download portion is idempotent on a cache hit; the `--with-deps` apt step is needed every run regardless because OS packages live outside `~/.cache/ms-playwright`. Closes a subtle failure mode where a cached browser binary on a fresh runner would have missing system libs.
- **F4 — `TOKEN_ENCRYPTION_KEY` visibility.** Added an inline comment under the `env:` block documenting that `src/lib/crypto/token-cipher.ts` derives a fallback key from `NEXTAUTH_SECRET` in non-production builds, so the var is intentionally absent for CI but must be set explicitly if a future workflow runs with `NODE_ENV=production`.
- **F5 — `--` arg separator.** Changed `pnpm test:e2e --project=chromium` to `pnpm test:e2e -- --project=chromium` so the flag is unambiguously forwarded to Playwright regardless of pnpm version.

**Rejected:**

- **F1 — claimed `actions/checkout@v6` / `setup-node@v6` / `upload-artifact@v5` don't exist.** They do; the existing `main_nextjs-template-build.yml` on `main` already uses `@v6` for checkout and setup-node and CI passes. The reviewer's training is stale here. No change.

Next push will include the F2-F5 fixes.

---

PR #376, opened 2026-06-03
feat(auth): automate OAuth token-key rotation with dual-read cipher (#215)

BenSeymourODB:
## Summary

PR #158 introduced the v1 token envelope but explicitly deferred the operational tooling for key rotation. Today rotating `TOKEN_ENCRYPTION_KEY` invalidates every stored envelope and forces every user to re-login on their next refresh — defensible for a single-family deploy but not a posture we want to ship to the self-host audience.

This change adds a **rolling-rotation pattern** that never breaks a user's session:

- **Dual-read in `src/lib/crypto/token-cipher.ts`** — a new `TOKEN_ENCRYPTION_KEY_PREVIOUS` env var holds the retired key during a rotation window. `decryptToken` tries the active key first, then falls back to previous on auth-tag mismatch. `encryptToken` always uses active.
- **`scripts/rotate-token-encryption-key.mjs`** — walks `Account` in batches, re-encrypts every encrypted token column under the active key, leaves legacy plaintext alone. `--dry-run` prints planned counts without writing; per-batch `prisma.$transaction` rolls back a batch on any in-batch failure; idempotent on partial reruns.
- **`docs/auth-token-encryption.md`** — replaces the "rotation = forced re-login" stub with a four-step rolling runbook (deploy both keys → run the CLI → confirm `failed=0` → drop the previous key).

Includes a separate prep commit (`fix(deps): repair broken pnpm-lock.yaml from #359 merge`) — `main`'s `pnpm-lock.yaml` had duplicate mapping keys (`tinyglobby@0.2.17`, `@microsoft/applicationinsights-cfgsync-js@3.3.11`) and a missing `ajv@6.15.0` entry left behind by the eslint-config-next bump in PR #359. Regenerating cleared all three; nothing else moved.

## Plan

Linked plan: [`.claude/plans/token-key-rotation-215.md`](./.claude/plans/token-key-rotation-215.md)

## Out of scope (deferred)

- A full live integration test against a seeded Postgres instance — the pure `reencryptAccountTokens` helper is unit-tested across null / legacy-plaintext / mixed / idempotent / tampered cases (6 tests), and the Prisma I/O wrapper in `main()` stays thin.
- Multiple previous keys — the cipher takes a single `TOKEN_ENCRYPTION_KEY_PREVIOUS`; expanding to a list is a small change later if needed.
- Envelope-version bump on rotation — the issue body suggested it but a pure key rotation under the same AES-256-GCM algorithm doesn't need it; the version byte stays reserved for algorithm changes.

## Test plan

- [x] `pnpm test` — 2231 tests pass (32 cipher tests including 7 new dual-read cases; 6 new CLI helper tests)
- [ ] CI — green
- [ ] Manual smoke against a dev DB with a real rotation cycle (deferred to follow-up; the unit harness exercises the same `reencryptAccountTokens` path the CLI uses)

Closes #215

---
_Generated by [Claude Code](https://claude.ai/code/session_01BzD5orBYarvhuJfXXMh1Xr)_

### DIFF
--- a/.claude/plans/token-key-rotation-215.md
+++ b/.claude/plans/token-key-rotation-215.md
@@ -1,0 +1,182 @@
+# Plan — feat(auth): automate OAuth token-key rotation (#215)
+
+## Goal
+
+PR #158 ("encrypt OAuth tokens at rest") introduced a v1 envelope but
+explicitly deferred operational tooling for key rotation. The
+`docs/auth-token-encryption.md` runbook today tells operators that a
+rotation forces every user to re-login — fine for a single-family
+deployment but not a defensible posture for the broader self-host
+audience.
+
+This issue adds:
+
+1. **Dual-read decryption** so an active key and a previous key are
+   both accepted for reads during a rotation window. Encryption always
+   uses the active key.
+2. **A rotation CLI** (`scripts/rotate-token-encryption-key.mjs`) that
+   walks the `Account` table in batches, decrypts each token with the
+   active-or-previous key, and re-encrypts it under the active key.
+3. **Documentation** in `docs/auth-token-encryption.md` covering the
+   rolling-rotation runbook.
+
+## Out of scope (deferred)
+
+- A full live integration test against a seeded Postgres instance. The
+  CLI's rotation logic will be unit-tested by extracting the pure
+  re-encrypt step into a testable function and feeding it fake rows;
+  the Prisma I/O wiring stays thin.
+- Multiple previous keys (a single `TOKEN_ENCRYPTION_KEY_PREVIOUS` is
+  enough for the one-at-a-time rotation pattern; expanding to a list
+  is trivial later if needed).
+- A new envelope version. The issue body suggested "bumping the
+  version byte" on rotation, but a pure key rotation under the same
+  AES-256-GCM algorithm does not need it — the version byte stays
+  reserved for algorithm/scheme changes, which keeps the rotation
+  scriptidempotent (running it twice produces identical envelope
+  shapes).
+
+## Design
+
+### Dual-read in `src/lib/crypto/token-cipher.ts`
+
+Today `resolveKey()` returns a single `Buffer`. After this change it
+becomes `resolveKeys()` and returns:
+
+```ts
+type ResolvedKeys = {
+  active: Buffer;
+  previous: Buffer | null;
+};
+```
+
+`encryptToken` reads `active` only. `decryptToken`:
+
+1. Attempts AES-GCM decrypt with `active`.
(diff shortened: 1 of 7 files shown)


BenSeymourODB (commented):
## First-pass review summary

A Sonnet review agent went over the diff with focus on the cipher's dual-read correctness, the CLI's transaction/cursor semantics, and CLAUDE.md compliance. Findings + actions taken in commit `445308b`:

### Addressed in code

- **[blocker] CLI batch transaction flow** — Wrap `prisma.$transaction` in explicit try/catch so `rotated += updates.length` and the page cursor only advance on commit. On rollback we now log, bump `failed`, and break out of the loop (the final summary stays accurate; the operator sees the partial state without scanning stderr backwards). _Note: in the original code, the bare `await` already prevented counter inflation — but the new wrapper makes the invariant explicit and future-proofs against a refactor that forgot it._
- **[question] Cross-key GCM false-positive concern** — Added a one-line comment to the decrypt fallback noting that GCM's 128-bit MAC makes a cross-key false-positive negligible (2⁻¹²⁸). Documents why try-active-then-previous is safe and not a security smell.
- **[nit] Docs runbook overstates idempotency** — Reworded step 3 to make it explicit that a rerun always restarts from row 0 (the page cursor is in-process only); idempotency means "running again is safe", not "resumable from a checkpoint".
- **[question] Test comments** — Added a note on the module-reset ordering in the rotation-window test (why `loadHelper()` is called after `loadCipher()`) and on the idempotency test's narrow scope (active-key path only; the dual-read fallback is exercised by the cipher suite + the earlier rotation test).

### Declined, with rationale

- **[nit] Hash the cache-key source string in `token-cipher.ts`** — The decoded `Buffer` is already in memory for the lifetime of the cached entry, and the base64-encoded source string is the same key material in a different shape. Hashing only avoids a second in-memory copy, not exposure (the Buffer is the bigger target). Skipping this; happy to revisit if the cipher ever moves to a longer-lived service.
- **[nit] `printUsageAndExit` not exported / not unit-tested** — Leaving it private. The `--help` branch is a one-line `process.exit(0)` after writing a static string; pulling it into a test would mean mocking `process.exit` and duplicating the usage string. Net cost > value.
- **[general] AC item 2 says "keyed on the envelope version byte"** — Deliberately diverged. Pure key rotation under the same AES-256-GCM algorithm doesn't need a version bump (the version byte stays reserved for algorithm changes); the env-var-based fallback gives the same operational behaviour with a smaller surface. Called out in the PR description already.

Validation after the review pass: `pnpm lint:fix && pnpm format:fix && pnpm check-types` clean, `pnpm test` for the cipher + CLI suites all green (38/38).

---

Issue #337, opened 2026-05-06
feat(settings): unify timeFormat — single source of truth in UserSettings (retire CalendarProvider.use24HourFormat duplicate)


BenSeymourODB:
## Context

We have two independent stores for the same setting:

- `UserSettings.timeFormat` (`prisma/schema.prisma:82`) — string `"12h"` / `"24h"`, edited from `DisplaySection` in the main Settings page (`src/components/settings/display-section.tsx:60`)
- `CalendarProvider.use24HourFormat` (`src/components/providers/CalendarProvider.tsx:74`) — boolean, persisted to its own `localStorage` key, exposed via `CalendarSettingsPanel` (`src/components/calendar/CalendarSettingsPanel.tsx:91`)

The two are unsynchronised: a user toggling time format inside `CalendarSettingsPanel` does not affect the value in `UserSettings`, and vice versa. Other surfaces (Tasks due-time, Meal slots) have to pick one or the other.

## Scope

- `UserSettings.timeFormat` becomes the single source of truth
- `CalendarProvider` reads `timeFormat` from `useUserSettings()` and exposes a derived `use24HourFormat: boolean = timeFormat === "24h"` so callers don't break
- `CalendarSettingsPanel`'s 24-hour switch PUTs to `/api/settings` instead of writing to `localStorage`
- Remove the `localStorage` key for `use24HourFormat` (with a one-time migration that reads any existing value into `UserSettings`)
- `DisplaySection` continues to be the canonical edit surface; the calendar panel is a convenience

## Acceptance criteria

- [ ] Toggling time format in either surface immediately updates the other on the same tab
- [ ] After reload, the value persists from the DB
- [ ] One-time migration: existing localStorage value populates `UserSettings.timeFormat` if the user hasn't set it explicitly yet
- [ ] No reads of the removed `localStorage` key remain
- [ ] Unit + Playwright coverage for the cross-surface sync

## Dependencies

- None

## Tracking

#328 — App-wide settings consolidation (S1).

PR #377, opened 2026-06-03
feat(cache): LRU eviction on QuotaExceededError (#290 sub-task 1)

BenSeymourODB:
## Summary

- When `EventCache.saveEvents` trips a `QuotaExceededError`, retry the write after evicting least-recently-cached rows in bounded batches instead of immediately falling back to the in-memory store.
- Add `EventStore.evictOldest(count)` (implemented for both `InMemoryEventStore` and `IndexedDBEventStore`); the IDB path reuses the existing `cachedAt` index, no schema bump.
- Successful eviction-then-retry surfaces a structured `logger.event("event-cache.lru-evicted", …)` carrying the evicted-row count and attempt number.

## Slice

This PR implements **only sub-task 1** of issue #290. The other two sub-tasks are deferred to follow-ups under the same issue:

- [ ] Background `visibilitychange` cleanup sweep (next PR).
- [ ] TTL wired to `UserSettings` — waits on PR #344 (issue #337) to land so it can reuse the `useUserSettings.mutate` + `user-settings-bus` cross-surface sync pattern instead of re-inventing it.

The issue body explicitly says "each can ship as its own PR", so `#290` intentionally stays open after this merges — the remaining sub-tasks track the rest. **No auto-close** keyword in this body.

## Plan

Linked plan: [`.claude/plans/event-cache-lru-eviction-290.md`](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/claude/ecstatic-mayer-XDMBH/.claude/plans/event-cache-lru-eviction-290.md)

### Design notes

- **Retry loop lives on the facade**, not the store. `EventStore` already exposes `put` + `evictOldest`; orchestrating the loop in `EventCache.saveEvents` keeps `IndexedDBEventStore` simple and lets the existing mock-`EventStore` test pattern inject `QuotaExceededError` deterministically without needing `fake-indexeddb`.
- **Recency = `cachedAt`**. The issue body suggests "likely `lastReadAt`"; I'm using the existing `cachedAt` because today's access pattern is bulk `getAll` (no per-row reads), so a `lastReadAt` field would either be a no-op (every row touched, all values converge) or require expensive per-row writes on every read. `cachedAt` is already indexed → no schema bump. Documented in the plan.
- **Termination**: bounded by `MAX_EVICTION_ATTEMPTS = 5`, empty-batch break (`evictOldest` returning 0 short-circuits), and non-quota errors propagate immediately to the existing fallback path. After exhausting the budget, the original `QuotaExceededError` is re-thrown so `withFallback` swaps in the in-memory store as before.

## Test plan

- [x] Unit tests cover `InMemoryEventStore.evictOldest` (full, partial, zero, empty-store).
- [x] Unit tests cover `EventCache` orchestration through a mock `EventStore`: retry-after-evict succeeds; structured `logger.event` fires; the bounded retry loop terminates and falls back when quota is unrecoverable; the loop breaks on empty `evictOldest`; non-quota errors skip the eviction path.
- [x] Pre-existing tests (in-memory fallback path, TTL filter, upsert semantics) all still pass.
- [x] `pnpm test` — 2120/2120 green.

Tracking: #290 (sub-task 1 of 3 — issue intentionally stays open).

### DIFF
--- a/src/lib/__tests__/event-cache.test.ts
+++ b/src/lib/__tests__/event-cache.test.ts
@@ -7,7 +7,29 @@ import {
   selectEventStore,
 } from "@/lib/event-cache";
 import type { GoogleCalendarEvent } from "@/lib/google-calendar";
+import { logger } from "@/lib/logger";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@/lib/logger", () => ({
+  logger: {
+    error: vi.fn(),
+    log: vi.fn(),
+    event: vi.fn(),
+    dependency: vi.fn(),
+  },
+}));
+
+/**
+ * Mimic the `DOMException` shape that IndexedDB throws when a write exceeds
+ * the storage quota. `name === "QuotaExceededError"` is the contract; in
+ * Chromium-family browsers this is a `DOMException` instance, but tests
+ * just need an `Error`-shaped object with the right `name`.
+ */
+function quotaExceededError(message = "Quota exceeded"): Error {
+  const err = new Error(message);
+  err.name = "QuotaExceededError";
+  return err;
+}
 
 function makeEvent(
   id: string,
@@ -91,6 +113,46 @@ describe("InMemoryEventStore", () => {
   it("returns an empty array when nothing has been written", async () => {
     expect(await store.getAll(1_000)).toEqual([]);
   });
+
+  describe("evictOldest (#290 — LRU support)", () => {
+    it("removes the N rows with the smallest cachedAt", async () => {
+      await store.put([makeEvent("old1")], 1_000);
+      await store.put([makeEvent("old2")], 2_000);
+      await store.put([makeEvent("recent1")], 3_000);
+      await store.put([makeEvent("recent2")], 4_000);
+
+      const evicted = await store.evictOldest(2);
+
+      expect(evicted).toBe(2);
+      // Use a `now` close to the most recent write so the TTL filter does
+      // not also remove rows; this isolates the LRU eviction effect.
+      const remaining = await store.getAll(4_000);
+      expect(remaining.map((e) => e.id).sort()).toEqual(["recent1", "recent2"]);
+    });
+
+    it("returns the actual evicted count when fewer than N rows exist", async () => {
+      await store.put([makeEvent("a"), makeEvent("b")], 1_000);
+
+      const evicted = await store.evictOldest(10);
+
+      expect(evicted).toBe(2);
+      expect(await store.getAll(1_000)).toEqual([]);
+    });
+
+    it("is a no-op when asked to evict zero rows", async () => {
+      await store.put([makeEvent("a"), makeEvent("b")], 1_000);
+
+      const evicted = await store.evictOldest(0);
+
+      expect(evicted).toBe(0);
+      const all = await store.getAll(1_000);
+      expect(all.map((e) => e.id).sort()).toEqual(["a", "b"]);
+    });
+
+    it("returns 0 when the store is already empty", async () => {
+      expect(await store.evictOldest(5)).toBe(0);
+    });
+  });
 });
 
 describe("EventCache facade", () => {
@@ -146,6 +208,7 @@ describe("EventCache facade", () => {
       }),
       getAll: vi.fn(async () => []),
       clear: vi.fn(async () => {}),
+      evictOldest: vi.fn(async () => 0),
     };
     const flakyCache = new EventCache(flaky);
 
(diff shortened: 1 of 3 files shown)


src/lib/event-cache.ts
```diff
@@ -175,6 +209,34 @@ export class IndexedDBEventStore implements EventStore {
...
+    return new Promise((resolve, reject) => {
+      const transaction = this.db!.transaction([STORE_NAME], "readwrite");
+      const objectStore = transaction.objectStore(STORE_NAME);
+      const index = objectStore.index("cachedAt");
+      // Ascending cursor over `cachedAt` — first key is the oldest row.
+      const cursorRequest = index.openCursor(null, "next");
+      let evicted = 0;
+      cursorRequest.onsuccess = () => {
+        const cursor = cursorRequest.result;
+        if (!cursor || evicted >= count) {
```
BenSeymourODB: **[blocker] IDB cursor: `resolve(evicted)` fires before all deletes have committed — confident**

The early-exit branch (`evicted >= count`) calls `resolve(evicted)` from inside the `onsuccess` callback _before_ the transaction's `oncomplete` event fires. IndexedDB `cursor.delete()` is asynchronous; it enqueues the delete in the transaction but does not flush it to disk until `oncomplete`. Resolving the promise here means the caller's `await store.evictOldest(...)` returns while the transaction is still open and the rows may not yet be gone. If `putWithLruRetry` then immediately issues `store.put(events, cachedAt)` — which opens a new transaction — there is a window where the old rows are still present and the quota is not yet freed.

The `!cursor` branch (natural end-of-range) has the same race for the same reason.

**Fix:** wire `resolve` to `transaction.oncomplete` instead of to the cursor callback.

```ts
return new Promise((resolve, reject) => {
  const transaction = this.db!.transaction([STORE_NAME], "readwrite");
  const objectStore = transaction.objectStore(STORE_NAME);
  const index = objectStore.index("cachedAt");
  const cursorRequest = index.openCursor(null, "next");
  let evicted = 0;
  cursorRequest.onsuccess = () => {
    const cursor = cursorRequest.result;
    if (!cursor || evicted >= count) return; // let transaction complete naturally
    cursor.delete();
    evicted += 1;
    cursor.continue();
  };
  cursorRequest.onerror = () => reject(cursorRequest.error);
  transaction.oncomplete = () => resolve(evicted);
  transaction.onerror = () => reject(transaction.error);
});
```

Note: stopping cursor iteration by simply not calling `cursor.continue()` is sufficient — the transaction will complete on its own once there are no pending requests. The early `resolve` in the current code is the only bug; `cursor.delete()` ordering itself is fine.

---

src/lib/event-cache.ts
```diff
@@ -201,11 +263,58 @@ export class EventCache {
...
+          // Nothing left to evict — propagate so the in-memory fallback
+          // can take over. The thrown error is the original quota error,
+          // preserved so logs upstream stay accurate.
+          throw error;
+        }
+        try {
+          await store.put(events, cachedAt);
+          logger.event(
+            "event-cache.lru-evicted",
+            { context: "EventCache.saveEvents" },
```
BenSeymourODB, line 310: **[blocker] `logger.event` properties type violation — confident**

`logger.event(name, properties, measurements)` requires `properties: Record<string, string | number | boolean>` (see `src/lib/logger.ts:103`, `src/lib/logger.ts:454`). The call passes `{ context: "EventCache.saveEvents" }` which satisfies that type at runtime, but the _measurements_ object `{ evicted: totalEvicted, attempts: attempt }` is `Record<string, number>` — that part is fine.

The real problem is subtler: the plan draft (`.claude/plans/event-cache-lru-eviction-290.md`) listed `context` as one of the properties, and the code correctly puts it there. However, `context` is a string, so the types _do_ match. TypeScript will accept this.

**Actual issue:** the plan's acceptance-criteria comment mentions `retried` as a third property (`logger.event("event-cache.lru-evicted", { evicted, attempts, retried })`), but the implementation omits `retried` entirely and moves `evicted`/`attempts` to the measurements bucket instead. That's actually _better_ (numeric data belongs in measurements, not properties), but the test at `event-cache.test.ts:262` asserts on `context` only and never asserts that `evicted` and `attempts` are in the _measurements_ position (third arg) rather than properties (second arg). If a future refactor accidentally swaps them, the test won't catch it.

**Fix:** tighten the test assertion to match the exact call signature:
```ts
expect(logger.event).toHaveBeenCalledWith(
  "event-cache.lru-evicted",
  { context: "EventCache.saveEvents" },          // properties — strings only
  { evicted: expect.any(Number), attempts: expect.any(Number) } // measurements
);
```
(The current test already uses `expect.objectContaining` for both args so the positions _are_ asserted, just loosely. This is a nit on the test, not the implementation.)

---

src/lib/__tests__/event-cache.test.ts
```diff
@@ -159,13 +222,150 @@ describe("EventCache facade", () => {
...
+
+      const c = new EventCache(store);
+      // Should not throw — falls back to in-memory after the bounded loop.
+      await c.saveEvents([makeEvent("a")]);
+
+      // The retry loop attempted at most a fixed bound (the constant is
+      // intentionally not pinned to a specific number in this assertion —
+      // we only care that it terminated and fell back).
+      expect(
+        (store.put as ReturnType<typeof vi.fn>).mock.calls.length
```
BenSeymourODB, line 308: **[blocker] Missing test: non-quota error thrown on retry (second put) — confident**

All five LRU tests inject the quota error on `put` call #1. None of them cover the case where the _first_ put throws `QuotaExceededError`, eviction succeeds, and then the _retry_ put throws a completely different error (e.g. `AbortError`, a generic `DOMException`, a network-level failure). In `putWithLruRetry` that path hits `catch (retryError) { if (!isQuotaExceededError(retryError)) throw retryError; }` — the non-quota error is rethrown, which is correct — but it is completely untested.

**Why this matters:** `withFallback` wraps the entire `putWithLruRetry` call and catches all errors to trigger the in-memory fallback. If a non-quota retry error is thrown, the fallback _should_ engage. Without a test, a future refactor of `putWithLruRetry` that accidentally swallows that branch (e.g. a wrong catch scope) will go undetected.

**Suggested test:**
```ts
it("propagates a non-quota error thrown on the retry put", async () => {
  let putCalls = 0;
  const nonQuotaError = new Error("AbortError");
  nonQuotaError.name = "AbortError";
  const store: EventStore = {
    put: vi.fn(async () => {
      putCalls += 1;
      if (putCalls === 1) throw quotaExceededError();
      throw nonQuotaError; // second attempt fails differently
    }),
    getAll: vi.fn(async () => []),
    clear: vi.fn(async () => {}),
    evictOldest: vi.fn(async (count: number) => count),
  };

  const c = new EventCache(store);
  // withFallback catches the rethrown AbortError and engages the in-memory fallback
  await c.saveEvents([makeEvent("a")]);
  expect((c as unknown as { hasFallenBack: boolean }).hasFallenBack).toBe(true);
  // evictOldest was called exactly once before the non-quota error terminated the loop
  expect(store.evictOldest).toHaveBeenCalledTimes(1);
});
```

---

src/lib/event-cache.ts
```diff
@@ -201,11 +263,58 @@ export class EventCache {
...
+        const evicted = await store.evictOldest(EVICTION_BATCH_SIZE);
+        totalEvicted += evicted;
+        if (evicted === 0) {
+          // Nothing left to evict — propagate so the in-memory fallback
+          // can take over. The thrown error is the original quota error,
+          // preserved so logs upstream stay accurate.
+          throw error;
+        }
+        try {
+          await store.put(events, cachedAt);
```
BenSeymourODB: **[nit] `cachedAt` captured before the first put may be stale across multiple retry attempts — uncertain**

`const cachedAt = Date.now()` is captured once before the first `store.put` call. If the eviction loop runs `MAX_EVICTION_ATTEMPTS` (5) times with non-trivial `evictOldest` latency (each IDB cursor walk), the final successful `store.put(events, cachedAt)` writes a timestamp that could be several hundred milliseconds in the past. In practice this is unlikely to matter (the TTL is 24 h), but it's worth a comment explaining the intentional choice, especially since the plan doesn't mention it.

If you do want a fresh timestamp per retry, moving `const cachedAt = Date.now()` to just before the inner `store.put(events, cachedAt)` call on line 300 would fix it with no other changes.

---

src/lib/event-cache.ts
```diff
@@ -64,6 +86,18 @@ export class InMemoryEventStore implements EventStore {
   async clear(): Promise<void> {
     this.rows.clear();
   }
+
+  async evictOldest(count: number): Promise<number> {
```
BenSeymourODB, line 90: **[praise] `InMemoryEventStore.evictOldest` — empty-map and partial-eviction cases handled cleanly**

The guard `count <= 0 || this.rows.size === 0` short-circuits both the zero-count and empty-map cases correctly. `sorted.slice(0, count)` when `count > this.rows.size` returns all entries (JavaScript `slice` clamps silently), so `toEvict.length` always reflects the actual count evicted — no off-by-one risk. The sort is on `cachedAt` numeric difference, which is a valid comparator. Stable behaviour on identical `cachedAt` values isn't guaranteed by the JS spec for `Array.sort` in all engines, but given that `cachedAt` is set per `put` call (not per event), ties can only occur across events written in the same batch, and evicting any subset of those is equally correct.

---

BenSeymourODB (commented):
Independent review of the LRU eviction PR (#290 sub-task 1). Three blockers found, one nit, one praise. The design and retry-loop structure are sound; the issues are in the IDB cursor completion race and a missing test path.

**Conflict surface with PR #344:** None. The change is entirely in `src/lib/event-cache.ts` / its test file. No overlap with `useUserSettings.mutate` or `user-settings-bus`.

**CLAUDE.md compliance:** Clean. No `useMemo`/`useCallback`/`React.memo`, no `any`, no Tailwind (pure-TS lib). `pnpm lint:fix && pnpm format:fix && pnpm check-types` should still need to be run before merge — not verified locally.

---

src/lib/event-cache.ts
BenSeymourODB: Fixed in 4ceaae8. `resolve(evicted)` now lives on `transaction.oncomplete`; the cursor callback only signals end-of-iteration by not calling `cursor.continue()`. The transaction commits naturally once no more requests are pending.

---

src/lib/event-cache.ts
BenSeymourODB: Tightened in 4ceaae8. Both args now use exact object matching — `objectContaining` is gone. A future swap of properties↔measurements will now fail this test.

---

src/lib/__tests__/event-cache.test.ts
BenSeymourODB: Added in 4ceaae8 — "propagates a non-quota error thrown on the retry put (after eviction)". First put throws `QuotaExceededError`, eviction succeeds (returns `count`), second put throws an `AbortError`; the test asserts `withFallback` engages (`hasFallenBack === true`) and that exactly one eviction + two puts ran before the loop terminated.

---

src/lib/event-cache.ts
BenSeymourODB: Fixed in 4ceaae8. `Date.now()` is now re-sampled inline on every `store.put` call (both the initial attempt and each retry). Comment on the helper documents the choice.

---

src/lib/event-cache.ts
BenSeymourODB: Acknowledged — the tie-on-`cachedAt` observation is worth keeping in mind if we ever start writing events one at a time instead of by batch. For now, batch-level ties are equivalent for eviction purposes (any subset evicted is correct).

---

PR #380, opened 2026-06-05
feat(auth): validate GOOGLE_CLIENT_ID / GOOGLE_CLIENT_SECRET at boot (#379)

BenSeymourODB:
## Summary

- New `validateGoogleOAuthEnv()` helper that throws on a missing or empty `GOOGLE_CLIENT_ID` / `GOOGLE_CLIENT_SECRET` at server start, parallel to the existing `validateEncryptionKey()` from #315 / PR #348.
- Wired into `src/lib/auth/auth.ts` module init under the same `NEXT_PHASE !== "phase-production-build"` guard so `next build` page-data collection still works without runtime env vars.
- Closes the last branch of the misconfig-as-mid-session-error failure class: with both checks in place, an operator restarting the server with broken env will now learn about it immediately instead of via user reports of recurring "Session expired" prompts.

## Plan

Linked plan: `.claude/plans/validate-google-oauth-env-379.md`

## Test plan

- [x] 7 new unit tests in `src/lib/auth/__tests__/validate-google-oauth-env.test.ts` covering:
  - Both env vars populated → no throw
  - Each var missing in isolation → throws and names the missing var
  - Both missing → single error mentions both var names (one restart fixes both)
  - Empty-string footgun (`export X=`) treated the same as unset
  - Error message references issues #315 / #379 for future triage
- [x] Full suite: `pnpm test` → 2337 tests pass across 147 files.
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` clean (only pre-existing warnings remain).
- [ ] (Out of CI) `pnpm build` still succeeds — guarded by `NEXT_PHASE` so page-data collection isn't affected.

Closes #379

---
_Generated by [Claude Code](https://claude.ai/code/session_01TMxPyhdmKPRHB1sbA2xLBa)_


BenSeymourODB (commented):
First-pass review (Sonnet subagent). Two findings worth addressing — see inline comments. Plan-vs-code matches; CLAUDE.md compliance clean.

---

BenSeymourODB (commented):
## In-depth review — `validateGoogleOAuthEnv` boot assertion

Independent review pass via `/implement-issue`. Merged current `main` into the branch (`e95a2b3`, clean), re-verified the full quality gate, then ran a Sonnet sub-reviewer against the validator, the test file, the `auth.ts` integration, and the parallel `validateEncryptionKey()` implementation.

### Verdict — **READY_WITH_NITS** → **MERGE-READY** (both nits applied in `b77751a`)

Two test-file nits were actionable, both small and orthogonal to the implementation; I applied them on this branch rather than blocking the PR on a follow-up. Two observations are noted below for the record (zero action required).

### Local verification after the rebase + cleanup

| Check | Result |
|---|---|
| `pnpm install --frozen-lockfile` | clean — Prisma client generated via `postinstall` |
| `pnpm lint` | 0 errors, 6 pre-existing warnings (all in unrelated files — `MockCalendarProvider`, `account-section`, `profile-avatar`, etc.) |
| `pnpm format:fix` | clean (no changes) |
| `pnpm test:run` (full) | **154 files / 2483 tests passing** |
| `pnpm test:run` on `validate-google-oauth-env.test.ts` | **9 / 9 passing** post-cleanup |
| `mergeable_state` (pre-push) | `clean` |

### What's good

- **Whitespace handling** — the validator uses `!clientId?.trim()` to catch unset, empty-string, **and** whitespace-only credentials. The `.trim()` is a deliberate improvement on the plan spec and is covered by dedicated tests (the `"   "` and `"\t\n  "` cases). A whitespace-only `export X=" "` was the silent footgun that would otherwise pass the validator and fail later as a Google `invalid_client`.
- **Operator-friendly error message** — names *both* missing variables in a single throw (so one restart fixes both), links to the Google Cloud Console URL, names the downstream failure mode (`invalid_client` → terminal → forced re-auth), and cites the issue trail (#315 / #379) for the next triager. No secret value is ever included in the message — only variable *names*.
- **`NEXT_PHASE` guard parity** — the placement and semantic are byte-for-byte identical to `validateEncryptionKey()`'s. Both validators are properly skipped during `next build` page-data collection and both run during `pnpm dev` and on production server start.
- **Doc-comment rewrite in `auth.ts`** — the old comment cited only #315; the new one accurately covers both validators and both misconfig classes (encryption-key gap → `RefreshTokenError`; OAuth-creds gap → `invalid_client`). No misleading language.
- **Naming** — `validateGoogleOAuthEnv` mirrors `validateEncryptionKey`. No `any`, strict TS, no banned React hooks (server-only code). CLAUDE.md compliance is clean across all axes.
- **Scope discipline** — `NEXT_PUBLIC_GOOGLE_CLIENT_ID` (the legacy browser-side var still read by `src/lib/google-calendar.ts`) is explicitly out-of-scope per the plan, which is the right call — that's a separate cleanup.

### Findings applied in `b77751a`

**🟢 NIT 1 — Test file used a manual `delete` + snapshot/restore pattern with an outdated comment** → fixed.

The original test file said `vi.stubEnv` "only writes strings, so the truly-absent case (`undefined`) has to be tested with a direct `delete process.env.X`." That's accurate for older Vitest builds but the installed version handles `vi.stubEnv(name, undefined)` by deleting the key (the sibling `src/app/api/auth/refresh-token/__tests__/route.test.ts` already uses this convention at lines 88 and 107). The `afterEach` was also doing double work — `vi.unstubAllEnvs()` already restores every key registered via `vi.stubEnv`, so the manual `ORIGINAL_CLIENT_ID === undefined ? delete : assign` block only had value for the raw-delete tests, which no longer exist post-refactor. Switched to `vi.stubEnv(name, undefined)` uniformly and dropped the manual snapshot/restore block. -39/+13 lines, no behaviour change, all 9 tests still pass.

**🟢 NIT 2 — Two tests used manual `try/catch/let caught: unknown` where `expect(() => ...).toThrow(pattern)` would suffice** → fixed.

The "both missing" test (original lines 73–88) and the "issue trail" test (original lines 90–102) used a manual try/catch pattern inconsistent with the other 7 tests in the file. The "both missing" case asserts both var names against the same call — implemented as two `.toThrow` assertions:

```ts
vi.stubEnv("GOOGLE_CLIENT_ID", undefined);
vi.stubEnv("GOOGLE_CLIENT_SECRET", undefined);
expect(() => validateGoogleOAuthEnv()).toThrow(/GOOGLE_CLIENT_ID/);
expect(() => validateGoogleOAuthEnv()).toThrow(/GOOGLE_CLIENT_SECRET/);
```

The issue-trail test collapses to a single `.toThrow(/#315|#379/)`.

### Observations (no action needed, recorded for the next reader)

**🔵 OBSERVATION 3 — NUL-byte-only credentials would pass `.trim()`.**

`"\0".trim()` returns `"\0"` (NUL is not whitespace per Unicode), so `!"\0".trim()` is `false` and a NUL-only credential would silently pass the check. Real-world probability is zero — no shell or secret manager produces a NUL-byte env var — but it's a logical gap versus the whitespace intent. Not worth a code change; flagging only because a future hardening pass might want to add a positive `/^[A-Za-z0-9._\-]+$/` shape check for OAuth-client-ID-looking values.

**🔵 OBSERVATION 4 — The `NEXT_PHASE` guard never trips in the test suite.**

During `pnpm test` (Vitest), `NEXT_PHASE` is `undefined` so the guard *would* run — but because no test file imports `auth.ts` directly (the plan explicitly avoids this on line 119), the module-init validators never fire in unit tests. This is the intended architecture and matches `validateEncryptionKey()`'s test posture; no `NODE_ENV === "test"` escape hatch is needed. Just making it explicit.

### Recommendation

**Ready to merge once CI confirms green on `b77751a`.** The change does exactly what it says: closes the last branch of the misconfig-as-mid-session-error failure class from #315, in a small, focused, well-tested PR that mirrors the existing pattern.

---
_Reviewed by `/implement-issue` PR-review run via Claude Code — independent Sonnet sub-review, findings applied directly. Branch is up-to-date with `main` after the merge in `e95a2b3`._

---

PR #381, opened 2026-06-05
refactor(settings): DisplaySection radios → shadcn RadioGroup (#373)

BenSeymourODB:
## Summary

Closes #373.

Replaces the three native `<input type="radio">` fieldsets (Theme, Time Format, Week starts on) in `src/components/settings/display-section.tsx` with shadcn `RadioGroup` / `RadioGroupItem` — the same primitive `CalendarSettingsPanel` already uses for its radio controls.

- Drops the residual no-op `text-blue-600` (and `accent-blue-600` on Week-starts-on) classes — Radix radio styling now flows through the shadcn primitive's `text-primary` / `border-input` tokens, all on the standard Tailwind palette per CLAUDE.md.
- Preserves the existing `onChange({ … })` contract end-to-end: theme is `string`, timeFormat is `string`, weekStartDay is `Number(value) as TWeekStartDay`. The Theme-section side effect (`setTheme(...)` from `next-themes`) is unchanged.
- The `setTheme` call routes through `handleThemeChange`, which is now the `RadioGroup.onValueChange` handler — semantically the same flow as before.

## Why

Filed from PR #354's review thread ([#discussion_r3204811542](https://github.com/BenSeymourODB/next-digital-wall-calendar/pull/354#discussion_r3204811542)). Theme + Time Format still carried the stale `text-blue-600` class (no visual effect on a native radio — accent comes from `accent-color`, not `color`). #354 fixed the new Week-starts-on radios to `accent-blue-600` as an in-scope correction; this PR finishes the job by moving all three to the shadcn primitive and dropping the colour classes entirely. Consistent radio styling + built-in keyboard/a11y semantics across Settings UI.

## Test plan

- [x] All 13 existing `display-section.test.tsx` tests still pass — `getByLabelText` and `toBeChecked()` work against Radix `role="radio"` + `aria-checked` semantics.
- [x] Full suite (`pnpm test`): 2330 / 2330 passing across 146 files.
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types`: clean (0 errors; pre-existing warnings unrelated to this change).

## Plan

No `.claude/plans/` file — this is a small, well-scoped refactor with crisp acceptance criteria in the issue body.

Project: https://github.com/users/BenSeymourODB/projects/1

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_011yck24117y7RAxcf1oQxVX)_

PR #390, opened 2026-06-08
feat(settings): formatUserDate helper + wire dateFormat to AccountSection (#339)

BenSeymourODB:
## Summary

- Introduces `src/lib/format-date.ts` (`formatUserDate` + `TDateFormat` + `isDateFormat`) — the centralised mapper for `UserSettings.dateFormat`.
- Wires `dateFormat` through `useUserSettings` (default, `pickCalendarFields` validation, bus propagation) so the value rides the in-tab user-settings bus introduced by #344 alongside `timeFormat`.
- Threads the user's preference into the `AccountSection` "Member since" line — previously the only `toLocaleDateString` consumer in user-facing production code.
- Adds a `/test/date-format-sync` page + Playwright spec mirroring `/test/time-format-sync` (#337) to prove cross-surface bus propagation end-to-end.

The setting was previously persisted (Prisma + `/api/settings`) but no client code read it as a formatting directive — exactly the "dead setting" #339 calls out.

## Plan

Linked plan: `.claude/plans/date-format-wireup-339.md`

## Test plan

- [x] `pnpm test:run src/lib/__tests__/format-date.test.ts` — 9 tests cover every supported format, ISO-string input, default fallback, and invalid-date safety.
- [x] `pnpm test:run src/hooks/__tests__/useUserSettings.test.tsx` — 33 tests; 3 new for `dateFormat` (server pickup, garbage rejection, bus propagation).
- [x] `pnpm test:run src/components/settings` — 93 tests; 3 new for `AccountSection` rendering each supported format.
- [x] `pnpm test:run` whole suite — 2,345 tests pass.
- [ ] `pnpm test:e2e e2e/date-format-sync.spec.ts` — not exercised locally (Playwright browser download blocked in this sandbox); spec mirrors `e2e/time-format-sync.spec.ts` and will run in CI under PR #369.

## Deferred (out of scope, called out in the issue claim)

- `dateFormat` radio control on `DisplaySection` — PR #381 is rewriting that file to shadcn `RadioGroup`; layered cleanly after that lands.
- Calendar / task / recipe long-form date patterns (`"EEEE, MMMM d, yyyy"`, `"EEE, MMM d"`) — those need a UX call before being switched to the numeric `dateFormat` preference and should be tracked separately.

Closes #339


src/lib/format-date.ts
```diff
@@ -0,0 +1,63 @@
...
+ * loose `string`) lets components consume the value through the type
+ * system instead of re-validating at each call site.
+ */
+export type TDateFormat = "MM/DD/YYYY" | "DD/MM/YYYY" | "YYYY-MM-DD";
+
+export const VALID_DATE_FORMATS: readonly TDateFormat[] = [
+  "MM/DD/YYYY",
+  "DD/MM/YYYY",
+  "YYYY-MM-DD",
+] as const;
```
BenSeymourODB, line 16: **Duplicate source of truth for the valid format list.**

`src/app/api/settings/route.ts` still has its own local `const VALID_DATE_FORMATS = ["MM/DD/YYYY", "DD/MM/YYYY", "YYYY-MM-DD"]` as a plain `string[]`. The comment here says they're "kept in sync", but that's a manual contract — any new format requires two edits.

Suggest importing this exported constant into the route instead of re-declaring it, which also eliminates the `as string` cast on the `.includes()` call there.

---

src/components/settings/settings-form.tsx
```diff
@@ -203,6 +204,9 @@ export function SettingsForm({
         createdAt={createdAt}
         providers={providers}
         onDeleteAccount={handleDeleteAccount}
+        dateFormat={
+          isDateFormat(settings.dateFormat) ? settings.dateFormat : undefined
```
BenSeymourODB: **Redundant runtime guard — `settings.dateFormat` is already typed `TDateFormat`.**

`settings` is `UserSettingsData` whose `dateFormat` field is typed as `string` (line 39), so `isDateFormat` does real narrowing here — but only because `UserSettingsData.dateFormat` hasn't been tightened to `TDateFormat` yet. Since `useUserSettings` already validates at the boundary (in `pickCalendarFields`), the value arriving here will always be one of the three documented formats.

The cleaner fix is to type `UserSettingsData.dateFormat` as `TDateFormat`, which lets you write the simpler `dateFormat={settings.dateFormat}` here and removes the footgun of `isDateFormat` returning `false` silently (which would quietly revert the display to the default format mid-session if the guard ever fires).

---

src/lib/__tests__/format-date.test.ts
```diff
@@ -0,0 +1,68 @@
...
+  });
+
+  it("zero-pads single-digit days and months for every format", () => {
+    const singleDigit = new Date(2026, 2, 5); // March 5
+    expect(formatUserDate(singleDigit, "MM/DD/YYYY")).toBe("03/05/2026");
+    expect(formatUserDate(singleDigit, "DD/MM/YYYY")).toBe("05/03/2026");
+    expect(formatUserDate(singleDigit, "YYYY-MM-DD")).toBe("2026-03-05");
+  });
+
+  it("accepts ISO strings as input", () => {
```
BenSeymourODB: **TZ skew risk for the ISO-string test.**

`"2026-01-07T12:00:00"` (no `Z`) is parsed as **local time** by `new Date()`, which means `date-fns`'s `format()` will also read local fields — this is fine and TZ-safe. But the test comment says "ISO strings as input", which could invite a future contributor to change it to `"2026-01-07T00:00:00Z"` (UTC midnight), which _would_ shift a day in negative-UTC-offset CI environments.

Consider renaming the test to `"accepts local-datetime strings as input"` and adding a sibling test that pins a UTC-noon ISO string (`"2026-01-07T12:00:00Z"`) as you did in the `AccountSection` and E2E tests — making the TZ-safety strategy explicit and tested.

---

BenSeymourODB (commented):
Overall the slice is clean: the `formatUserDate` helper is well-structured, the PATTERNS map correctly translates user-facing tokens to date-fns case-sensitive patterns, NaN is handled, the `isDateFormat` bus guard mirrors `isCalendarTransitionSpeed`, and TDD coverage is solid across unit/component/hook/E2E layers. The TZ-safe fixture strategy (local-noon Date construction) is applied consistently.

Three things worth addressing before merge:

1. **Duplicate `VALID_DATE_FORMATS`** — `src/app/api/settings/route.ts` still has its own copy as a plain `string[]`. The new shared export should replace it (see inline).
2. **`UserSettingsData.dateFormat: string` → should be `TDateFormat`** — tightening the local interface removes the need for the `isDateFormat` re-guard in `SettingsForm` and eliminates the silent fallback-to-default footgun (see inline).
3. **ISO-string test label** — minor, but worth a sibling test for UTC-midnight ISO to make the TZ strategy explicit (see inline).

None of these are blockers, but items 1 and 2 are easy wins that close real correctness gaps.

---

src/lib/format-date.ts
BenSeymourODB: Addressed in `5fdcc03`. `src/app/api/settings/route.ts` now imports `VALID_DATE_FORMATS` from `@/lib/format-date` and dropped its local copy. The error message is regenerated from the same constant so the API and the helper can never drift on the format list.

---

src/components/settings/settings-form.tsx
BenSeymourODB: Addressed in `5fdcc03`. `UserSettingsData.dateFormat`, `DisplayValues.dateFormat`, and `MockUserSettings.dateFormat` are all `TDateFormat` now; the `isDateFormat` guard at the `AccountSection` call site is gone (just `dateFormat={settings.dateFormat}`). Narrowing moved up to `src/app/settings/page.tsx` — the only place that bridges Prisma's `string` row into the form — using the same `isDateFormat(...) ? ... : DEFAULT_DATE_FORMAT` pattern as `calendarTransitionSpeed`.

---

src/lib/__tests__/format-date.test.ts
BenSeymourODB: Partially addressed in `5fdcc03`. The test title now reads `"accepts a local-datetime string and reads its local Y/M/D fields"` and an inline comment spells out the TZ-safety contract so a future contributor won't silently swap in a `Z`-suffixed input.

Skipped the sibling UTC-noon test deliberately: any `Z`-suffixed input maps to a different local Y/M/D in some IANA TZ (`12:00:00Z` is the next day in `Pacific/Kiritimati`, the previous day's evening in US zones, etc.). Asserting a specific calendar day against a UTC input requires either pinning the harness TZ (which other tests in the file deliberately avoid) or a UTC-aware sibling helper. Either is more scope than this slice warrants — happy to add it in a follow-up if you'd like the explicit coverage.

---

BenSeymourODB (commented):
## In-depth review — `formatUserDate` helper + wire `dateFormat` to `AccountSection`

Independent review via `/implement-issue`. Merged current `main` into the branch (`9b8ace9` → now part of this push's history), re-verified the full quality gate, then ran a Sonnet sub-reviewer against the helper, the hook integration, `pickCalendarFields`, the `AccountSection` consumer, the test fixture page, and the Playwright spec.

### Verdict — **READY_WITH_NITS** → **MERGE-READY** (concrete fixes applied in `5392ec3`)

Two genuine concerns and one nit were actionable; all applied directly on this branch. Two observations are filed as follow-up issues so they don't get lost (#423, #424).

### Local verification after the rebase + cleanup

| Check | Result |
|---|---|
| `pnpm install --frozen-lockfile` | clean — Prisma client generated via `postinstall` |
| `pnpm lint` | 0 errors, 6 pre-existing warnings (all in unrelated files — `MockCalendarProvider`, `account-section`'s ``, `profile-avatar`, etc.) |
| `pnpm format` | clean |
| `pnpm test:run` (full) | **157 files / 2559 tests passing** (+3 new for `useDateFormat`) |
| `pnpm test:run` on the PR's touched files | **59 / 59 passing** post-cleanup |

### What's good

- **TZ handling in the unit tests is exemplary.** `format-date.test.ts` uses `new Date(year, month, day)` (local constructor) throughout, with an inline comment explaining *why* the `Z` suffix is intentionally omitted in the one ISO-string test. That's the right pattern to model for future formatter tests.
- **The `PATTERNS` map carries case-sensitivity commentary** — the note that `"YYYY"` in user-facing tokens is *not* the date-fns ISO week-year token (`"YYYY"` vs `"yyyy"`) is exactly the footgun documentation that prevents a future editor from "correcting" the map.
- **`pickCalendarFields` narrowing is sound.** Using `isDateFormat()` at both the HTTP-GET boundary and the bus-subscription boundary, with a single shared implementation, means there's no duplicated allow-list that could drift. The `VALID_DATE_FORMATS` → `PATTERNS` → `isDateFormat` chain is tight.
- **Single source of truth for valid formats.** The PUT validator in `src/app/api/settings/route.ts` now imports `VALID_DATE_FORMATS` from `src/lib/format-date.ts` (commit `5fdcc03`'s refactor) — the hardcoded list is gone.
- **CLAUDE.md compliance is clean across the board** — no `any`, strict TS, no manual memoization, standard Tailwind palette, TDD ordering preserved (tests reference the helper before the helper exists).

### Findings applied in `5392ec3`

**🟡 CONCERN 1 — `account-section.test.tsx` TZ bug at noon-UTC** → fixed.

The original `const ISO = "2026-03-05T12:00:00Z"` looked TZ-safe but actually rolls forward a day in UTC+13 (`Pacific/Auckland` during DST) and UTC+14 (`Pacific/Kiritimati`):

```
TZ=Pacific/Kiritimati node -e "
  const d = new Date('2026-03-05T12:00:00Z');
  console.log(d.getMonth()+1, d.getDate()); // → 3 6
"
```

`date-fns`'s `format()` reads local calendar fields, so the rendered glyphs would be `"03/06/2026"` and all three `dateFormat` assertions would fail. CI runs UTC (so the existing CI is unaffected), but a contributor running tests in Auckland or Samoa would see a red suite with no code change. Switched to a local-constructed `Date` (`new Date(2026, 2, 5, 12, 0, 0).toISOString()`) so the rendered date pins to March 5 in every host TZ.

**🟡 CONCERN 2 — Plan specifies `useDateFormat()` hook; implementation omitted it** → added.

`.claude/plans/date-format-wireup-339.md` lines 82–97 specifies the convenience hook explicitly: *"Thin wrapper around `useUserSettings()` for the common case where a component just wants a memo-stable formatter."* It was absent from the shipped code, which would push future calendar/task/recipe consumers to either call `formatUserDate(x, useUserSettings().settings.dateFormat)` at every site or — worse — copy the pattern inconsistently across surfaces.

Added `src/hooks/useDateFormat.ts` returning `{ dateFormat, format }`. **Placed in `src/hooks/` rather than `src/lib/format-date.ts`** because `useUserSettings.ts` already imports from `format-date.ts` for `pickCalendarFields`, and colocating the hook there would create a `lib → hooks → lib` cycle. The leaf-module discipline matters here — `format-date.ts` is also imported by server-only call sites (the PUT validator), and pulling a React hook into it would force a `"use client"` boundary where none is needed today. Three new tests cover the default-when-unauthenticated, server-pickup, and garbage-rejection paths.

**🔵 NIT — `as TDateFormat` cast in the writer fixture** → fixed.

`/test/date-format-sync` populates the `<select>` from `VALID_DATE_FORMATS`, so the runtime value is always valid, but the cast (`e.target.value as TDateFormat`) silently bypassed the `isDateFormat` guard the rest of the PR establishes. Routed the onChange through `isDateFormat(next)` so the fixture models the production guard pattern.

### Observations filed as follow-up issues

**🔵 NIT 3 → #423.** `UserSettingsData.timeFormat: string` while `dateFormat: TDateFormat` — pre-existing asymmetry, not introduced here. The PR makes the narrowed `dateFormat` path the more idiomatic one going forward; the loose `timeFormat` should follow. Tracked as a small refactor.

**🟡 CONCERN 4 → #424.** `SettingsForm` keeps a local `useState` copy of fields that also live in `useUserSettings`; it emits to the bus on PUT but doesn't subscribe. Today this is benign because `SettingsForm` is the only writer for production fields, but the first second-writer (a calendar-side `dateFormat` control, the QA fixture page sharing a tab) creates a real lost-write hazard — the stale form value would clobber a fresh mutation when the user next saves. Pre-existing architectural debt; flagging now so PR #381's `DisplaySection` refactor and any future settings work can consolidate the source of truth.

**🔵 NIT 5 (recorded, no action) — dead `??` fallback in `formatUserDate`.**
`PATTERNS[dateFormat] ?? PATTERNS[DEFAULT_DATE_FORMAT]`: the `??` branch is unreachable under the type system since `PATTERNS` is `Record<TDateFormat, string>` and `dateFormat` is typed. Kept as a runtime belt-and-suspenders — the comment explaining the future-locale refactor intent is enough context for a future maintainer, and removing it would change runtime behaviour if a `Record<K, V>` ever becomes a `Partial<Record<K, V>>`.

### Recommendation

**Ready to merge once CI confirms green on `5392ec3`.** The shipped slice does exactly what the plan promises: a centralised numeric-format mapper, type-safe end-to-end across the bus, server validator, and consumer, with a working production wireup (`AccountSection`) and an end-to-end sync test page. The two follow-up issues are honest deferrals — neither is regressed by this PR — and the `useDateFormat` hook restores plan-vs-implementation parity for future consumers.

---
_Reviewed by `/implement-issue` PR-review run via Claude Code — independent Sonnet sub-review, findings applied directly. Branch is up-to-date with `main` after the merge in `9b8ace9`._

---

PR #299, opened 2026-05-05
feat(calendar): intra-view slide animation in Day/Week/Year (follow-up to #250)

BenSeymourODB:
Follow-up to merged PR #250, which landed only the `useSlideDirection` hook + plan; this delivers the actual intra-view slide wiring that #207 asked for.

## Summary
- Wraps the body of `DayCalendar`, `WeekCalendar`, and `YearCalendar` in `AnimatedSwap` so prev/next navigation slides directionally — matching the month slide already in `SimpleCalendar`.
- Direction comes from `useSlideDirection(index)` (PR #250) using:
  - **Day** — `floor(startOfDay(selectedDate) / 86_400_000)`
  - **Week** — `floor(startOfDay(weekStart) / (86_400_000 * 7))` (intra-week date moves don't trigger a slide)
  - **Year** — `selectedDate.getFullYear()`
- Component tests for each view: wrapper presence, forward/backward `translateX`, intra-period no-op (Week, Year), reduced-motion bypass.
- Playwright video spec `e2e/intra-view-transitions.spec.ts` (`video: "on"`) covering Day/Week/Year prev-next + reduced-motion path.
- Drive-by: pre-existing tsc miss in `AgendaList.test.tsx` (no `deleteEvent` mock factory) was fixed in PR #250 — already on main.

## Why open after #207 closed
PR #250 was merged before phases 2/3 had landed; the issue auto-closed but its first acceptance criterion ("Switching between any pair of (Month, Agenda, Year, Week, Day) animates with the same `AnimatedSwap` API") was only half-delivered (cross-view fade existed; intra-view slide was missing for Day/Week/Year). This PR completes that criterion. The configurable-duration / settings-panel work is still **deferred** for a separate slice that touches Prisma + UI.

`prefers-reduced-motion` is short-circuited inside `AnimatedSwap`, so the off-path is inherited automatically — verified by component + E2E tests.

## Test plan
- [x] `pnpm exec vitest run src/components/calendar/__tests__/{DayCalendar,WeekCalendar,YearCalendar}.test.tsx` (87/87 pass)
- [x] `pnpm exec vitest run src/hooks/__tests__/use-slide-direction.test.ts` (5/5 pass; from #250)
- [x] `pnpm lint:fix && pnpm format:fix` clean
- [x] `pnpm check-types` clean for files in scope (pre-existing `src/components/ui/resizable.tsx` errors on main are unrelated)
- [ ] CI runs `e2e/intra-view-transitions.spec.ts` with video capture

[...]

PR #257, opened 2026-05-03
feat(calendar): add IEvent.category + group-by-category agenda mode (#211)

BenSeymourODB:
## Summary

Closes #211.

Adds an optional `category` field on `IEvent`, sourced from Google Calendar's `extendedProperties.shared.category` (preferred) or `.private.category` (fallback), and exposes a third "Group by category" mode in `AgendaCalendar` so users can browse upcoming events by category. Events without a category land under an "Uncategorised" bucket.

## Phases

- **Phase 1 — data model + mapper** (commit `d9e8d83`): widened `IEvent`, extracted `category` in `transformGoogleEvent`, added 7 unit tests for shared/private precedence + whitespace handling.
- **Phase 2 — provider union widening + agenda UI** (commit `07bb473`):
  - New shared `TAgendaGroupBy = "date" | "color" | "category"` type so the provider, mock provider, settings panel, and `AgendaCalendar` all reference one union. Pre-#211 persisted `"date" | "color"` payloads validate against the widened type without a runtime migration.
  - `groupEventsByCategory` + `sortCategoryEntries` helpers — alphabetical case-insensitive locale-aware ordering, `"Uncategorised"` pinned last, whitespace-only categories treated as absent.
  - Third toggle button ("Group by category") in `AgendaCalendar`, third radio option ("By category") in `CalendarSettingsPanel`, parity in `AgendaList` (used by Day/Week agenda mode).
  - Component tests cover alphabetical order, "Uncategorised" placement (present and absent), per-bucket count badges, search-within-category, toggle wire-up, and the whitespace edge case.
  - Provider tests verify `setAgendaModeGroupBy("category")` round-trips through `localStorage` and rehydrates on next mount.
- **Phase 3 — sweep + final**: confirmed no other consumer of the `"date" | "color"` union remained (`src/test/fixtures/calendar-context.ts` already passes the string literal `"date"`, which still satisfies the widened union; `useUserSettings.weekStartDay` migration is unrelated). Full quality gate green.

## Test plan

- [x] `pnpm test` — 2347/2347 pass (10 new tests for Phase 2)
- [x] `pnpm check-types` — clean
- [x] `pnpm lint:fix && pnpm format:fix` — clean (no new warnings)
- [x] Existing E2E specs (`agenda-toggle.spec.ts`, `agenda-search.spec.ts`, `calendar-settings-panel.spec.ts`) unaffected — the new button is additive and the existing tests target the date/color buttons by accessible name.

## Out of scope

[...]

PR #391, opened 2026-06-08
feat(calendar): fold AgendaCalendar search into AgendaList (#264)

BenSeymourODB:
## Summary

- Restore the search input that day/week agenda mode lost in PR #195 — title / description / attendee substring filter, clear control, match count, sr-only live region — by porting it from `AgendaCalendar` into `AgendaList`.
- Extract `filterEventsBySearch` into a shared `agenda-helpers.ts` module so both renderers depend on one implementation. `AgendaCalendar.tsx` re-exports it for back-compat with the existing `agenda-helpers.test.ts` (no test changes required).
- Distinguish the two empty states: contextual `emptyLabel` when no events fall in the range (search input hidden — no dead control), and `No events match "<query>"` when a search returns nothing in a non-empty range (search input stays mounted so the user can clear).
- Key the `AgendaList` element on `rangeStart` at both call sites (`DayCalendar.tsx`, `WeekCalendar.tsx`) so React remounts it on prev/next navigation — prevents a search query typed on one day/week from silently carrying over.

## Plan

Linked plan: `.claude/plans/fold-agenda-search-264.md`

## Review feedback addressed (round 1)

- **Suggestion 1** (UX): key `AgendaList` on `rangeStart` at both call sites — DayCalendar.tsx, WeekCalendar.tsx (Option A from review).
- **Suggestion 2** (test gap): added a search + color-groupBy test in `AgendaList.test.tsx`.
- **Suggestion 3** (test gap): extended the no-matches empty-state test to click the clear control and assert events return.
- **Nit 2** (plan divergence): updated `.claude/plans/fold-agenda-search-264.md` to reflect `aria-live="polite" aria-atomic="true"` and the re-export-based migration of `agenda-helpers.test.ts`.

**Deferred:**

- **Nit 1** (Webkit double clear control): pre-existing identical concern in `AgendaCalendar`. Filed as #411 to fix in both components together via a single global CSS rule.

## Merge resolution

Main advanced past PR #391's original base. Merged `origin/main` (commit `4052f56`) and resolved conflicts in `AgendaList.tsx` / `DayCalendar.tsx` / `WeekCalendar.tsx`:

- `AgendaList.tsx`: integrated this PR's search input with main's new `agendaModeGroupBy === "category"` branch (from PR #257). Both color and category groupings now drive off the search-`filtered` list, and both stay gated on `!noMatches` so the dedicated no-match empty state short-circuits all groupings.
- `DayCalendar.tsx` / `WeekCalendar.tsx`: kept main's `AnimatedSwap` slide-animation wrapper (from PR #299) and moved the `key={…}` prop onto the `AgendaList` child inside the swap.
- `DayCalendar.test.tsx` / `WeekCalendar.test.tsx`: the search-reset assertions now query inside `animated-swap-incoming`, since the slide animation keeps the previous `AgendaList` mounted as the outgoing snapshot during the transition.

## Test plan

- [x] `pnpm test` — 2532 / 2532 passing
- [x] `pnpm lint:fix` — 0 errors (6 pre-existing warnings in untouched files)
- [ ] Manual: open `/calendar?view=day&agenda=1` and `/calendar?view=week&agenda=1`, type into the search box, confirm filtering, clear, no-matches behaviour, and that navigating to the next day/week clears the query (matches `/test/calendar?view=agenda`). Cross-check that the `agendaModeGroupBy="category"` branch in `AgendaList` also honours the search query.

## Out of scope

Full retirement of `AgendaCalendar.tsx`. `/test/calendar?view=agenda` still mounts the legacy renderer; removing the file is tracked separately.

Closes #264

### DIFF
--- a/.claude/plans/fold-agenda-search-264.md
+++ b/.claude/plans/fold-agenda-search-264.md
@@ -29,8 +29,11 @@ removing it.
 Move `filterEventsBySearch` into a new pure module
 `src/components/calendar/agenda-helpers.ts`. Both `AgendaCalendar.tsx` and
 `AgendaList.tsx` import from there. The helper is already heavily tested in
-`agenda-helpers.test.ts` (which currently imports from `AgendaCalendar`); the
-test file moves its import to the new module path with no test changes.
+`agenda-helpers.test.ts` (which currently imports from `AgendaCalendar`). To
+keep the diff minimal and avoid touching the existing test file, the
+implementation re-exports `filterEventsBySearch` from `AgendaCalendar.tsx`,
+so `agenda-helpers.test.ts` continues to import from `../AgendaCalendar`
+unchanged.
 
 We do **not** also relocate `sortEventsByStartTime` / `groupEventsByColor` /
 `groupEventsByDate` / `parseDateKey` / `capitalize` / `getColorClasses`. Those
@@ -72,9 +75,21 @@ days).
 
 - `aria-label="Search events"` on the input.
 - `aria-label="Clear search"` on the clear button.
-- `role="status"` sr-only live region announces match count while typing.
+- `aria-live="polite" aria-atomic="true"` on an sr-only `<p>` announces the
+  match count while typing. (Functionally equivalent to `role="status"`,
+  which is just syntactic sugar for the same ARIA implicit live region.)
 - `aria-hidden="true"` on the visible match-count span so the live region is
   the sole announcement vector.
+
+### Resetting search on date navigation
+
+`AgendaList` keeps `searchQuery` in local component state. When the user
+navigates to the next/previous day or week, `DayCalendar` and `WeekCalendar`
+re-render with new `rangeStart`/`rangeEnd` props but the `AgendaList`
+component instance is preserved, so the search query would silently carry
+over. The call sites in `DayCalendar.tsx` and `WeekCalendar.tsx` set
+`key={rangeStart.toISOString()}` on `<AgendaList>` so React remounts it on
+navigation and the search query (plus any open modal) resets cleanly.
 
 ## TDD plan
 
--- a/.claude/plans/fold-agenda-search-264.md
+++ b/.claude/plans/fold-agenda-search-264.md
@@ -1,0 +1,114 @@
+# Fold AgendaCalendar search into AgendaList day/week agenda mode (#264)
+
+## Context
+
+After PR #195 (day/week agenda toggle), the production `ViewSwitcher` no longer
+mounts the search-driven `AgendaCalendar`; day/week agenda mode renders the
+slimmer `AgendaList`. The search input, match count, and clear control that
+landed in PR #144 only live on `/test/calendar?view=agenda`. Users have lost a
+visible capability.
+
+## Goal
+
+Port the search input UI from `AgendaCalendar` into `AgendaList` so day and
+week agenda mode filter their rendered events by query the same way
+`AgendaCalendar` does, without changing the legacy `AgendaCalendar` UX or
+removing it.
+
+## Out of scope
+
+- Retiring `AgendaCalendar.tsx`. `/test/calendar?view=agenda` still mounts it,
+  and removal is tracked separately.
+- Replacing the per-view header search affordance the calendar surface already
+  ships elsewhere; this is strictly the agenda-mode-internal filter.
+
+## Design
+
+### Shared helper
+
+Move `filterEventsBySearch` into a new pure module
+`src/components/calendar/agenda-helpers.ts`. Both `AgendaCalendar.tsx` and
+`AgendaList.tsx` import from there. The helper is already heavily tested in
+`agenda-helpers.test.ts` (which currently imports from `AgendaCalendar`); the
+test file moves its import to the new module path with no test changes.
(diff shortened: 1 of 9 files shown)


BenSeymourODB (commented):
## Code Review — feat(calendar): fold AgendaCalendar search into AgendaList (#264)

Overall the PR is well-structured, CLAUDE.md compliant (no manual memoization, no `any`, standard Tailwind palette), and the TDD approach is solid. No blockers found. Three suggestions and two nits follow.

---

### `src/components/calendar/AgendaList.tsx` — Suggestion 1 (UX / state hygiene)

**`searchQuery` is not reset when the date range changes (day/week navigation).**

`AgendaList` stays mounted as the same component instance when the user presses "Next Day" or "Next Week" — only the `rangeStart`/`rangeEnd` props change. Because `searchQuery` lives in local state, a query typed on Monday carries over silently to Tuesday. In the no-events case (day is empty) the early-return fires before the search box renders, so the stale query is invisible but still held in state. When the user navigates back to Monday it reappears.

Fix options (cheapest first):

```tsx
// Option A — give AgendaList a key based on the range, so React
// remounts it (and resets all state) on navigation.
// In DayCalendar.tsx / WeekCalendar.tsx:
<AgendaList
  key={rangeStart.toISOString()}   // ← forces remount on nav
  events={...}
  rangeStart={rangeStart}
  rangeEnd={rangeEnd}
  ...
/>
```

```tsx
// Option B — useEffect inside AgendaList that resets on prop change:
useEffect(() => { setSearchQuery(""); }, [rangeStart, rangeEnd]);
```

Option A is one line each in `DayCalendar.tsx` and `WeekCalendar.tsx` and avoids leaking concern into `AgendaList` itself.

---

### `src/components/calendar/AgendaList.tsx` — Suggestion 2 (test coverage gap)

**Search + `agendaModeGroupBy === "color"` interaction is not exercised.**

The new test suite defaults to `agendaModeGroupBy: "date"`. The color-groupBy path has its own existing test, but the combination of "active search query + color groupBy" is untested. It's worth adding one assertion in the describe block:

```tsx
it("filters events by query when grouped by color", async () => {
  renderList({ events: [morning, afternoon], ...range },
    { agendaModeGroupBy: "color" });

  await user.type(screen.getByTestId("agenda-list-search-input"), "soccer");

  const titles = screen
    .getAllByTestId("agenda-list-event-title")
    .map((el) => el.textContent);
  expect(titles).toEqual(["Soccer Practice"]);
});
```

---

### `src/components/calendar/__tests__/AgendaList.test.tsx` — Suggestion 3 (incomplete no-match reset coverage)

**The no-match empty-state test does not assert that clearing restores the list.**

The plan's item 3 says "clearing brings the results back," but the no-matches test (lines 232–247) only checks that the no-matches card appears; it never clicks clear. The clear test (lines 249–267) starts from a _matching_ query. Neither test covers the path: type non-matching query → clear → events reappear. One extra `user.click` + assertion at the end of the existing no-matches test would close this gap.

---

### `src/components/calendar/AgendaList.tsx` — Nit 1 (double clear control on Webkit)

`<Input type="search">` renders a browser-native cancel button in Webkit/Blink when the field is non-empty. Combined with the custom `<X>` clear button that also appears when `searchActive`, Webkit users see two overlapping clear controls simultaneously. This pre-exists identically in `AgendaCalendar` so it's not a regression, but it's worth fixing in both components at the same time:

```css
/* in globals.css or Input component */
input[type="search"]::-webkit-search-cancel-button { display: none; }
```

---

### `.claude/plans/fold-agenda-search-264.md` — Nit 2 (plan/implementation divergence)

The plan (line 75) says the live region should use `role="status"`, but the implementation uses `aria-live="polite" aria-atomic="true"` (which is functionally equivalent — `role="status"` is simply syntactic sugar for both). The plan also says `agenda-helpers.test.ts` "moves its import to the new module path with no test changes" — the file still imports from `AgendaCalendar` via the re-export. Both are harmless, but worth updating the plan to reflect reality to avoid confusing future readers.

---

### Confirmed green ✓

- `agenda-list` testid move: inner card still renders whenever `windowed.length > 0`, which is the case in every E2E scenario (`events=default` always has today's events). All `toBeVisible()` and `toHaveCount(0)` assertions in `agenda-toggle.spec.ts`, `week-day-views.spec.ts`, `calendar.spec.ts`, `calendar-transitions.spec.ts`, and `calendar-transition-speed.spec.ts` continue to pass correctly.
- `DayCalendar.test.tsx` line 538 and `WeekCalendar.test.tsx` line 550: both render events in range → `agenda-list` still in DOM. ✓
- `DayCalendar.test.tsx` line 544 (`agenda-list-empty`): empty-range early return path unchanged. ✓
- No manual memoization, no `any`, standard Tailwind palette throughout.
- Live region: `<p>` always in DOM before content inserted — correct AT pattern. `aria-hidden="true"` on visible count span prevents double-announce. ✓
- `colorGroups` conditional ordering in JSX matches `noMatches` guard — no branch can reach color-group rendering while `noMatches` is true. ✓
- `filterEventsBySearch` re-export from `AgendaCalendar` correctly maintains back-compat for `agenda-helpers.test.ts` and any downstream importers.

---

Issue #212, opened 2026-05-02
feat(calendar): event creation dialog Phase 2 — reminders, recurrence, richer date pickers


BenSeymourODB:
## Context

Deferred from PR #143 ("event creation dialog"):

> Reminders, recurrence, richer date pickers — Phase 2 of the calendar feature set.

## Scope

- **Reminders:** add a "Reminders" section to `EventCreateDialog` allowing one or more `{ method: 'popup' | 'email', minutesBefore: number }` entries. Persist via Google Calendar's `reminders.overrides`.
- **Recurrence:** add a recurrence editor (None / Daily / Weekly / Monthly / Yearly + custom RRULE). Map to Google's `recurrence: ["RRULE:..."]`.
- **Richer date pickers:** swap the current basic inputs for the shadcn date+time picker, including all-day toggle and end-time auto-adjust on start change.

## Acceptance criteria

- [ ] Create-dialog UI exposes reminders, recurrence, and rich pickers
- [ ] Created events round-trip through Google Calendar with the chosen options preserved
- [ ] Recurrence editor handles common cases without exposing raw RRULE to the user (advanced users can still input RRULE)
- [ ] Component tests cover each new control's serialization
- [ ] Playwright E2E creating a weekly recurring event with a 30-min popup reminder

## Dependencies

- Blocked by #116 (write-through to Google Calendar API). Phase 2 is meaningless until events are persisted server-side.

PR #392, opened 2026-06-08
feat(calendar): event detail edit (PATCH) — completes #115 (#265)

BenSeymourODB:
## Summary

Completes the **edit half** of #115 (the delete half landed via #197). Adds the PATCH path end-to-end across the server, the provider, and the UI.

## What ships

**Server** (`src/lib/calendar/event-body.ts`, `src/app/api/calendar/events/[id]/route.ts`)
- `validateEventBody` + `buildGoogleEventBody` extracted from the POST route so PATCH reuses the same wire format and validator. POST behaviour is unchanged.
- `PATCH /api/calendar/events/[id]?calendarId=...` forwards to Google Calendar `events.patch`. The `calendarId` query string is authoritative; a `calendarId` field in the body is ignored.
- Auth + error shape mirrors the DELETE handler (401 / `requiresReauth`, 403 / 404 / 502 / 500). On success returns 200 with the normalised event so the client can reconcile its optimistic row.

**Provider** (`src/components/providers/CalendarProvider.tsx`, `MockCalendarProvider.tsx`)
- `editEvent(eventId, calendarId, input): Promise` — optimistic local merge, PATCHes the API, reconciles with the server response through `transformGoogleEvent`. Rolls back the original row on failure.
- Rejects without issuing a PATCH when the event isn't in the local list.
- Mock provider + test fixture (`src/test/fixtures/calendar-context.ts`) gain the same field.

**UI** (`src/components/calendar/EventDetailModal.tsx`, `src/hooks/useEventEdit.ts`)
- `EventDetailModal` gains an optional `onEdit` prop. When supplied and the calendar is writable, an "Edit event" button swaps the read-only body for an inline form (title / all-day / start / end / color / description) seeded from the event. Save → `onEdit(event, patch)`. Success closes the modal; rejection keeps the form open.
- Date parser / formatter helpers extracted into `src/lib/calendar/date-input.ts` so `EventCreateDialog` and the new edit form share one source of truth.
- `useEventEdit` wraps `editEvent` with sonner toasts, mirroring `useEventDelete`. Wired into `SimpleCalendar`, `AnalogClockView`, and `AgendaCalendar` — the three surfaces that already pass `onDelete`.

## Test plan

- [x] Phase 1: 18 route tests — happy path (timed + all-day), 400 (missing calendarId / id / title / color, malformed JSON, end-before-start, ISO-datetime-for-all-day), 401 / 403 / 404 / 502 / 500, URL-encoding, body-`calendarId`-ignored
- [x] Phase 2: 4 `CalendarProvider.editEvent` tests — happy path, 403 rollback, 502 rollback, missing-row early-reject
- [x] Phase 2: 7 `EventDetailModal` tests — button gating, form open / cancel / save, error keeps form open, all-day toggle wire format, in-flight disabled state
- [x] **Review fix (ed9c28d):** regression test asserts `isEditing` / `confirmingDelete` reset across modal close→reopen with a different event
- [x] `pnpm test` — **2645 / 2645 passing**
- [x] `pnpm lint:fix` — 0 errors, 5 pre-existing warnings
- [ ] Manual browser smoke (deferred — no local browser in this run; covered by acceptance criteria when reviewer pulls)

## Out of scope (tracked separately)

- Attendees / reminders / recurrence — #212 / #213
- Calendar-id move (PATCH only mutates the same-calendar event)
- Local Postgres write path — Google stays the source of truth until the DB lands
- Wiring `onEdit` (and `onDelete`) into `AgendaList` (day/week agenda mode) — tracked as **#451**

## Plan

`.claude/plans/event-detail-edit-265.md`. Project: https://github.com/users/BenSeymourODB/projects/1.

Closes #265.

### DIFF
--- a/src/components/calendar/AnalogClockView.tsx
+++ b/src/components/calendar/AnalogClockView.tsx
@@ -60,7 +60,7 @@ function isAllDayToday(event: IEvent, today: Date): boolean {
  * events in a sibling list so they remain visible in this view.
  */
 export function AnalogClockView() {
-  const { events, use24HourFormat } = useCalendar();
+  const { events, use24HourFormat, getAccessRole } = useCalendar();
   const today = useDateNow();
   const [selectedEvent, setSelectedEvent] = useState<IEvent | null>(null);
   const triggerRef = useRef<HTMLElement | SVGElement | null>(null);
@@ -196,6 +196,9 @@ export function AnalogClockView() {
         returnFocusTo={triggerRef}
         onDelete={handleDelete}
         onEdit={handleEdit}
+        accessRole={
+          selectedEvent ? getAccessRole(selectedEvent.calendarId) : undefined
+        }
       />
     </div>
   );
--- a/src/components/calendar/AnalogClockView.tsx
+++ b/src/components/calendar/AnalogClockView.tsx
@@ -4,6 +4,7 @@ import { AnalogClock } from "@/components/calendar/analog-clock";
 import { useCalendar } from "@/components/providers/CalendarProvider";
 import { ThemeScope } from "@/components/theme/theme-scope";
 import { useEventDelete } from "@/hooks/useEventDelete";
+import { useEventEdit } from "@/hooks/useEventEdit";
 import { getColorClass } from "@/lib/calendar-helpers";
 import { useDateNow } from "@/lib/hooks/use-date-now";
 import type { IEvent } from "@/types/calendar";
@@ -64,6 +65,7 @@ export function AnalogClockView() {
   const [selectedEvent, setSelectedEvent] = useState<IEvent | null>(null);
   const triggerRef = useRef<HTMLElement | SVGElement | null>(null);
   const handleDelete = useEventDelete();
+  const handleEdit = useEventEdit();
   const { resolvedTheme } = useTheme();
 
   // Emphasize the clock face with a light scope (issue #319). Only meaningful
@@ -193,6 +195,7 @@ export function AnalogClockView() {
         use24HourFormat={use24HourFormat}
         returnFocusTo={triggerRef}
         onDelete={handleDelete}
+        onEdit={handleEdit}
       />
     </div>
   );
--- a/src/components/calendar/EventDetailModal.tsx
+++ b/src/components/calendar/EventDetailModal.tsx
@@ -36,7 +36,7 @@ import type {
   TCalendarAccessRole,
   TEventColor,
 } from "@/types/calendar";
-import { type RefObject, useId, useState } from "react";
+import { type RefObject, useEffect, useId, useState } from "react";
 import { format, isSameDay } from "date-fns";
 import { Pencil, Trash2 } from "lucide-react";
 
@@ -431,6 +431,18 @@ export function EventDetailModal({
   const [confirmingDelete, setConfirmingDelete] = useState(false);
   const [isDeleting, setIsDeleting] = useState(false);
   const [isEditing, setIsEditing] = useState(false);
+
+  // The parents keep `<EventDetailModal>` permanently mounted and flip the
+  // `event` prop between `null` (closed) and an `IEvent` (open). React's
+  // state hooks survive the `null` render, so `isEditing` / `confirmingDelete`
+  // would leak across opens — re-opening the modal for a different event
+  // would land the user mid-edit (or in the delete confirmation) without
+  // ever clicking the trigger. Reset both whenever the `event.id` changes
+  // so each open starts in the read-only view.
+  useEffect(() => {
+    setIsEditing(false);
+    setConfirmingDelete(false);
+  }, [event?.id]);
 
   if (!event) return null;
 
--- a/src/components/calendar/EventDetailModal.tsx
+++ b/src/components/calendar/EventDetailModal.tsx
@@ -12,6 +12,7 @@ import {
 } from "@/components/ui/alert-dialog";
 import { Avatar, AvatarFallback, AvatarImage } from "@/components/ui/avatar";
 import { Button } from "@/components/ui/button";
+import { Checkbox } from "@/components/ui/checkbox";
 import {
   Dialog,
   DialogContent,
@@ -20,15 +21,24 @@ import {
   DialogHeader,
   DialogTitle,
 } from "@/components/ui/dialog";
+import { Input } from "@/components/ui/input";
+import { Label } from "@/components/ui/label";
+import { Textarea } from "@/components/ui/textarea";
+import {
+  parseDateOnly,
+  parseDateTimeLocal,
+  toDateOnly,
+  toDateTimeLocal,
+} from "@/lib/calendar/date-input";
 import { canWriteToCalendar } from "@/lib/google-calendar-mappers";
 import type {
   IEvent,
   TCalendarAccessRole,
   TEventColor,
 } from "@/types/calendar";
-import { type RefObject, useState } from "react";
+import { type RefObject, useId, useState } from "react";
 import { format, isSameDay } from "date-fns";
-import { Trash2 } from "lucide-react";
+import { Pencil, Trash2 } from "lucide-react";
 
 const COLOR_INDICATOR_CLASSES: Record<TEventColor, string> = {
   blue: "bg-blue-500",
@@ -38,6 +48,37 @@ const COLOR_INDICATOR_CLASSES: Record<TEventColor, string> = {
   purple: "bg-purple-500",
   orange: "bg-orange-500",
 };
+
+const COLOR_OPTIONS: readonly {
+  value: TEventColor;
+  label: string;
+  swatch: string;
+}[] = [
+  { value: "blue", label: "Blue", swatch: "bg-blue-500" },
+  { value: "green", label: "Green", swatch: "bg-green-500" },
+  { value: "red", label: "Red", swatch: "bg-red-500" },
+  { value: "yellow", label: "Yellow", swatch: "bg-yellow-500" },
+  { value: "purple", label: "Purple", swatch: "bg-purple-500" },
+  { value: "orange", label: "Orange", swatch: "bg-orange-500" },
+];
+
+/**
+ * Edit-form patch payload. Matches `EditEventInput`'s body shape from
+ * `CalendarProvider` minus `calendarId` (the modal doesn't know about
+ * other calendars; the parent forwards the event's own `calendarId`).
+ *
+ * `isAllDay = true` ⇒ `startDate` / `endDate` are `YYYY-MM-DD` strings
+ * (matching the `POST` / `PATCH` route wire format). `isAllDay = false`
+ * ⇒ ISO datetime strings.
+ */
+export interface EventEditPatch {
+  title: string;
+  description: string;
+  color: TEventColor;
+  isAllDay: boolean;
+  startDate: string;
+  endDate: string;
+}
 
 interface EventDetailModalProps {
   event: IEvent | null;
@@ -58,12 +99,23 @@ interface EventDetailModalProps {
    */
   onDelete?: (event: IEvent) => Promise<void>;
   /**
+   * Optional edit handler (#265). When provided, an "Edit event" button
+   * renders in the footer. Activating it swaps the read-only body for an
+   * inline edit form; saving calls `onEdit(event, patch)`. On success the
+   * modal closes; on rejection the form stays open so the caller can show
+   * a toast and the user can retry.
+   *
+   * Gated by `accessRole` exactly like delete — `reader` / `freeBusyReader`
+   * hides the button (Google's server-side 403 is still the backstop).
+   */
+  onEdit?: (event: IEvent, patch: EventEditPatch) => Promise<void>;
+  /**
    * The user's permission level on the event's source calendar (#266).
-   * When `reader` or `freeBusyReader`, mutating actions (currently the
-   * delete button) are hidden — Google's server-side 403 stays as the
-   * backstop for races where the role changes mid-session. Treat
-   * `undefined` as "unknown / writable" so a not-yet-loaded calendar
-   * list doesn't accidentally hide the button on owned calendars.
+   * When `reader` or `freeBusyReader`, mutating actions (delete, edit) are
+   * hidden — Google's server-side 403 stays as the backstop for races
+   * where the role changes mid-session. Treat `undefined` as "unknown /
+   * writable" so a not-yet-loaded calendar list doesn't accidentally hide
+   * the buttons on owned calendars.
    */
   accessRole?: TCalendarAccessRole;
 }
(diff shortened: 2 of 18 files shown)


src/lib/calendar/event-body.ts
```diff
@@ -0,0 +1,247 @@
...
+  if (event.isAllDay) {
+    body.start.date = event.startDateStr;
+    body.end.date = addOneDay(event.endDateStr);
+  } else {
+    body.start.dateTime = event.start.toISOString();
+    body.end.dateTime = event.end.toISOString();
+  }
+
+  return body;
+}
```
BenSeymourODB, line 277: **BLOCKER — PATCH body doesn't clear the opposite date field when switching event type**

Google Calendar's `events.patch` merges the supplied body over the existing event. If you send `start: { dateTime: "…" }` on an event that was previously all-day, Google still has `start.date` set and will return a 400 because an event cannot have both `date` and `dateTime` on the same time object. The same breakage happens in reverse (all-day patch on a timed event).

Fix: explicitly null out the unused field in the returned body:

```ts
if (event.isAllDay) {
  body.start = { date: event.startDateStr, dateTime: undefined };
  body.end   = { date: addOneDay(event.endDateStr), dateTime: undefined };
} else {
  body.start = { dateTime: event.start.toISOString(), date: undefined };
  body.end   = { date: undefined, dateTime: event.end.toISOString() };
}
```

Setting fields to `undefined` causes `JSON.stringify` to omit them, which is what you want for INSERT too — so this change is safe for the POST path and fixes the PATCH path. This could also be solved by sending `null` explicitly (Google accepts `null` to clear a field), but the `undefined` approach keeps the schema clean. A test for "all-day patch on a previously-timed event" and vice-versa would lock this in.

---

src/components/providers/CalendarProvider.tsx
```diff
@@ -664,6 +684,105 @@ export function CalendarProvider({
...
+      logger.event("CalendarEventUpdated", {
+        eventId: reconciled.id,
+        calendarId,
+      });
+
+      return reconciled;
+    } catch (error) {
+      // Rollback: restore the original row in place. Use a functional update
+      // so any concurrent edits to *other* rows are preserved.
+      setAllEvents((current) =>
```
BenSeymourODB: **BLOCKER — rollback overwrites a refresh-removed event with a stale `original` row**

The rollback uses a functional updater (`setAllEvents(current => current.map(…))`), which is good for preserving concurrent edits to *other* rows. However, it unconditionally maps every row: if `refreshEvents()` ran while the PATCH was in-flight and the server no longer returns the event (e.g. it was deleted on another device), the refresh will have removed the row from `current`, so `current.map(e => e.id === eventId ? original : e)` won't find the ID — and `original` is silently dropped from the list altogether. The calendar then shows the event missing, which is actually correct behaviour, but the *next* issue is that rolling back a row that was legitimately deleted by a concurrent refresh puts a ghost event back into the list.

The `deleteEvent` handler avoids this correctly by checking `removed !== undefined` before re-adding. The same guard is needed here:

```ts
setAllEvents((current) => {
  const stillPresent = current.some((e) => e.id === eventId);
  if (!stillPresent) return current; // refresh already removed it — don't resurrect
  return current.map((e) => (e.id === eventId ? original : e));
});
```

---

src/components/calendar/EventDetailModal.tsx
```diff
@@ -99,26 +151,301 @@ function formatDateLabel(event: IEvent): string {
...
+
+function EventEditForm({ event, onSave, onCancel }: EditFormProps) {
+  const titleId = useId();
+  const startId = useId();
+  const endId = useId();
+  const descriptionId = useId();
+  const errorId = useId();
+
+  const [state, setState] = useState<EditFormState>(() =>
+    buildInitialEditState(event)
```
BenSeymourODB, line 198: **BLOCKER — `EventEditForm` form state goes stale when the parent's `event` prop changes mid-edit**

`buildInitialEditState(event)` is only called once (it's the lazy initialiser for `useState`). If the parent receives an optimistic update — for example, a concurrent `refreshEvents()` modifies the event's start time while the user is editing — the form continues to show the old values. Worse, if the parent's optimistic update completely replaces the `event` object reference (as `setAllEvents` does), the form will submit an edit based on stale data, silently overwriting the refresh.

The fix is to add a `useEffect` (or a `key` on the form component) so the form reseeds when `event.id` changes:

```tsx
// Simplest approach: give EventEditForm a key tied to the event id + an
// epoch so cancel → re-open also resets the state.
<EventEditForm
  key={event.id}
  event={event}
  onSave={handleSaveEdit}
  onCancel={() => setIsEditing(false)}
/>
```

This also covers the edge case where the parent swaps to a *different* event (e.g. keyboard navigation while the modal is open), which is unlikely today but the current code would serve the wrong form data for the new event.

---

src/components/calendar/EventDetailModal.tsx
```diff
@@ -136,6 +463,21 @@ export function EventDetailModal({
     }
   };
 
+  const handleSaveEdit = async (patch: EventEditPatch) => {
+    if (!onEdit) return;
```
BenSeymourODB: **BLOCKER — `handleSaveEdit` silently hangs when `onEdit` becomes undefined mid-edit**

If `onEdit` is defined when the user opens the form but becomes `undefined` before they click Save (e.g. `accessRole` flips to `reader` from a re-render), the `if (!onEdit) return;` guard on line 467 exits without re-throwing, without clearing `isSaving`, and without calling `onCancel`. The form is now stuck in a saving-disabled state with no escape — the Save button is `disabled` (because `isSaving` is still `false` — wait, actually `setIsSaving(true)` hasn't been called yet since it's in `EventEditForm.handleSubmit`), but actually the issue is more subtle: the form's `canSubmit` check blocks the submit, so `setIsSaving(true)` never runs. However, `handleSaveEdit` is the *outer* async handler; `onEdit` being undefined here means the save silently resolves without doing anything, and `EventEditForm.handleSubmit`'s `await onSave(patch)` resolves cleanly, so `setIsSaving` goes back to `false` and the form stays open, showing no error. The user would have no idea their edit was dropped.

Consider throwing an error so the form's catch path surfaces it, or at minimum asserting the invariant:

```ts
const handleSaveEdit = async (patch: EventEditPatch) => {
  if (!onEdit) {
    // This should never happen (canEdit is false when onEdit is absent).
    // If it does, re-throw so the form's catch shows the error.
    throw new Error("Edit is no longer available. Please close and retry.");
  }
  await onEdit(event, patch);
  setIsEditing(false);
  onClose();
};
```

---

src/components/calendar/AnalogClockView.tsx
```diff
@@ -119,6 +121,7 @@ export function AnalogClockView() {
         use24HourFormat={use24HourFormat}
         returnFocusTo={triggerRef}
         onDelete={handleDelete}
+        onEdit={handleEdit}
```
BenSeymourODB, line 198: **BLOCKER — `AnalogClockView` passes `onEdit` but never passes `accessRole`, so read-only calendar events show a live Edit (and Delete) button**

`AgendaCalendar` and `SimpleCalendar` both compute `getAccessRole(selectedEvent.calendarId)` and pass it as `accessRole`. `AnalogClockView` omits this prop entirely. The modal's `canEdit` / `canDelete` derive from `accessRole === undefined` being treated as "writable" (intentional for the not-yet-loaded case), so events on shared read-only calendars will show both the Edit and Delete buttons in this view — and clicking Edit will optimistically mutate the local list before receiving a 403 from Google.

This is a pre-existing gap for `onDelete` too (introduced in #197), but this PR widens the surface. Fix: add the same `getAccessRole` call that the other two views use:

```tsx
const { events, use24HourFormat, getAccessRole } = useCalendar();
// …
<EventDetailModal
  …
  onEdit={handleEdit}
  accessRole={selectedEvent ? getAccessRole(selectedEvent.calendarId) : undefined}
/>
```

---

src/lib/calendar/event-body.ts
```diff
@@ -0,0 +1,247 @@
...
+        error: "startDate must be a YYYY-MM-DD string for all-day events",
+      };
+    }
+    if (!DATE_ONLY_RE.test(raw.endDate)) {
+      return {
+        ok: false,
+        error: "endDate must be a YYYY-MM-DD string for all-day events",
+      };
+    }
+    if (raw.endDate < raw.startDate) {
```
BenSeymourODB, line 157: **QUESTION — all-day same-day event is incorrectly rejected as "end before start"**

The all-day comparison is a string lexicographic comparison. For a single-day all-day event the user would supply `startDate === endDate` (same day, inclusive). The test is `raw.endDate < raw.startDate`, which correctly passes when they're equal. So a single-day event is accepted — good.

However: the error message says "endDate must be *after* startDate", but same-day is actually valid. This was already the case in the original `validateCreateBody` and is preserved here, so it's not a regression. Worth fixing the error message to "endDate must be on or after startDate" to match the UI error text in `EventEditForm` ("End must be on or after start"). Currently the API error message and the UI error message disagree, which could confuse clients that surface the server error verbatim.

---

src/components/providers/MockCalendarProvider.tsx
```diff
@@ -195,6 +195,36 @@ export function MockCalendarProvider({
...
+          description: input.description,
+          color: input.color,
+          isAllDay: input.isAllDay,
+          startDate: input.startDate,
+          endDate: input.endDate,
+        };
+        return updated;
+      })
+    );
+    if (!updated) {
```
BenSeymourODB: **NIT — `updated` is read outside the `setState` updater callback, which is unsafe under React's batching / concurrent mode**

`updated` is assigned inside the `setAllEvents` callback but read synchronously after `setAllEvents(…)` returns. In React 18+ with concurrent rendering, the updater may be called more than once (React can discard and replay state updates), so `updated` could be set, then unset, then set again — but the read after the call sees whatever the last invocation assigned. In practice this mock is only used in tests and the timing doesn't affect correctness here, but it creates a misleading pattern.

A cleaner approach that mirrors the real provider: scan the current snapshot before setting state (acceptable in a mock because there are no real concurrent renders):

```ts
const editEvent = async (eventId: string, _calendarId: string, input: CreateEventInput): Promise<IEvent> => {
  // allEvents is readable here because this is a mock with no real concurrency.
  const existing = allEvents.find((e) => e.id === eventId);
  if (!existing) throw new Error(`Event ${eventId} not found in mock list`);
  const updated: IEvent = { ...existing, ...input };
  setAllEvents((prev) => prev.map((e) => (e.id === eventId ? updated : e)));
  return updated;
};
```

---

src/hooks/useEventEdit.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+import type { EventEditPatch } from "@/components/calendar/EventDetailModal";
+import { useCalendar } from "@/components/providers/CalendarProvider";
+import type { IEvent } from "@/types/calendar";
+import { toast } from "sonner";
+
+export function useEventEdit() {
+  const { editEvent } = useCalendar();
+
+  return async (event: IEvent, patch: EventEditPatch) => {
```
BenSeymourODB, line 24: **NIT — the returned handler is a new function reference every render, which will cause unnecessary re-renders in any component that uses it as a prop**

`useEventEdit` returns a plain `async (event, patch) => { … }` closure created fresh on every render. The React Compiler should handle this automatically if it can prove the dependencies are stable, but `editEvent` from `useCalendar()` itself is a new function on every render (it's declared inside `CalendarProvider` without any stable reference). This mirrors `useEventDelete`, so it's a pre-existing pattern and not a regression — but since both hooks are new-ish, documenting that the React Compiler is expected to memoize the result (or adding a comment to that effect) would help future readers understand why there's no explicit `useCallback`.

---

src/lib/calendar/date-input.ts
```diff
@@ -0,0 +1,50 @@
...
+    0,
+    0,
+    0,
+    0
+  );
+}
+
+/** Parse a `YYYY-MM-DDTHH:mm` string as a local `Date`. */
+export function parseDateTimeLocal(value: string): Date | null {
+  // `new Date(value)` interprets `YYYY-MM-DDTHH:mm` as local time, which is
```
BenSeymourODB, line 46: **QUESTION — `parseDateTimeLocal` relies on `new Date(value)` treating `YYYY-MM-DDTHH:mm` as local time, but this is spec-undefined for the no-seconds form**

The comment says "new Date interprets `YYYY-MM-DDTHH:mm` as local time, which is what we want". That is correct per the HTML spec for the `datetime-local` input value format and is consistent with current V8/SpiderMonkey behaviour. However, the ECMAScript spec (ECMA-262) only defines local-time interpretation for the full ISO-8601 forms *without* a `Z` or offset suffix when they include a time component — the exact behaviour for the abbreviated `YYYY-MM-DDTHH:mm` (no seconds, no timezone) has historically varied across engines. All major modern browsers do interpret it as local time, so this is fine in practice. A brief `// Verified: V8/Safari both treat the no-seconds HH:mm form as local time` comment would silence future lint-style doubts without code change.

---

BenSeymourODB (commented):
**Review: event detail edit (PATCH) — #265**

Overall this is a well-structured PR with solid test coverage. The validator extraction is clean and the POST tests all pass untouched. Four blockers need to be fixed before merge; two questions are informational.

**Blockers (4):**

1. **`buildGoogleEventBody` doesn't null out the opposite date field on type switches** — when patching a previously-all-day event to timed (or vice-versa), Google's PATCH merges the body and ends up with both `date` and `dateTime` set, which it rejects with a 400. Fix: explicitly set the unused field to `undefined` in the returned body object. See comment on `src/lib/calendar/event-body.ts`.

2. **Rollback resurrects refresh-deleted events** — the functional updater in the catch path unconditionally maps over `current`, so if `refreshEvents()` removed the event while the PATCH was in-flight the rollback silently re-inserts a ghost row. See comment on `CalendarProvider.tsx`.

3. **`EventEditForm` form state goes stale on concurrent `event` prop changes** — `useState(() => buildInitialEditState(event))` only runs once; a concurrent optimistic update changes the parent's `event` reference but the form keeps the old values. Add a `key` prop on `<EventEditForm>` tied to `event.id`. See comment on `EventDetailModal.tsx`.

4. **`AnalogClockView` never passes `accessRole`** — read-only calendar events show live Edit (and Delete) buttons in the analog-clock view. Pre-existing gap for delete, but this PR widens the surface. Add the same `getAccessRole` call that `AgendaCalendar` and `SimpleCalendar` use.

**Questions (2):**

- All-day API error message says "endDate must be *after* startDate" but same-day events are valid and the UI says "on or after". Inconsistency in `event-body.ts`.
- `parseDateTimeLocal` relies on `new Date("YYYY-MM-DDTHH:mm")` being local time — correct in all current engines but worth a brief comment since the no-seconds form is not in the ECMAScript spec's explicit list.

---

src/lib/calendar/event-body.ts
BenSeymourODB: Fixed in 60c6900. `buildGoogleEventBody` now emits explicit `null` for the unused `date`/`dateTime` sibling, and the `GoogleEventTime` interface widens those fields to `string | null` so the cast isn't needed. Two regression tests added in `events/[id]/__tests__/route.test.ts`: timed→all-day and all-day→timed PATCH type-switches both assert the unused sibling serialises as `null`. `null` is a no-op for INSERT (no prior value to clear) so POST is safe.

---

src/components/providers/CalendarProvider.tsx
BenSeymourODB: Fixed in 60c6900. The rollback updater now reads `current.some((e) => e.id === eventId)` first and returns the current snapshot unchanged when the event has already been removed by a concurrent refresh. Same `stillPresent` shape as the `removed !== undefined` guard in `deleteEvent`.

---

src/components/calendar/EventDetailModal.tsx
BenSeymourODB: Fixed in 60c6900. Added `key={event.id}` on `<EventEditForm>` so a parent re-render that swaps to a different event remounts the form and re-runs `buildInitialEditState`. Inline comment on the call site explains the React-identity reasoning.

---

src/components/calendar/EventDetailModal.tsx
BenSeymourODB: Fixed in 60c6900. `handleSaveEdit` now throws when `onEdit` is undefined — the form's `handleSubmit` catch path catches it, resets `isSaving`, and the user can hit Cancel to escape (or Save again, which will throw again). Without the throw, `await onSave(...)` resolved cleanly, `isSaving` stayed true, and the form was wedged. Inline comment explains the invariant and the reason for the throw.

---

src/components/calendar/AnalogClockView.tsx
BenSeymourODB: Fixed in 60c6900. `AnalogClockView` now pulls `getAccessRole` from the context and forwards `accessRole={selectedEvent ? getAccessRole(selectedEvent.calendarId) : undefined}` to the modal, matching `SimpleCalendar` and `AgendaCalendar`. Closes the pre-existing gap for `onDelete` here too.

---

src/lib/calendar/event-body.ts
BenSeymourODB: Fixed in 60c6900. Error message now reads "endDate must be on or after startDate" so it matches the modal's "End must be on or after start". The condition (`endDate < startDate`) still correctly accepts same-day events.

---

src/components/providers/MockCalendarProvider.tsx
BenSeymourODB: Fixed in 60c6900. `editEvent` in the mock now reads `allEvents` from the closure first (existence check + capture), then issues the setState updater for the merged row. Comment on the function documents that the closure read is safe for the mock specifically because it has no real concurrency.

---

src/hooks/useEventEdit.ts
BenSeymourODB: Inline comment added in 11eb760 — calls out that the React Compiler memoises the returned closure based on captured deps, that `CLAUDE.md` forbids manual `useCallback`, and that this mirrors the existing `useEventDelete` pattern.

---

src/lib/calendar/date-input.ts
BenSeymourODB: Reassuring comment added in 11eb760 noting that V8, JavaScriptCore, and SpiderMonkey all treat the no-seconds `YYYY-MM-DDTHH:mm` form as local time, even though ECMA-262 doesn't strictly require it. No behaviour change.

---

BenSeymourODB (commented):
## Independent review pass — LGTM with one regression fix pushed

Re-reviewed the full diff against the 9 prior review threads to verify each fix actually landed, then read the rest of the surface for anything missed. One new regression found and fixed in `ed9c28d`; everything else verified.

(Can't formally APPROVE since GitHub doesn't allow approving your own PR — treat this comment as a green light from the implement-issue review pass.)

### Prior review threads — all properly fixed

Walked the diff against each of the 9 threads on commit `ed68d7d` and confirmed each fix is present and correct:

| Thread | Fix verified |
|---|---|
| **Blocker — PATCH type-switch leaves both `date` + `dateTime` set** | `event-body.ts:265-274` emits explicit `null` for the unused sibling. Two regression tests in `events/[id]/__tests__/route.test.ts` exercise timed↔all-day. ✅ |
| **Blocker — rollback resurrects refresh-deleted events** | `CalendarProvider.tsx:939-943` `stillPresent` guard mirrors `deleteEvent`'s `removed !== undefined`. ✅ |
| **Blocker — `EventEditForm` form state stale on `event` prop change** | `key={event.id}` on `<EventEditForm>` at `EventDetailModal.tsx:527`. ✅ |
| **Blocker — `handleSaveEdit` silently hangs when `onEdit` becomes undefined** | Throws so the form's catch resets `isSaving` (`EventDetailModal.tsx:471-473`). ✅ |
| **Blocker — `AnalogClockView` missing `accessRole`** | Pulls `getAccessRole` from context and forwards (`AnalogClockView.tsx:63 + 199`). Closes the pre-existing delete-button gap from #197 too. ✅ |
| **Q — same-day error message** | "endDate must be on or after startDate" aligned with the modal text. ✅ |
| **Q — datetime-local engine support** | Reassuring comment in `date-input.ts`. ✅ |
| **Nit — Mock `setState` read outside updater** | `MockCalendarProvider.editEvent` reads `allEvents` from closure (mock-only safe), then issues the updater. ✅ |
| **Nit — `useEventEdit` memoization** | Inline comment in `useEventEdit.ts` cites the React Compiler + CLAUDE.md ban on manual `useCallback`. ✅ |

### New regression found and fixed

`EventDetailModal` keeps `isEditing` (new in this PR) and `confirmingDelete` (pre-existing from #197) state across modal close→reopen cycles, because the three parent surfaces (`SimpleCalendar`, `AgendaCalendar`, `AnalogClockView`) keep the modal permanently mounted and flip `event` between `null` (closed) and an `IEvent` (open). The component's `if (!event) return null;` short-circuits rendering but does NOT unmount, so React's state hooks survive. Re-opening the modal for a different event would land the user mid-edit (or in the delete confirmation) without ever clicking the trigger.

Reproduced with a component test that:
1. Opens event A, clicks Edit (sets `isEditing=true`)
2. Re-renders with `event={null}` (parent close)
3. Re-renders with `event=eventB` (parent opens a different event)
4. Asserts: no edit form visible; Edit button visible; heading shows Event B's title

Test fails against `83ecab9` (pre-fix). Fix in `ed9c28d`:

```tsx
useEffect(() => {
  setIsEditing(false);
  setConfirmingDelete(false);
}, [event?.id]);
```

This also resets `confirmingDelete` (the analogous pre-existing leak from #197) — same root cause, same one-line fix, so I bundled them.

### Deferred work — tracked

PR body lists `AgendaList` `onEdit`/`onDelete` wiring as out-of-scope but didn't link an issue. Filed **#451** so the day/week agenda-mode surface gap doesn't get lost.

### Non-blocking observations

- `CalendarProvider.editEvent:859` reads `allEvents` from the render closure to capture `original`. Theoretical staleness if `refreshEvents()` had just replaced this row — we'd rollback to the OLD version, overwriting the refresh. The inline comment acknowledges the trade-off (functional updater can't gate on existence). Same shape as `deleteEvent`'s pattern. Not a blocker.
- `editEvent`'s PATCH body includes `calendarId: input.calendarId` (line 898) but the route ignores body-`calendarId` per its contract. Harmless redundancy; matches the POST shape so the routes share a wire format.

### Quality gates (local)

- `pnpm test` — **2645 / 2645 passing** (one more than the PR body's 2361 because the suite has grown via merges)
- `pnpm lint:fix` — 0 errors, 5 pre-existing warnings (unchanged by this PR)
- `pnpm check-types` — clean
- `pnpm format:fix` — clean

CI ran green on `83ecab9` (commit before my push); `ed9c28d` will re-run and should stay green.

---

LGTM.

---

PR #394, opened 2026-06-09
chore(tasks): in-flight guard for TaskItem.handleToggle (#384)

BenSeymourODB:
## Summary

- Adds a `useRef(false)` in-flight guard to `TaskItem.handleToggle` so a rapid second click while the first is still pending early-returns instead of firing a second `onToggle` + `awardPoints`.
- The guard is cleared in a `finally` block — an error in either the toggle or the award path can never permanently lock the item.
- Two new component tests pin the behaviour: a held-pending double-click verifies single-flight; sequential clicks verify the guard releases between calls.

## Plan

No standalone plan file — the issue body (#384) contains the full spec and acceptance criteria, including the exact `useRef` + try/finally shape.

## Why this is worth doing even though the server-side index exists

The unique `(profileId, taskId, reason)` index already prevents the double-credit (the second `POST /api/points/award` returns `alreadyAwarded: true` and the existing test asserts the animation is suppressed). This change is purely additive client-side defense:

- One avoided network round-trip per double-click.
- No small window where two competing optimistic-UI states could flicker.
- The component's contract is now explicit — it no longer relies on an invariant that lives in `/api/points/award`.

## Acceptance criteria

- [x] `handleToggle` early-returns while a prior invocation is in flight
- [x] The guard is cleared in a `finally` so an error path doesn't permanently lock the item
- [x] New test: simulated double-click → `onToggle` and `awardPoints` each called exactly once
- [x] All existing TaskItem tests still pass

## Test plan

- [x] `pnpm vitest run src/components/tasks/__tests__/task-item.test.tsx` — 27/27 (was 25; +2 new)
- [x] `pnpm test:run` — 2476/2476 across 153 files
- [x] `pnpm lint:fix` — clean (6 pre-existing warnings on unrelated files)

Closes #384

---
_Generated by [Claude Code](https://claude.ai/code/session_01QrgTcP41xCyzFz8X3uiumB)_

### DIFF
--- a/src/components/tasks/__tests__/task-item.test.tsx
+++ b/src/components/tasks/__tests__/task-item.test.tsx
@@ -2,13 +2,7 @@
  * Tests for TaskItem component
  */
 import { type ReactNode } from "react";
-import {
-  act,
-  fireEvent,
-  render,
-  screen,
-  waitFor,
-} from "@testing-library/react";
+import { act, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { PointsProvider } from "../../rewards/points-context";
@@ -569,7 +563,8 @@ describe("TaskItem", () => {
               showPointsOnCompletion: true,
             }),
         })
-        // POST /api/points/award (should fire exactly once)
+        // POST /api/points/award — should fire exactly once even though
+        // both clicks reach the checkbox.
         .mockResolvedValueOnce({
           ok: true,
           json: () =>
@@ -587,6 +582,7 @@ describe("TaskItem", () => {
         releaseToggle = resolve;
       });
       const onToggle = vi.fn().mockReturnValue(togglePromise);
+      const user = userEvent.setup();
 
       render(
         <TaskItem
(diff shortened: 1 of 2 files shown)


BenSeymourODB (commented):
First-pass review by a Sonnet subagent (per `/implement-issue` step 10). Three findings worth addressing — surfaced here so the audit trail lives on the PR; addressing them in the next push.

---

**[BLOCKING] `fireEvent.click` on Radix Checkbox may be a false-positive vector** — `src/components/tasks/__tests__/task-item.test.tsx:607-611`

Concern: `fireEvent.click` dispatches a bare synthetic click; Radix UI's checkbox is a button-role element driven by its own press/release state machine. In jsdom (no pointer events), the bare click *might* not route through `onCheckedChange` at all — in which case both `fireEvent.click` calls are no-ops, `onToggle` is called 0 times, and the `toHaveBeenCalledTimes(1)` assertion passes for the wrong reason (or fails for the wrong reason). A version of `handleToggle` with **no guard** would also satisfy the assertion at that synchronous checkpoint because the held-pending promise hasn't resolved.

Counter-evidence: the test failed with **"got 2"** in the RED phase before the guard was added, so empirically `fireEvent.click(checkbox)` *does* fire `onCheckedChange` twice on this shadcn Checkbox. Still, the more defensive pattern is two parallel `user.click(checkbox)` invocations under `Promise.all`, which routes through Radix's full pointer-event sequence and removes the theoretical false-positive. Switching to that.

---

**[CONCERN] Redundant call-count assertion after `waitFor`** — `src/components/tasks/__tests__/task-item.test.tsx:625`

The trailing `expect(onToggle).toHaveBeenCalledTimes(1)` after the `waitFor` adds no signal beyond the synchronous assertion on line 611. Once `releaseToggle()` runs, the guard releases on the first invocation's `finally`; no further work in the test fires `onToggle`. Removing it.

---

**[CONCERN] Sequential-click test inadvertently exercises the award error path** — `src/components/tasks/__tests__/task-item.test.tsx:628-680`

The second click renders against the same `status: "needsAction"` (the `onToggle` mock doesn't mutate the prop), so it tries to award points a second time. Only one `mockResolvedValueOnce` for `/api/points/award` is queued, so the second award fetch returns `undefined` and `awardPoints` throws — caught silently by the inner try/catch inside `handleToggle`. The test passes for the right top-level reason (`onToggle` called twice ⇒ guard released) but the award path is broken in a way the test doesn't intend to exercise.

Fixing by dropping the `PointsProvider` wrapper for this test entirely — the guard-release contract is purely about `handleToggle`'s `finally`; the award path is irrelevant.

---

**[NIT] Comment redundancy** — `src/components/tasks/task-item.tsx:48-49`

The "Cleared in `finally` so an error path cannot permanently lock the item" sentence restates what the `finally` block three lines down already shows. Trimming to keep only the WHY (server-side guard exists; client guard avoids round-trip + flicker).

---

Merge-readiness after fixes: should be **READY**.

---

PR #409, opened 2026-06-16
chore(theme): tighten dark variant to respect ThemeScope (#324)

BenSeymourODB:
## Summary

- Tighten the `@custom-variant dark` selector in `src/app/globals.css` so the variant does NOT fire on descendants of `[data-theme-scope="light"]`, and DOES fire on descendants of `[data-theme-scope="dark"]` (symmetric with the token blocks below it).
- Use `:not(:where(...))` so the variant keeps the same zero-bumped specificity as before.
- Centralizes the leakage fix — shadcn ships `dark:` overrides on ~19 components in `src/components/ui/`, which `pnpm bump-ui` overwrites; a single CSS rule survives that.

## Why not the per-component audit

`CLAUDE.md` carves out `src/components/ui/**` from lint rules because `pnpm bump-ui` overwrites the directory. Any per-component cleanup (e.g. swapping `dark:bg-input/30` for semantic tokens in Button's outline variant) would be reverted on the next upgrade. The issue body explicitly lists "tighten the `dark` custom variant" as the Alternative; we take it.

## Plan

Linked plan: `.claude/plans/theme-scope-dark-variant-tightening-324.md`

## TDD notes

The new test cases in `src/components/theme/__tests__/theme-scope.test.tsx` **read the variant selector out of `globals.css` at test time** (paren-balanced parser) and inject it as a CSS rule that flags matching elements with a sentinel `--dark-variant-active: 1`. The probe rule is therefore coupled to the production source: the red phase failed 4/5 new cases before `globals.css` was tightened, and the green phase passed all 10 after.

Matrix covered:

| Outer | Scope | dark variant active? |
|---|---|---|
| `.dark` | none | yes |
| `.dark` | `light` | **no** (was yes — leakage fixed) |
| (none) | `dark` | **yes** (was no — symmetric) |
| `.wall-projector` | `light` | **no** (was yes — leakage fixed) |
| (none) | `dark` (element-of) | **yes** |

## Test plan

- [x] `pnpm test:run src/components/theme/__tests__/theme-scope.test.tsx` — 10/10 pass
- [x] `pnpm test:run` — 2479/2479 pass across 153 files
- [x] `pnpm lint:fix` — 0 errors (6 pre-existing warnings in unrelated files)

## Known limitation

Re-nested scopes (`light → dark → light`, etc.) are not handled by the `:not()` exclusion — the outermost light scope always wins. CSS custom-property inheritance does re-nest correctly for the token blocks themselves, so semantic-token utilities stay correct; only `dark:` utilities are stuck off inside an innermost re-entered dark island. No current consumer re-nests. Documented in the updated JSDoc.

## Acceptance criteria

- [x] Components in `src/components/ui/` audited (audit log in `docs/styling.md`); per-component cleanup is unnecessary because the variant change neutralises leakage centrally.
- [x] `dark:` leakage caveat in `theme-scope.tsx` JSDoc and `globals.css` variant comment updated to point at the audited list.
- [x] No regressions — existing 5 ThemeScope tests still pass alongside the 5 new cases.

Closes #324

---
_Generated by [Claude Code](https://claude.ai/code/session_01Ds2hp6fGF8RUFXSkYd4RrP)_

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -12,16 +12,19 @@
  * Why: `ThemeScope` (src/components/theme/theme-scope.tsx) re-declares all
  * shadcn/ui semantic tokens within `[data-theme-scope]`, but it cannot by
  * itself reverse the `dark:` variant on Tailwind utilities. shadcn ships
- * `dark:` overrides on ~19 components in `src/components/ui/` (e.g. Button's
+ * `dark:` overrides on 19 components in `src/components/ui/` (e.g. Button's
  * outline variant, Switch thumb, Badge destructive). Tightening the variant
  * here neutralizes the leakage centrally — durable across `pnpm bump-ui`,
- * which overwrites `src/components/ui/`. Issue: #324.
+ * which overwrites `src/components/ui/`. The audit list lives in
+ * `docs/styling.md`; re-run `grep -rln 'dark:' src/components/ui/` after
+ * `pnpm bump-ui` to refresh it. Issue: #324.
  *
  * Limitation: re-nesting (light scope inside dark inside light) is not
- * handled — the outermost light scope always wins the `:not()` exclusion.
- * The token blocks themselves DO re-nest via CSS custom-property inheritance,
- * but `dark:` utilities will be stuck off inside the innermost re-entered
- * dark island. No current consumer re-nests; documented in
+ * handled — any `[data-theme-scope="light"]` ancestor anywhere in the tree
+ * suppresses `dark:` for all of its descendants, including a re-entered
+ * dark island nested inside it. The token blocks themselves DO re-nest via
+ * CSS custom-property inheritance, but `dark:` utilities are stuck off in
+ * the re-entered dark island. No current consumer re-nests; documented in
  * `src/components/theme/theme-scope.tsx`.
  *
  * NB: Tailwind requires `@custom-variant` on a single line — do not break it.
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -2,18 +2,32 @@
 @import "tw-animate-css";
 
 /*
- * `dark` Tailwind variant — also activates under the Wall-Projector theme,
- * which inherits the dark token set by default. Components nested inside a
- * `[data-theme-scope="light"]` subtree should use semantic tokens
- * (bg-background, text-foreground, etc.) rather than `dark:` overrides; the
- * scope re-declares the tokens but cannot reverse the `dark:` variant on
- * descendants. The repo-wide audit of components that use `dark:` overrides
- * and may end up inside a ThemeScope island is tracked in #324.
+ * `dark` Tailwind variant — fires for descendants of `.dark` /
+ * `.wall-projector` (the dark-by-default themes) AND for descendants of a
+ * `[data-theme-scope="dark"]` island (symmetric with the token blocks below),
+ * but NOT for elements that sit inside `[data-theme-scope="light"]`. The
+ * `:not(:where(...))` suffix uses zero-specificity `:where()` so the variant
+ * keeps the same specificity weight as the original two-class form.
+ *
+ * Why: `ThemeScope` (src/components/theme/theme-scope.tsx) re-declares all
+ * shadcn/ui semantic tokens within `[data-theme-scope]`, but it cannot by
+ * itself reverse the `dark:` variant on Tailwind utilities. shadcn ships
+ * `dark:` overrides on ~19 components in `src/components/ui/` (e.g. Button's
+ * outline variant, Switch thumb, Badge destructive). Tightening the variant
+ * here neutralizes the leakage centrally — durable across `pnpm bump-ui`,
+ * which overwrites `src/components/ui/`. Issue: #324.
+ *
+ * Limitation: re-nesting (light scope inside dark inside light) is not
+ * handled — the outermost light scope always wins the `:not()` exclusion.
+ * The token blocks themselves DO re-nest via CSS custom-property inheritance,
+ * but `dark:` utilities will be stuck off inside the innermost re-entered
+ * dark island. No current consumer re-nests; documented in
+ * `src/components/theme/theme-scope.tsx`.
  *
  * NB: Tailwind requires `@custom-variant` on a single line — do not break it.
  */
 /* prettier-ignore */
-@custom-variant dark (&:where(.dark, .dark *, .wall-projector, .wall-projector *));
+@custom-variant dark (&:where([data-theme-scope="dark"], [data-theme-scope="dark"] *, .dark, .dark *, .wall-projector, .wall-projector *):not(:where([data-theme-scope="light"], [data-theme-scope="light"] *)));
 
 /* shadcn/ui semantic color tokens — light theme (slate base) */
 :root {
--- a/src/components/theme/__tests__/theme-scope.test.tsx
+++ b/src/components/theme/__tests__/theme-scope.test.tsx
@@ -9,8 +9,60 @@
  * production pages rely on the rules in `globals.css`.
  */
 import { render } from "@testing-library/react";
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
 import { describe, expect, it } from "vitest";
 import { ThemeScope } from "../theme-scope";
+
+const GLOBALS_CSS_PATH = join(
+  __dirname,
+  "..",
+  "..",
+  "..",
+  "app",
+  "globals.css"
+);
+
+/**
+ * Read the `@custom-variant dark (...)` selector from globals.css and turn it
+ * into a CSS rule that flags matched elements with `--dark-variant-active: 1`.
+ * The test then asserts presence/absence of that flag in cross-scope scenarios.
+ * Pulling the selector from the source-of-truth file (not duplicating it in
+ * the test) is what gives this a real red→green cycle when globals.css
+ * changes.
+ */
+function darkVariantProbeRule(): string {
+  const css = readFileSync(GLOBALS_CSS_PATH, "utf8");
+  // Directive form: `@custom-variant dark (<balanced-parens-selector>);`.
+  // Find the opening `(` then walk paren-balanced to the matching `)`.
+  const start = css.search(/@custom-variant\s+dark\s+\(/);
+  if (start === -1) {
+    throw new Error("Could not locate @custom-variant dark in globals.css");
+  }
+  const open = css.indexOf("(", start);
+  let depth = 1;
+  let end = -1;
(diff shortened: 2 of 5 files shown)


src/app/globals.css
```diff
@@ -2,18 +2,32 @@
...
+ * The token blocks themselves DO re-nest via CSS custom-property inheritance,
+ * but `dark:` utilities will be stuck off inside the innermost re-entered
+ * dark island. No current consumer re-nests; documented in
+ * `src/components/theme/theme-scope.tsx`.
  *
  * NB: Tailwind requires `@custom-variant` on a single line — do not break it.
  */
 /* prettier-ignore */
-@custom-variant dark (&:where(.dark, .dark *, .wall-projector, .wall-projector *));
+@custom-variant dark (&:where([data-theme-scope="dark"], [data-theme-scope="dark"] *, .dark, .dark *, .wall-projector, .wall-projector *):not(:where([data-theme-scope="light"], [data-theme-scope="light"] *)));
```
BenSeymourODB, line 33: **Minor doc inaccuracy: "~19 components" vs exactly 19**

The comment on line 15 says `~19 components in src/components/ui/`. Running `grep -rln 'dark:' src/components/ui/` returns exactly 19 files — not approximately 19. The `~` implies uncertainty; either remove the tilde or add a note that the count may change after `pnpm bump-ui`. The docs/styling.md audit list correctly lists all 19 without hedging, which is the right pattern.

Suggestion: `dark: overrides on 19 components in src/components/ui/ (as of this audit; re-run grep -rln 'dark:' src/components/ui/ after pnpm bump-ui)` or simply drop the `~`.

---

src/components/theme/__tests__/theme-scope.test.tsx
```diff
@@ -66,6 +118,91 @@ describe("ThemeScope", () => {
...
+      style.textContent = darkVariantProbeRule();
+      document.head.appendChild(style);
+      try {
+        return run();
+      } finally {
+        style.remove();
+      }
+    }
+
+    function probeActive(el: Element): boolean {
```
BenSeymourODB, line 142: **jsdom / nwsapi CSS custom-property probe: caveats worth documenting**

`getComputedStyle(el).getPropertyValue("--dark-variant-active")` relies on jsdom's `@asamuzakjp/nwsapi` (v2.3.9 as of this PR) correctly evaluating `:not(:where(...))` in the stylesheet. This works today — the five activation/suppression assertions all pass — but this is a non-obvious runtime dependency.

Two issues worth noting:

1. **`getComputedStyle` on custom properties in jsdom has historically been unreliable** for inherited/cascaded values (not for directly-matched properties). Here it works because the rule sets the property directly on the matched element, not via inheritance. The existing test comment on line ~207 acknowledges this for `--background`. The new tests don't carry the same caveat comment, even though they depend on the same jsdom limitation.

2. **The suppression tests (cases 2 and 4 — `toBe(false)`) could pass vacuously** if jsdom silently discarded the entire rule due to an unsupported selector. The activation tests (1, 3, 5 — `toBe(true)`) act as a live guard: if jsdom couldn't parse the selector, those would fail with `false !== true`. This implicit dependency is sound logic but is not explained in any comment. A brief note like:

   ```ts
   // Cases 1, 3, 5 (toBe(true)) act as a parser smoke-test: if jsdom drops
   // the :not(:where()) rule entirely, they fail, ensuring 2 and 4 are not vacuously true.
   ```

   …would make the test suite's integrity self-documenting.

---

src/components/theme/__tests__/theme-scope.test.tsx
```diff
@@ -66,6 +118,91 @@ describe("ThemeScope", () => {
...
+            <ThemeScope mode="light">
+              <span data-testid="probe">x</span>
+            </ThemeScope>
+          </div>
+        );
+        expect(probeActive(getByTestId("probe"))).toBe(false);
+      });
+    });
+
+    it("activates dark variant on the dark scope element itself", () => {
```
BenSeymourODB, line 200: **Good catch testing the scope element itself — consider adding `.dark` self-match for symmetry**

Test 5 correctly verifies that `[data-theme-scope="dark"]` itself (not just its children) activates `dark:`. That's the right instinct.

The plan spec (`.claude/plans/`) lists 5 test cases and doesn't call for this, so it's in scope not to add it. But for completeness: there's no test for `.dark` itself (the element that carries the class, not a child). Since `.dark` always matches `.dark` in the `:where()` positive arm and has no light-scope ancestor, it trivially matches — low risk. Raising it as a gap, not a blocker.

---

src/app/globals.css
```diff
@@ -2,18 +2,32 @@
...
+ * shadcn/ui semantic tokens within `[data-theme-scope]`, but it cannot by
+ * itself reverse the `dark:` variant on Tailwind utilities. shadcn ships
+ * `dark:` overrides on ~19 components in `src/components/ui/` (e.g. Button's
+ * outline variant, Switch thumb, Badge destructive). Tightening the variant
+ * here neutralizes the leakage centrally — durable across `pnpm bump-ui`,
+ * which overwrites `src/components/ui/`. Issue: #324.
+ *
+ * Limitation: re-nesting (light scope inside dark inside light) is not
+ * handled — the outermost light scope always wins the `:not()` exclusion.
+ * The token blocks themselves DO re-nest via CSS custom-property inheritance,
```
BenSeymourODB: **Re-nesting limitation comment is accurate but has one imprecise phrase**

Lines 20–25: "the outermost light scope always wins the `:not()` exclusion."

Technically `:not(:where([data-theme-scope="light"], [data-theme-scope="light"] *))` fires as soon as _any_ ancestor carries `[data-theme-scope="light"]`, not specifically the _outermost_ one. In a `dark → light → dark` nesting, the first (innermost from the element's perspective going up) light ancestor already triggers the exclusion — which happens to also be the outermost in the described nesting pattern. The wording is correct for the documented use case but could mislead if the reader has a more complex tree in mind.

Suggested rewording: "any `[data-theme-scope="light"]` ancestor anywhere in the tree will suppress `dark:` for all of its descendants, including a re-entered dark island nested inside it."

Non-blocking — the current wording is accurate for all current consumers.

---

BenSeymourODB (commented):
## Review: theme-scope dark variant tightening (#324)

Reviewed all 5 changed files. All 2479 tests pass. No blockers found — posting comments for transparency, not to block.

### Selector correctness ✓

Walked the full matrix manually and simulated the test's paren-balance parser in Python:

| Scenario | Expected | Verdict |
|---|---|---|
| `.dark` child, no scope | `dark:` ON | Correct |
| `.dark` child inside `[data-theme-scope="light"]` | `dark:` OFF | Correct |
| Child of `[data-theme-scope="dark"]`, no outer dark class | `dark:` ON | Correct |
| `.wall-projector` child inside `[data-theme-scope="light"]` | `dark:` OFF | Correct |
| `[data-theme-scope="dark"]` element itself | `dark:` ON | Correct |
| `[data-theme-scope="light"]` element itself (inside `.dark`) | `dark:` OFF | Correct |
| Re-nested `dark > light > dark > span` | `dark:` OFF (documented limitation) | Correct |

No sneaky misses. The `:not(:where())` correctly excludes an element that matches _any_ light-scope ancestor — it does not require the outermost ancestor to be the light scope, and the selector is evaluated per-element, not per-subtree.

### Specificity ✓

`:where()` contributes 0 specificity regardless of its argument list. `:not(:where(...))` takes the specificity of its argument (`:where()` → 0). So the full variant adds 0 to the generated utility selector — identical to the original two-class form. No cascade regression.

### Test rigor ✓ (with minor documentation gap)

The `darkVariantProbeRule()` approach (read selector from `globals.css` → inject as probe rule → check computed property) is genuinely coupled to the source file. A regression in `globals.css` would be caught. The paren-balance extractor is correct and handles the nested `:not(:where())` structure cleanly — verified by re-running the logic in Python. Test isolation is correct (`withProbeStylesheet` cleans up in a `finally` block; `@testing-library/react` `cleanup()` runs `afterEach`). 

The one genuine gap: the suppression tests (`toBe(false)`) depend implicitly on the activation tests (`toBe(true)`) as a parser smoke-test. This is sound reasoning but isn't explained — see inline comment.

### Docs accuracy ✓

`grep -rln 'dark:' src/components/ui/` returns exactly 19 files. The list in `docs/styling.md` matches precisely. The plan doc, JSDoc, and CSS comment are mutually consistent.

### CLAUDE.md compliance ✓

- No `useMemo`/`useCallback`/`React.memo` in any changed file.
- No `any` types.
- No non-standard Tailwind colors introduced.
- `src/components/ui/` not edited (variant fix is in `globals.css`, durable across `pnpm bump-ui`).
- TDD path: the old selector in `HEAD~1` is confirmed (`&:where(.dark, .dark *, .wall-projector, .wall-projector *)`) — the new tests would have failed red against that selector (light-scope suppression cases), then gone green with the new selector.

### Production impact ✓

The only current `ThemeScope` consumer is `AnalogClockView` (uses `mode="light"` inside `.wall-projector`). The fix specifically corrects that scenario — `dark:` utilities inside that scope will now correctly be suppressed. All existing tests continue to pass.

---

src/app/globals.css
BenSeymourODB: Fixed in 4d8c8ee — dropped the `~` and added a pointer to `docs/styling.md` + the refresh command (`grep -rln 'dark:' src/components/ui/` after `pnpm bump-ui`).

---

src/components/theme/__tests__/theme-scope.test.tsx
BenSeymourODB: Added in 4d8c8ee — a block comment on `withProbeStylesheet` now spells out (a) why direct-match custom-property lookup is reliable in jsdom even where its `var(...)` resolution historically isn't, and (b) that the `activates …` cases serve as a parser smoke-test so the `suppresses …` cases cannot pass vacuously.

---

src/components/theme/__tests__/theme-scope.test.tsx
BenSeymourODB: Added in 4d8c8ee — small self-match symmetry case for `.dark` itself, mirroring the existing `[data-theme-scope="dark"]` self-match. Cheap to add and locks in the positive arm.

---

src/app/globals.css
BenSeymourODB: Reworded in 4d8c8ee — the comment now reads "any `[data-theme-scope="light"]` ancestor anywhere in the tree suppresses `dark:` for all of its descendants, including a re-entered dark island nested inside it," which is precise for arbitrary nesting depth.

---

PR #410, opened 2026-06-16
fix(calendar): unify bare-date parsing across overlap helpers (#375)

BenSeymourODB:
## Summary

- Promote `parseEventStartLocal` from `YearCalendar.tsx` into shared `parseEventStart` / `parseEventEnd` in `src/lib/calendar-helpers.ts`.
- Wire the four `getEventsForX` overlap helpers (day/week/month/year) onto the shared parser so bare `YYYY-MM-DD` events are bucketed on the local calendar day, not UTC midnight.
- Eliminates the dot-vs-count asymmetry where the YearCalendar dot rendered on the correct day but the count silently undercounted bare-date events in negative-offset timezones.

## Bug surface

For a bare-date all-day event `{ startDate: "2026-01-01", endDate: "2026-01-01", isAllDay: true }` in `America/Los_Angeles` (UTC-8):

- **Before:** dot renders on Jan 1 (uses local `parseEventStartLocal`) but `getEventsForYear` returns `[]` because `parseISO("2026-01-01") → 2025-12-31T16:00 local`. Count and dot disagree.
- **After:** both paths share `parseEventStart`, dot and count agree.

The bug fired for the `MockCalendarProvider`, `/test/calendar` fixtures, and any non-Google event source. Google-sourced events were unaffected because `transformGoogleEvent` already appends `T00:00:00` to bare dates.

## Plan

Linked plan: `.claude/plans/bare-date-parsing-unification-375.md`.

## Test plan

- [x] Unit tests for `parseEventStart` / `parseEventEnd` covering bare-date and full-ISO inputs (incl. `Z`-suffixed UTC).
- [x] `getEventsForX` inclusion tests for bare-date events at Jan-1, Dec-31, first/last day of month, first/last day of week, and `getEventsForDay`.
- [x] `YearCalendar.test.tsx` regression: count and dot agree for `{ startDate: "2026-01-01", endDate: "2026-01-01" }`.
- [x] `pnpm test` — 2534/2534 passing.

## Out of scope

Per the issue, only the four overlap helpers were rewired. Other `parseISO` callers (`groupEvents`, `getEventsCount`, `computeEventColumns`, `assignBarRows`, `calculateMonthEventPositions`, `getMonthCellEvents`, `getEventTimePosition`) remain on `parseISO` — they don't drive the bare-date count/dot mismatch this issue targets, and broadening would invite scope creep.

Closes #375

---
_Generated by [Claude Code](https://claude.ai/code/session_01AfFYsuzVBE5GG2N1nyVjUd)_

### DIFF
--- a/.claude/plans/bare-date-parsing-unification-375.md
+++ b/.claude/plans/bare-date-parsing-unification-375.md
@@ -41,13 +41,15 @@ targeted fix.
      `process.env.TZ` manipulation, matching the PR #251 test style).
 
 2. `src/lib/calendar-helpers.ts`:
-   - Export a `BARE_DATE_RE` and two helpers:
+   - Export two helpers:
      ```ts
      export function parseEventStart(event: IEvent): Date;
      export function parseEventEnd(event: IEvent): Date;
      ```
-   - Each tries `BARE_DATE_RE` first, falls back to `parseISO` on the
-     respective field.
+   - Each tries the module-private `BARE_DATE_RE` first, falls back to
+     `parseISO` on the respective field. The regex stays private — its
+     only callers are `parseEventStart` / `parseEventEnd`, so it doesn't
+     belong in the public surface.
 3. Swap `parseISO(event.startDate)` / `parseISO(event.endDate)` for the
    new helpers inside `getEventsForDay`, `getEventsForWeek`,
    `getEventsForMonth`, `getEventsForYear`. Other helpers stay on
--- a/.claude/plans/bare-date-parsing-unification-375.md
+++ b/.claude/plans/bare-date-parsing-unification-375.md
@@ -1,0 +1,95 @@
+# Issue #375 — Unify bare-date parsing across calendar-helpers + YearCalendar
+
+## Goal
+
+Eliminate the dot-vs-count asymmetry for bare-date all-day events. PR #251
+introduced `parseEventStartLocal` inside `YearCalendar.tsx`, but the
+sibling `getEventsForX` overlap helpers in `src/lib/calendar-helpers.ts`
+still use `parseISO`, which parses bare `YYYY-MM-DD` as UTC midnight.
+Promote the local helper to a shared `parseEventStart` (+ symmetric
+`parseEventEnd`) and use it everywhere the overlap math needs it.
+
+## Failure mode (from the issue)
+
+Bare-date all-day event `{ startDate: "2026-01-01", endDate: "2026-01-01",
+isAllDay: true }` in a negative-offset zone (UTC-8): the YearCalendar dot
+renders on Jan 1 but `getEventsForYear` returns `[]` because
+`parseISO("2026-01-01") → 2025-12-31T16:00 local`. Count and dot disagree.
+
+## Scope
+
+Touches the four overlap helpers only. Other `parseISO` callers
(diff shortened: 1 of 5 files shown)


BenSeymourODB (commented):
## First-pass review — PR #410 (issue #375)

### Overall

The core fix is correct and well-scoped. All four overlap helpers are fully rewired, no straggling `parseISO(event.startDate/endDate)` remain inside the four target functions, the local `parseEventStartLocal` + `BARE_DATE_RE` are cleanly removed from `YearCalendar.tsx`, and the `bucketEventColorsByDayKey` test stays exercised. Out-of-scope helpers (`groupEvents`, `getEventsCount`, `computeEventColumns`, etc.) keep their existing `parseISO` calls — no scope creep. Test coverage is solid: parser semantics, Z-suffix sanity, all four helpers, Dec-31 boundary, end-of-week boundary, regression test.

Three issues worth addressing:

---

**1. `BARE_DATE_RE` not exported — plan deviation (nit / question)**

`src/lib/calendar-helpers.ts` line 38

The plan spec says:
> Export a `BARE_DATE_RE` and two helpers

`BARE_DATE_RE` is `const` (module-private), not `export const`. That's arguably *better* from an encapsulation standpoint (the regex is an implementation detail of `parseEventDateField`, callers should use the exported functions), but it deviates from the plan's stated contract. Either update the plan's acceptance checklist to remove the `BARE_DATE_RE` export requirement, or export it. Given the private helper `parseEventDateField` fully encapsulates it, I'd lean toward keeping it unexported and noting the plan was overly specific — no action needed if you agree, but worth a brief acknowledgement.

---

**2. `getEventsForDay` skips `isValid` guards (nit)**

`src/lib/calendar-helpers.ts` lines 384–396

`getEventsForWeek`, `getEventsForMonth`, and `getEventsForYear` all gate on `isValid(eventStart) && isValid(eventEnd)` before comparing. `getEventsForDay` does not — it calls `startOfDay(parseEventStart(event))` directly. This pre-dates this PR (pre-existing), but since this PR touches both branches of that function it's a good moment to note: if `event.startDate` or `event.endDate` is ever an empty string or otherwise unparseable, `getEventsForDay` will propagate an `Invalid Date` through `startOfDay` while the other three helpers quietly filter it out. Not a blocker for this PR since the fix is correct, but worth a follow-up issue if not already tracked.

---

**3. Regression test doesn't verify dot absence on the *wrong* day (nit)**

`src/components/calendar/__tests__/YearCalendar.test.tsx` (new test block, ~line 558)

The regression test checks that the Jan-1 cell shows a dot and the count reads `1 event`. It doesn't assert that the Dec-31 cell has *no* dot — which would have caught the original UTC-midnight slip. Adding `expect(within(dec31Cell).queryAllByTestId("year-calendar-dot")).toHaveLength(0)` would make the test fail if the old `parseISO` regression reappears. Not a blocker, but it would make the test a complete regression guard rather than a partial one.

---

### Acceptance criteria coverage

All six items from issue #375 are satisfied. Commit message format matches repo style. No manual memoization, no `any` types, standard Tailwind only (no UI changes). Single clean commit.

**Verdict: approve after the `BARE_DATE_RE` export question is clarified.** The other two are nits that could be addressed as follow-ups.

---

PR #415, opened 2026-06-16
fix(auth): validate Google token-refresh response shape at runtime (#405)

BenSeymourODB:
## Summary

- Define `GoogleRefreshedTokensSchema` (Zod v4) and validate the 200 body via `safeParse` before returning; the existing cast was silently letting malformed payloads through.
- On parse failure, throw `GoogleTokenRefreshError(status, { error: "malformed_token_response", issues })` — `classifyTokenRefreshError` treats the unknown error code as transient, so the singleflight slot purges and the next callback retries instead of forcing re-auth.
- Re-export `GoogleRefreshedTokens` as `z.infer<typeof GoogleRefreshedTokensSchema>` so the in-tree callers (`refresh-session-tokens.ts`, `session-callback.test.ts`) keep the same type without a duplicate declaration.

## Why

Without a runtime shape check, a Google 200 with a partial / proxy-rewritten body left `expires_in` as `undefined`. The orchestrator then computed `Math.floor(now/1000 + undefined) → NaN` and Prisma either wrote `NaN` (rejected) or silently coerced to `NULL` on the optional column, corrupting the refresh window. Either path is hard to diagnose. PR #401's singleflight raises the impact: every concurrent session callback for that account collapses onto the same broken refresh.

## Plan

No plan file existed under `.claude/plans/`; the issue body is the spec (5 acceptance criteria, all covered).

## Test plan

- [x] `pnpm test:run src/lib/auth/__tests__/refresh-google-token.test.ts` — 13 passed (6 new + 7 existing)
- [x] `pnpm test:run` — full suite, 2531 passed across 153 files
- [x] `pnpm lint:fix` — 0 errors (6 pre-existing warnings unrelated to this diff)

New test cases (issue #405 acceptance criteria):

- [x] Throws on 200 missing `expires_in`
- [x] Throws on 200 with `expires_in` as a string
- [x] Throws on 200 with `null` body
- [x] Throws on 200 with unrelated JSON value (array)
- [x] Throws on 200 with non-positive `expires_in` (defensive — zero would write an already-expired window)
- [x] Synthetic envelope tags `body.error === "malformed_token_response"` so `classifyTokenRefreshError` returns `"transient"`
- [x] Canonical 200 (only the required pair, no optional fields) still passes through

Closes #405

---
_Generated by [Claude Code](https://claude.ai/code/session_01RuS2DGGXmyvKK63KFtc17D)_

### DIFF
--- a/src/lib/auth/__tests__/refresh-google-token.test.ts
+++ b/src/lib/auth/__tests__/refresh-google-token.test.ts
@@ -10,6 +10,7 @@
  */
 import { jsonResponse } from "@/lib/test-utils/api-test-helpers";
 import { beforeEach, describe, expect, it, vi } from "vitest";
+import { classifyTokenRefreshError } from "../refresh-error-classifier";
 import {
   GoogleTokenRefreshError,
   refreshGoogleAccessToken,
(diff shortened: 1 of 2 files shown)


src/lib/auth/refresh-google-token.ts
```diff
@@ -67,11 +86,18 @@ export async function refreshGoogleAccessToken(
...
-    throw new GoogleTokenRefreshError(response.status, tokensOrError);
+    throw new GoogleTokenRefreshError(response.status, rawBody);
   }
 
-  return tokensOrError as GoogleRefreshedTokens;
+  const parsed = GoogleRefreshedTokensSchema.safeParse(rawBody);
+  if (!parsed.success) {
+    throw new GoogleTokenRefreshError(response.status, {
+      error: "malformed_token_response",
+      issues: parsed.error.issues,
```
BenSeymourODB: `parsed.error.issues` is embedded directly in the thrown error body, which eventually reaches `deps.logger.error(err, ...)` in `refreshGoogleSessionTokensIfNeeded`. Zod issues can carry a `received` field on type-mismatch issues (and an `input` field on the issue root in newer Zod v4 builds). For the current schema, the `min(1)` constraint on `access_token` would emit a `too_small` issue with `input: ""` — harmless. But if a future schema addition tightens a string field (e.g. format check on `scope`), token material could land in logs verbatim.

Defensive fix is one line — strip to `{ code, path, message }` before embedding:

```ts
issues: parsed.error.issues.map(({ code, path, message }) => ({ code, path, message })),
```

This preserves all diagnostic value while guaranteeing no raw field values reach the logger.

---

src/lib/auth/__tests__/refresh-google-token.test.ts
```diff
@@ -119,6 +120,95 @@ describe("refreshGoogleAccessToken", () => {
...
+      // A proxy rewriting the response to a string or array still parses as
+      // JSON but doesn't satisfy the token shape.
+      mockFetch.mockResolvedValue(jsonResponse([]));
+
+      await expect(
+        refreshGoogleAccessToken("rt", "cid", "sec")
+      ).rejects.toBeInstanceOf(GoogleTokenRefreshError);
+    });
+
+    it("throws GoogleTokenRefreshError when expires_in is non-positive", async () => {
```
BenSeymourODB: The test comment ("Treat zero / negative as malformed") names both, but only `expires_in: 0` is exercised. Negative integers are already rejected by `positive()`, but pinning that explicitly with a `expires_in: -1` sibling case closes the gap cheaply and matches the comment's stated intent.

---

BenSeymourODB (commented):
First-pass review by Sonnet subagent (via `/implement-issue`). Two items — one defensive log-hygiene fix worth taking, one nit for test coverage. Schema correctness, classifier routing, type compatibility, and CLAUDE.md compliance all check out. Addressing both in a follow-up commit.

---

src/lib/auth/refresh-google-token.ts
BenSeymourODB: Fixed in 096a64d. The `parsed.error.issues` mapping now strips to `{ code, path, message }` at the throw site, with an inline comment explaining the rationale so a future schema tightening doesn't re-introduce the leak.

---

src/lib/auth/__tests__/refresh-google-token.test.ts
BenSeymourODB: Fixed in 096a64d. Split the zero case into its own `it()` ("when expires_in is zero") and added a sibling `it()` for `expires_in: -1` that pins `positive()` directly.

---

PR #416, opened 2026-06-16
fix(settings): rollback only touched keys in updateTaskSettings (#413)

BenSeymourODB:
## Summary

- `SettingsForm.updateTaskSettings` snapshotted `previous = taskSettings` from the render closure and restored the whole object on PUT failure. With two overlapping PUTs to different task-settings keys, the failed one's rollback wiped the concurrent successful change.
- Apply PR #366's `updateSettings` pattern: snapshot only the keys being updated via the functional setter's current state, then rollback by merging that partial snapshot back into the truly-current state.
- Add a component regression test that fires two overlapping task-settings PUTs (toggle `showCompletedTasks` + change `taskSortOrder`), resolves the sort-order PUT first then rejects the toggle PUT, and asserts only the toggle reverts while the sort-order change is preserved.

## Plan

No `.claude/plans/` entry — the fix is a direct port of the #363/#366 pattern to the sibling `updateTaskSettings` function. Acceptance criteria from the issue body:

- [x] `updateTaskSettings` rolls back only the keys it attempted, sourced from the functional-setter current state (not a closure snapshot)
- [x] Component test mirroring the `"preserves a concurrent successful PUT"` spec — two overlapping task-settings PUTs (one fails, one succeeds), asserting the successful field's value is preserved
- [x] `pnpm test:run src/components/settings/__tests__/settings-form.test.tsx` passes

## Test plan

- [x] `pnpm test` — full suite (2,525 tests) passes
- [x] New test fails against the pre-fix code (whole-object rollback wipes the sort-order change) and passes against the partial-keys rollback

Closes #413

---
_Generated by [Claude Code](https://claude.ai/code/session_015LHPSX6SQLGGNpXaYSsziH)_

### DIFF
--- a/src/components/settings/__tests__/settings-form.test.tsx
+++ b/src/components/settings/__tests__/settings-form.test.tsx
@@ -754,3 +754,98 @@ describe("SettingsForm — same-key rollback race (#420)", () => {
     expect(busHandler).toHaveBeenCalledWith({ theme: "light" });
   });
 });
+
+/**
+ * Tests for SettingsForm.updateTaskSettings optimistic-update rollback
+ * semantics (#413).
+ *
+ * Mirrors the #363/#366 rollback contract for the profile-scoped task
+ * settings: the failure path must restore the pre-call value of *only*
+ * the keys being updated, read from the truly-current state via the
+ * functional setter — not a stale closure of the entire
+ * `taskSettings` object. Without this, an overlapping successful PUT to
+ * a different task-settings key loses its optimistic value when an
+ * earlier PUT fails.
+ */
+describe("SettingsForm — updateTaskSettings rollback (#413)", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+    mockActiveProfile(ACTIVE_PROFILE_ID);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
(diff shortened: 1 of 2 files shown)


BenSeymourODB (commented):
First-pass review (Sonnet subagent, verdict: APPROVE).

The fix is a clean, mechanical port of the `updateSettings` pattern (PR #366) to `updateTaskSettings`. The implementation is structurally identical to the reference: functional-setter snapshot of only the touched keys, functional-setter merge on rollback, `previousPartial` guarded by an `if (revert)` check. The new test exercises the exact failure scenario the bug describes (two overlapping PUTs to disjoint keys, second resolves before first rejects) and the final assertions distinguish buggy whole-object rollback from the correct partial rollback. The `emitUserSettingsChange` call that `updateSettings` makes on success is intentionally absent in `updateTaskSettings` (task settings are profile-scoped, not user-scoped), so that asymmetry is correct — #414 tracks the parallel bus-rollback emit for `updateSettings` separately.

One review nit was raised on the final test assertion (`expect(updatedCombobox).toBeDefined()` was a weak existence check that another "priority"-bearing element could satisfy by accident). Addressed in 741f6a9 by asserting against the original combobox node identity with `toHaveTextContent(/priority/i)` and `.not.toHaveTextContent(/due date/i)`.

---

PR #417, opened 2026-06-16
fix(a11y): suppress Webkit native cancel button on type=search inputs (#411)

BenSeymourODB:
## Summary

- Both `AgendaCalendar` and `AgendaList` render their search field as `<input type="search">` paired with a custom clear `<button>`. On Webkit/Blink (Chrome, Edge, Safari) the UA renders a native `::-webkit-search-cancel-button` pseudo as soon as the field has text, so users see two overlapping clear controls.
- Hide the UA pseudo via a single rule in `src/app/globals.css` so the custom button is the sole affordance. A global rule (rather than per-component) means future `type="search"` inputs inherit the fix automatically.
- Add a Playwright spec asserting `getComputedStyle(input, "::-webkit-search-cancel-button").display === "none"` after typing, skipped on Firefox where the pseudo does not exist.

## Plan

No plan file existed; the issue (`#411`) is small enough that a plan file would be over-engineering. The diff is 13 lines of CSS + 33 lines of test.

## Test plan

- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — all clean
- [x] `pnpm test` — 153 files / 2524 tests pass (no regressions)
- [ ] **CI:** `pnpm test:e2e -- --project=chromium agenda-search` — the new spec asserts the pseudo is hidden in Chromium. Skipped on Firefox by design.
- [ ] **Manual (per issue acceptance criteria):**
  - [ ] Open `/test/calendar?view=agenda` in Chrome — type a query, confirm only the custom `×` clear control is visible.
  - [ ] Same in Safari.
  - [ ] Firefox is unaffected (pseudo doesn't exist) — sanity-check no regression.
  - [ ] Once PR #391 (`AgendaList`) merges, repeat on `/calendar?view=day&agenda=1` and `/calendar?view=week&agenda=1` — the global rule covers both renderers.

## Environment note

Playwright browser downloads are blocked by this remote env's network policy (`cdn.playwright.dev` not in egress allowlist), so the new spec was not exercised in this session — verification deferred to CI and a manual local pass.

Closes #411
Closes #412

---
_Generated by [Claude Code](https://claude.ai/code/session_014ZZp6nHqrcMcuUmDz1CNsu)_


e2e/agenda-search.spec.ts
```diff
@@ -80,6 +80,39 @@ test.describe("Agenda Calendar — Search (family scenario)", () => {
...
+    await expect(page.getByText("Upcoming Events")).toBeVisible();
+
+    const input = page.getByTestId("agenda-search-input");
+    await input.fill("standup");
+
+    const cancelButtonDisplay = await input.evaluate(
+      (el) =>
+        window.getComputedStyle(el, "::-webkit-search-cancel-button").display
+    );
+    expect(cancelButtonDisplay).toBe("none");
```
BenSeymourODB, line 125: Reliability concern (raised by Sonnet first-pass review):

On the `webkit` Playwright project, `window.getComputedStyle(el, '::-webkit-search-cancel-button').display` can return `""` (empty string) in **both** the before-fix and after-fix states because WebKit's JS reflection of non-standard pseudo-elements is inconsistent across versions. The assertion `toBe('none')` will therefore be a no-op on `webkit` while still correctly catching regressions on `chromium` (which returns `""` before the rule and `"none"` after).

The current `test.skip(browserName === 'firefox', …)` leaves the test running on webkit, where a false-green is possible. Recommend tightening the skip to `browserName !== 'chromium'` so the assertion only runs where it is provably reliable.

---

BenSeymourODB (commented):
First-pass review (Sonnet subagent). One minor reliability nit on the new Playwright spec; otherwise merge-ready.

---

e2e/agenda-search.spec.ts
BenSeymourODB: Fixed in `02fb090`. Tightened the skip from `browserName === 'firefox'` to `browserName !== 'chromium'`, so the assertion now runs only on the chromium project (and skips `webkit`, `mobile-chrome`'s parent profile is also chromium so it still runs, and `tablet`/`firefox` are skipped). Added an inline comment explaining why webkit is excluded.

The CSS rule itself is still global, so WebKit/Safari users still get the visual fix at runtime — only the automated regression guard is scoped down to the engine where `getComputedStyle` on the pseudo is provably reliable.

---

BenSeymourODB (commented):
Second-pass review (Sonnet subagent + Opus driver). The merge from `main` (commit `c97d50b`) clears the CI failure — the `calendar-helpers.test.ts` TS2554 errors that the 2026-06-22 merge inherited were resolved on `main` between then and now, so a plain merge of the latest `main` drops them. CI is re-running on `c97d50b` now.

**Verified locally on the merged branch:**
- `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (5 pre-existing warnings unchanged in unrelated files).
- `pnpm test` — 160 files / 2708 tests passing.
- `pnpm build` — production build succeeds.
- Playwright spec (`e2e/agenda-search.spec.ts:93-127`) could not be exercised locally — `cdn.playwright.dev` is still blocked by this remote env's egress allowlist (same constraint the original session ran into). Verification deferred to CI.

**Sonnet review findings — all minor, no change needed in this PR:**

1. **`src/app/globals.css:316-320` (CSS rule)** — Selector `input[type="search"]::-webkit-search-cancel-button` matches the JSX in both `AgendaCalendar.tsx:372` and `AgendaList.tsx:244`. The unprefixed `appearance: none` on line 318 has no practical effect on a webkit-specific pseudo-element — it's defensive layering. Harmless; not worth churning the diff for.
2. **`e2e/agenda-search.spec.ts:115`** — The spec navigates to `/test/calendar?view=agenda` (AgendaCalendar). Since the CSS rule is global and `AgendaList` also uses `<input type="search">`, the chromium pass on AgendaCalendar is a fair proxy that closes both #411 and #412.
3. **`mobile-chrome` Playwright project (Pixel 5 viewport, chromium engine)** — also reports `browserName === "chromium"`, so the assertion runs there too. Same engine, same pseudo-element behavior — expected to pass.
4. **`tablet`** — iPad Pro 11, WebKit engine — correctly skipped by `browserName !== "chromium"`.

**CLAUDE.md compliance check (clean):**
- Standard Tailwind palette only (the CSS rule uses the UA pseudo-element, not a Tailwind color).
- No `useMemo` / `useCallback` / `React.memo`.
- No `any` types.
- No tests weakened. The new spec is the regression guard for #411/#412.

**Other usages of `<input type="search">` in `src/`:** only two — `AgendaCalendar.tsx:372` and `AgendaList.tsx:244`. Both are paired with a custom clear `<button>`, so the global rule is safe; no place in the codebase actually wants the UA pseudo to remain.

Ready to merge once CI on `c97d50b` is green.

---

BenSeymourODB (commented):
## In-depth review

🤖 `implement-issue` session — couldn't grab an issue this round (concurrent sessions beat me twice), so spending it on an in-depth review of this PR per the session's "in-depth review of one open PR ≥ 3 days old" fallback.

### Verdict — small spec fix needed before merge (CSS rule is good)

The CSS rule is correct, scoped intentionally, and addresses #411 / #412 in production. The new Playwright spec, however, asserts against selectors and text that **do not exist on the page it navigates to** — the regression guard would always fail when Playwright actually runs it. CI doesn't currently run Playwright (the PR description acknowledges this), which is why the broken spec hasn't been caught. Without a fix, the next person to run the E2E suite will see a false regression.

The spec is one selector-rename away from being correct.

### What the diff does well

- `input[type="search"]::-webkit-search-cancel-button { display: none }` is the canonical fix and the right scope. Placing it in `@layer base` next to the other UA-style resets (scrollbar-hide, grecaptcha-badge) is appropriate.
- Setting `appearance: none; -webkit-appearance: none;` alongside `display: none` is mild belt-and-suspenders (`display: none` alone suffices), but it's a defensive pattern and not a bug. Fine to leave.
- The chromium-only scope is well-justified by the docblock — Firefox doesn't render the pseudo, and WebKit's `getComputedStyle` returns `""` in both pre- and post-fix states (the fix the prior review thread requested in `02fb090`).
- Asserting BOTH `cancelButtonDisplay === "none"` AND `agenda-search-clear` visibility together is good — it confirms the rule didn't accidentally hide the custom affordance.

### Blocker — spec navigates to a page that no longer renders the asserted UI

`await page.goto("/test/calendar?events=default&view=agenda")` mounts `DayCalendar` → `AgendaList`, **not** `AgendaCalendar`. This was the #287 / #150 migration that the same file's own header comment (lines 3–11) already calls out:

> "Issue #287 replaced the /test/calendar agenda surface with DayCalendar + AgendaList, which has no on-page search…"

That comment is itself slightly stale (`AgendaList` does in fact render an on-page search input today — see [`AgendaList.tsx` lines 244–263](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/main/src/components/calendar/AgendaList.tsx#L244-L263)), but the structural point holds: the rendered component is `AgendaList`, not `AgendaCalendar`. Concretely, the new spec's assertions don't match what the page renders:

| Assertion | Expected to find | Actually rendered at `/test/calendar?view=agenda` |
|---|---|---|
| `getByText("Upcoming Events")` | The `<h2>` from `AgendaCalendar` ([`AgendaCalendar.tsx:328`](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/main/src/components/calendar/AgendaCalendar.tsx#L328)) | No such heading — `AgendaList` does not render one |
| `getByTestId("agenda-search-input")` | `AgendaCalendar`'s search input ([`AgendaCalendar.tsx:377`](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/main/src/components/calendar/AgendaCalendar.tsx#L377)) | Doesn't exist on this page — `AgendaList` uses `data-testid="agenda-list-search-input"` ([`AgendaList.tsx:249`](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/main/src/components/calendar/AgendaList.tsx#L249)) |
| `getByTestId("agenda-search-clear")` | `AgendaCalendar`'s clear button | Same — `AgendaList` uses `data-testid="agenda-list-search-clear"` |

Cross-checking: `<AgendaCalendar />` has no JSX call sites anywhere under `src/app/` (`rg "<AgendaCalendar"` returns nothing). It's referenced only by its own unit tests (`AgendaCalendar.test.tsx`, `agenda-helpers.test.ts`) and `AgendaList.tsx` (which imports `groupEventsByCategory` / `sortCategoryEntries` helpers from it). The component is effectively dead at runtime; retiring the file lives in #264.

### Recommended fix (3 selector renames, drop the heading check)

```ts
test("hides the native Webkit cancel button so only the custom clear control is visible", async ({
  page,
  browserName,
}) => {
  test.skip(
    browserName !== "chromium",
    "Only chromium reliably reflects display:none on ::-webkit-search-cancel-button via getComputedStyle"
  );

  await page.goto("/test/calendar?events=default&view=agenda");
  // `AgendaList` is what /test/calendar?view=agenda mounts since #287 —
  // wait on its wrapper rather than the "Upcoming Events" header that
  // `AgendaCalendar` (no longer mounted) used to render.
  await expect(page.getByTestId("agenda-list-wrapper")).toBeVisible();

  const input = page.getByTestId("agenda-list-search-input");
  await input.fill("standup");

  const cancelButtonDisplay = await input.evaluate(
    (el) =>
      window.getComputedStyle(el, "::-webkit-search-cancel-button").display
  );
  expect(cancelButtonDisplay).toBe("none");

  // Custom clear control must still appear and function.
  await expect(page.getByTestId("agenda-list-search-clear")).toBeVisible();
});
```

That's a four-line edit:

- `getByText("Upcoming Events")` → `getByTestId("agenda-list-wrapper")` (one line).
- `agenda-search-input` → `agenda-list-search-input` (one line).
- `agenda-search-clear` → `agenda-list-search-clear` (one line).
- Wrapper-vs-heading comment (one line) so the reason for the change is visible.

The CSS rule is global to all `input[type="search"]` so the underlying fix already covers `AgendaList` — only the assertion is broken.

### Non-blocking observations

- **`AgendaCalendar.test.tsx` unit tests still mount the dead component.** The PR doesn't touch them and shouldn't — they belong to #264. Just noting that the unit-test path remains the live regression check for `AgendaCalendar`'s internals.
- **Stale header comment in `agenda-search.spec.ts`** ("AgendaList … has no on-page search") could be cleaned up in a tiny follow-up — `AgendaList` *does* render `agenda-list-search-input` today, so the rationale for the skips is partly stale. Not in scope for this PR.
- **`appearance: none` placement.** With `display: none` first, the subsequent `appearance` / `-webkit-appearance` properties are unreachable. Cosmetic — leave them; they're harmless and document the intent.

### Process notes

- Spec assertion was not caught by CI (`pnpm test` doesn't run Playwright in the GH Actions workflow; the PR description acknowledges this). The blocking signal here is "would this guard catch the regression it claims to" — and answer is no until the selectors line up with what `/test/calendar?view=agenda` actually renders.
- After the spec fix, the PR is otherwise green to merge: `mergeable_state: clean`, last `ci` check `success` on `c97d50b` (2026-06-23 02:02 UTC), no other outstanding review threads.

---

PR #429, opened 2026-06-17
refactor(settings): tighten UserSettingsBusPayload to Partial&lt;UserSettingsData&gt; (#419)

BenSeymourODB:
## Summary

- Extract `UserSettingsData` out of `settings-form.tsx` into a shared `src/types/user-settings.ts` module.
- Redefine `UserSettingsBusPayload` as `Partial<UserSettingsData>` so the bus payload faithfully models every column the form can write — including the three keys previously missing from the hand-maintained subset (`weekStartDay`, `calendarWorkingHoursStart`, `calendarTransitionSpeed`).
- Add a regression test in `src/lib/__tests__/user-settings-bus.test.ts` that exercises round-tripping of those keys; the test would not type-check before this change.

## Plan

Linked plan: `.claude/plans/user-settings-bus-payload-tightening-419.md`

## Behaviour change

None at runtime. The bus already carried any payload via `CustomEvent.detail`; this PR closes the *typing* gap that let `Partial<UserSettingsData>` widen structurally into the narrower `UserSettingsBusPayload`. Subscribers (`useUserSettings` → `pickCalendarFields`) continue to validate individual keys at the consumer boundary, so no consumer change is required.

Adding a new column to `UserSettingsData` now automatically extends the bus contract — no more drift.

## Files

- `src/types/user-settings.ts` *(new)* — shared `UserSettingsData` interface.
- `src/lib/user-settings-bus.ts` — `UserSettingsBusPayload = Partial<UserSettingsData>`; docstring updated.
- `src/components/settings/settings-form.tsx` — drop the local `UserSettingsData`, import from the shared module.
- `src/lib/__tests__/user-settings-bus.test.ts` — round-trip test for the previously-missing keys.

## Test plan

- [x] `pnpm lint:fix` — clean (no new warnings).
- [x] `pnpm format:fix` — clean.
- [x] `pnpm check-types` — clean (would have failed before the widening; this is the type-level acceptance criterion).
- [x] `pnpm test` — 2542 tests pass, 155 files.
- [x] Visual smoke not applicable (type-only refactor; no UI affected).

## Relationship to #418

Independent. #418 only adds a 5-line block to the catch path of `updateSettings`; this PR touches the import and removed `UserSettingsData` block earlier in the same file (different region). If #418 lands first, the rebase is trivial — its new `emitUserSettingsChange(revert)` call accepts the widened payload automatically.

Closes #419

---
_Generated by [Claude Code](https://claude.ai/code/session_0141322NpQ4WzdrDKkLyejqG)_

### DIFF
--- a/src/lib/__tests__/user-settings-bus.test.ts
+++ b/src/lib/__tests__/user-settings-bus.test.ts
@@ -76,4 +76,38 @@ describe("user-settings-bus", () => {
 
     globalThis.window = original;
   });
+
+  /**
+   * Regression coverage for #419. Before tightening, `UserSettingsBusPayload`
+   * was a hand-maintained subset that omitted `weekStartDay`,
+   * `calendarWorkingHoursStart`, and `calendarTransitionSpeed`. The values
+   * still flowed through `CustomEvent.detail` at runtime (and were silently
+   * discarded by `pickCalendarFields`), but the *type* claimed they were
+   * not part of the bus contract — leaving `Partial<UserSettingsData>`
+   * call sites to widen structurally.
+   *
+   * Compile-time assertion: this file would not type-check on `main` before
+   * the bus payload was widened. If the type drifts back to a narrower
+   * subset, `pnpm check-types` flags the literals below as excess
+   * properties.
+   */
+  it("round-trips weekStartDay / calendarWorkingHoursStart / calendarTransitionSpeed", () => {
+    const handler = vi.fn();
+    const unsubscribe = subscribeUserSettings(handler);
+
+    emitUserSettingsChange({
+      weekStartDay: 1,
+      calendarWorkingHoursStart: 8,
+      calendarTransitionSpeed: "slow",
+    });
+
+    expect(handler).toHaveBeenCalledTimes(1);
+    expect(handler).toHaveBeenCalledWith({
+      weekStartDay: 1,
+      calendarWorkingHoursStart: 8,
+      calendarTransitionSpeed: "slow",
+    });
+
+    unsubscribe();
+  });
 });
--- a/src/lib/__tests__/user-settings-bus.test.ts
+++ b/src/lib/__tests__/user-settings-bus.test.ts
@@ -86,10 +86,12 @@ describe("user-settings-bus", () => {
    * not part of the bus contract — leaving `Partial<UserSettingsData>`
    * call sites to widen structurally.
    *
-   * Compile-time assertion: this file would not type-check on `main` before
-   * the bus payload was widened. If the type drifts back to a narrower
-   * subset, `pnpm check-types` flags the literals below as excess
-   * properties.
+   * The real regression gate is `pnpm check-types`: the literals below
+   * fail TS excess-property checking if `UserSettingsBusPayload` ever
+   * drifts back to a narrower subset. The runtime `expect` assertions
+   * are belt-and-braces — `CustomEvent` carries any payload regardless
+   * of type, so they catch only the orthogonal case where the bus
+   * starts mutating its payload (which would be a different regression).
    */
   it("round-trips weekStartDay / calendarWorkingHoursStart / calendarTransitionSpeed", () => {
     const handler = vi.fn();
--- a/src/lib/user-settings-bus.ts
+++ b/src/lib/user-settings-bus.ts
@@ -15,39 +15,23 @@
  * Server-safe: `window` is guarded so the module can be imported from
  * places that load on the server (e.g. shared hooks).
  */
+import type { UserSettingsData } from "@/types/user-settings";
 
 const EVENT_NAME = "user-settings-changed";
 
 /**
- * Closed payload type for bus events. Listing every column in the
- * `UserSettings` Prisma model that any in-tab consumer might care about
- * lets typo'd keys (`timeformat` instead of `timeFormat`) fail at
- * compile time rather than silently no-op at runtime. Add new fields
- * here as additional `UserSettings` columns get wired through to live
- * UI sync.
+ * Payload type for bus events: any partial of the settings shape the form
+ * can write. Derived from the shared `UserSettingsData` so the bus can
+ * never silently drift from the form — adding a column to
+ * `UserSettingsData` automatically extends the bus contract (#419).
  *
- * Value types intentionally mirror the existing `SettingsForm`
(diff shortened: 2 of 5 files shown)


src/lib/user-settings-bus.ts
```diff
@@ -15,39 +15,23 @@
...
- * Closed payload type for bus events. Listing every column in the
- * `UserSettings` Prisma model that any in-tab consumer might care about
- * lets typo'd keys (`timeformat` instead of `timeFormat`) fail at
- * compile time rather than silently no-op at runtime. Add new fields
- * here as additional `UserSettings` columns get wired through to live
- * UI sync.
+ * Payload type for bus events: any partial of the settings shape the form
+ * can write. Derived from the shared `UserSettingsData` so the bus can
+ * never silently drift from the form — adding a column to
+ * `UserSettingsData` automatically extends the bus contract (#419).
```
BenSeymourODB, line 26: [follow-up] The docstring says "adding a column to `UserSettingsData` automatically extends the bus contract" — which is true, but only for the _bus_. The dependency chain runs:

```
Prisma UserSettings model → (manual) → UserSettingsData → (automatic) → UserSettingsBusPayload
```

The Prisma → `UserSettingsData` step is still manual. If a new column is added to the Prisma model without a corresponding field in `UserSettingsData`, it will silently not flow through the bus. The docstring makes the guarantee sound tighter than it is.

Consider rephrasing to something like: "adding a column to `UserSettingsData` automatically extends the bus contract — ensure new Prisma columns are also added here." Not a blocker for this PR (it's a pre-existing human-discipline gap), but the current wording may create false confidence.

---

src/lib/__tests__/user-settings-bus.test.ts
```diff
@@ -76,4 +76,38 @@ describe("user-settings-bus", () => {
...
+   * discarded by `pickCalendarFields`), but the *type* claimed they were
+   * not part of the bus contract — leaving `Partial<UserSettingsData>`
+   * call sites to widen structurally.
+   *
+   * Compile-time assertion: this file would not type-check on `main` before
+   * the bus payload was widened. If the type drifts back to a narrower
+   * subset, `pnpm check-types` flags the literals below as excess
+   * properties.
+   */
+  it("round-trips weekStartDay / calendarWorkingHoursStart / calendarTransitionSpeed", () => {
```
BenSeymourODB, line 96: [nit] The compile-time assertion described in the docblock ("this file would not type-check on `main` before the bus payload was widened") is the real enforcement mechanism here — the runtime round-trip test is mostly redundant because `CustomEvent` carries any payload regardless of type. Both are valuable, but it's worth being clear in the comment that the _type-level_ gate is `pnpm check-types`, not this `expect` assertion.

The test as written is fine and catches real regressions (e.g. if the field were accidentally removed from `UserSettingsData`). Just a phrasing note — no change needed.

---

BenSeymourODB (commented):
## Review — PR #429 (type-only refactor, closes #419)

Overall: the approach is correct, the implementation is clean, and the acceptance criteria are met. Two inline notes plus a couple of observations that don't need code changes.

### Acceptance-criteria check

| Criterion | Status |
|---|---|
| No `Partial<UserSettingsData>` silently widened at emit sites | **Pass** — `settings-form.tsx:132` now types `partial` as `Partial<UserSettingsData>` which is structurally identical to `UserSettingsBusPayload`. No widening gap remains. |
| All `useUserSettings` consumers work without changes | **Pass** — `pickCalendarFields` is untouched; bus payload is a superset so subscriptions see the same or more keys. |
| Emit-site audit | **Pass** — 3 emit sites total: `settings-form.tsx:132`, `useUserSettings.ts:193`, `useUserSettings.ts:214`. All are typed against `UserSettingsPartial`/`Partial<UserSettingsData>`. |

### Is `src/types/user-settings.ts` the right home?

Yes. The alternative (alongside the bus or alongside Prisma generated types) would couple a UI-facing shape to infrastructure. `src/types/` is appropriate for a cross-module interface that isn't tied to any single layer.

### Does `UserCalendarSettings` need deduplication?

No — it's a legitimately different projection. `UserCalendarSettings` (9 fields) is the consumer-side view: it omits `theme`, `dateFormat`, `rewardSystemEnabled`, `defaultTaskPoints`, `showPointsOnCompletion`, and the scheduler fields that `useUserSettings` doesn't own. The bus carries the full shape; each consumer picks what it needs via `pickCalendarFields`. That's the intended layering.

### Rollback cast (`useUserSettings.ts:214`)

`rollback as UserSettingsPartial` casts `Partial<UserCalendarSettings>` to `Partial<UserSettingsData>`. This is pre-existing, untouched by this PR, and currently sound because every key in `UserCalendarSettings` exists in `UserSettingsData`. Worth keeping in mind if `UserCalendarSettings` ever gains a field that isn't in `UserSettingsData` — the cast would silently suppress a type error at that point.

### Settings-form rollback does not re-emit

`settings-form.tsx` reverts local state on PUT failure (line 135–138) but does _not_ re-emit the pre-update snapshot on the bus. `useUserSettings` does re-emit on its PUT failures (line 214). This asymmetry means a form-level failure leaves other in-tab `useUserSettings` consumers with the stale optimistic value. Pre-existing gap, not introduced here — flagging for a follow-up issue.

### Test coverage

The new test is well-targeted and the docblock correctly explains the compile-time enforcement mechanism. Coverage of the three previously-missing keys is the right focus. A negative/typo test wouldn't add value here because typo detection is purely compile-time (excess property checking on object literals) and can't be expressed as a runtime `expect` assertion.

---

src/lib/user-settings-bus.ts
BenSeymourODB: Addressed in `991a4fa` — the docstring now explicitly calls out that the Prisma → `UserSettingsData` step is still manual and only the `UserSettingsData` → bus link is type-enforced. Matches your suggested phrasing intent.

---

src/lib/__tests__/user-settings-bus.test.ts
BenSeymourODB: Rephrased in `991a4fa` — the docblock now states that `pnpm check-types` is the real regression gate (literals fail TS excess-property checking if the bus narrows again) and notes that the runtime `expect` calls cover only the orthogonal case of the bus mutating its payload.

---

PR #432, opened 2026-06-17
fix(settings): suppress same-key rollback when a later PUT has committed (#420)

BenSeymourODB:
## Summary

- `SettingsForm.updateSettings` and `useUserSettings.mutate` both snapshot pre-call values for rollback and re-emit on the user-settings bus on PUT failure. When two calls touch the **same key**, the failing call's rollback wiped a later successful PUT's committed value from local state and every in-tab bus subscriber.
- Switch the rollback to compare-and-swap: read the live state through a `settingsRef` (`SettingsForm` gains the same wiring `useUserSettings` already has) and only revert keys whose current value still equals what *this* call set. Other keys (and other calls' values) are left alone.
- Strict generalization of the existing #363 different-key semantics; single-call failure path (#414) is unchanged.

## Plan

Linked plan: `.claude/plans/user-settings-bus-rollback-race-420.md`

## Test plan

- [x] New `SettingsForm — same-key rollback race (#420)` test: rapid `theme=dark` then `theme=system`, second succeeds first, first fails → bus and local form stay at `system`, no `{ theme: "light" }` emit lands.
- [x] Regression guard in the same describe block: single-call failure still emits the reverted partial (`{ theme: "light" }`) so the #414 contract is intact.
- [x] New `useUserSettings > same-key rollback race (#420)` test: optimistic `24h` mutate held mid-PUT, a same-key bus event flips state back to `12h`, then the PUT fails — no stale rollback emit, no local regression.
- [x] Full suite: `pnpm test` → 2546/2546 passing.

## Deferred to follow-ups

- #433 — direct two-concurrent-`mutate` test at the hook level (this PR uses a same-key bus-emit stand-in inside the hook test; the integration test in `settings-form.test.tsx` covers the full two-mutate flow at the component level).

Closes #420

### DIFF
--- a/.claude/plans/user-settings-bus-rollback-race-420.md
+++ b/.claude/plans/user-settings-bus-rollback-race-420.md
@@ -1,0 +1,100 @@
+# Plan — guard user-settings-bus subscribers against same-key PUT-race rollback (#420)
+
+## Problem
+
+Two `updateSettings` calls in flight against the same key:
+
+1. A: `theme: "dark"`, snapshots `previousPartial = { theme: "light" }`, optimistic state → `dark`.
+2. B: `theme: "system"`, snapshots `previousPartial = { theme: "dark" }` (A's optimistic value), optimistic state → `system`.
+3. B's PUT succeeds first → `emitUserSettingsChange({ theme: "system" })`. Subscribers and form both at `system`.
+4. A's PUT fails → catch path runs `setSettings(curr => ({ ...curr, ...{ theme: "light" } }))` and `emitUserSettingsChange({ theme: "light" })`. Both the form and bus subscribers regress to `light`, even though B's later, successful PUT was the most-recent truth.
+
+The same shape exists at the second emit site in `src/hooks/useUserSettings.ts` (`mutate` catch block, lines 207–216) — `settingsRef` is already wired but the rollback writes `snapshot` unconditionally.
+
+## Approach — compare-and-swap rollback
+
+For each key in the saved `previousPartial`, before reverting:
+
+- Read the current state from a `settingsRef` (mirror of `settings`, updated in `useEffect`).
+- If `settingsRef.current[key] === thisCallOptimisticValue[key]`, this call's optimistic write is still in place → safe to revert.
+- If they differ, a later call has touched the key → leave it alone.
+
+Emit only the keys actually reverted. Rolls back both local state and bus emit in lockstep, fixes both same-key and different-key cases (latter already covered by #363; this is a strict generalization).
+
+Pseudo-code (SettingsForm):
+
+```ts
+const settingsRef = useRef(settings);
+useEffect(() => {
+  settingsRef.current = settings;
+}, [settings]);
+
+const updateSettings = async (partial: Partial<UserSettingsData>) => {
+  const optimistic = partial; // values this call wrote
+  let previousPartial: Partial<UserSettingsData> | undefined;
+  setSettings((curr) => {
+    previousPartial = Object.fromEntries(
+      (Object.keys(partial) as Array<keyof UserSettingsData>).map((k) => [k, curr[k]])
+    ) as Partial<UserSettingsData>;
+    return { ...curr, ...partial };
+  });
+
+  try {
+    /* PUT, throw on !ok, emit(partial) */
+  } catch {
+    toast.error("Failed to save settings");
+    if (!previousPartial) return;
+
+    // Build the actual-revert subset from settingsRef (post-render).
+    const liveRevert: Partial<UserSettingsData> = {};
+    const curr = settingsRef.current;
+    for (const key of Object.keys(previousPartial) as Array<keyof UserSettingsData>) {
+      if (Object.is(curr[key], optimistic[key])) {
+        (liveRevert as Record<string, unknown>)[key] = previousPartial[key];
+      }
+    }
+
+    if (Object.keys(liveRevert).length === 0) return;
+    setSettings((prev) => ({ ...prev, ...liveRevert }));
+    emitUserSettingsChange(liveRevert);
+  }
+};
+```
+
+`useUserSettings.mutate` gets the analogous change — it already has `settingsRef`, just needs the per-key compare-and-swap before the rollback + emit.
+
+## Test plan (TDD)
+
+In `src/components/settings/__tests__/settings-form.test.tsx`:
+
+1. Same-key race (the headline acceptance case):
+   - Click `Dark` (A), then `System` (B) before either resolves.
+   - Resolve B as ok, reject A.
+   - Assert: bus handler observed `{ theme: "dark" }` (B's optimistic emit on success) but NOT `{ theme: "light" }` (A's stale rollback). Local form still shows System checked, not Light.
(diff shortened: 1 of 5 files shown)


BenSeymourODB (commented):
## Independent code review — same-key rollback race fix (#420)

I reviewed this fresh, without the design discussion. Overall the approach is correct and the tests pass for the scenarios described. I found one latent timing issue, one test coverage gap worth documenting, a pre-existing bus-type gap made slightly more visible by this PR, and a few nits. No show-stopper blockers — but item 1 is worth understanding before merging.

---

### Finding 1 — Moderate: `settingsRef` timing gap in `SettingsForm` — the guard can be bypassed if B's render hasn't committed by the time A's catch fires

**File:** `src/components/settings/settings-form.tsx`

The new `settingsRef` is updated in a `useEffect`:
```ts
useEffect(() => {
  settingsRef.current = settings;
}, [settings]);
```

`useEffect` runs *after* React commits and paints. Microtasks (Promise resolutions) drain *before* paint. So the following ordering is possible in production:

1. A click → `setSettings({ theme: "dark" })` → React commits → effect fires → `settingsRef.current = { theme: "dark" }`
2. B click → `setSettings({ theme: "system" })` → React batches a re-render (not yet committed)
3. B's `second.resolve({ ok: true })` microtask fires → `emitUserSettingsChange({ theme: "system" })` — but `settingsRef` hasn't been updated yet (effect hasn't run)
4. A's `first.reject(...)` microtask fires → catch reads `settingsRef.current` → still `{ theme: "dark" }` → `Object.is("dark", "dark")` → guard fires → stale rollback emitted → bug survives

The tests avoid this because `userEvent.click` `await`s a microtask flush, so B's render commits (and effects run) before `second.resolve` / `first.reject` are called back-to-back. In production this window is also narrow (UI interactions are always at least one event loop tick apart), so this is unlikely to be observed in practice. But it means the fix is not hermetically sound — it works by incidental timing, not by design.

`useUserSettings` already has the same `settingsRef = useRef` + `useEffect` pattern. Since the plan consciously adopted it (citing "post-render"), this is an accepted trade-off. But it should be documented explicitly in the comment block so the next reader doesn't assume the ref is always current.

**Suggested addition** to the comment in `settings-form.tsx` above the `settingsRef` block:
```ts
// NOTE: the ref update is async (useEffect fires post-commit). In practice
// all UI-triggered updateSettings calls are separated by at least one event
// loop tick, so the ref is current by the time a concurrent catch fires.
// A fully synchronous same-tick race (e.g. programmatic rapid-fire calls)
// would not be guarded. See the analogous pattern in useUserSettings.ts.
```

---

### Finding 2 — Moderate: `useUserSettings` test simulates B's success via a direct bus emit, not a second `mutate` call — the test is valid but the code path is subtly different

**File:** `src/hooks/__tests__/useUserSettings.test.tsx`, lines ~669–693 (new test)

```ts
// Simulate B's success — a later same-key bus event overrides to "12h",
// standing in for a parallel `mutate({ timeFormat: "12h" })` whose PUT
// resolved before A's.
await act(async () => {
  emitUserSettingsChange({ timeFormat: "12h" });
});
expect(result.current.settings.timeFormat).toBe("12h");
```

This is a legitimate test of the rollback guard: after `act()`, React has committed and `settingsRef.current` is `"12h"`, so A's rejection correctly suppresses the rollback. The test passes and the test comment is honest about what it's standing in for.

However the actual real-world race would be two concurrent `mutate` calls. In that path, B's optimistic `setSettings` and A's rejection may happen in the same microtask batch without an intervening `act()`. A direct test of `mutate(B) without await → reject(A)` would cover this path more precisely and catch any scenario where the two-mutate sequence bypasses the guard.

This is a **coverage gap**, not a false positive — the existing test is good and catches the important half of the race. A follow-up issue or a second test case would close the remaining gap.

---

### Finding 3 — Nit: `emitUserSettingsChange(liveRevert)` causes a double-write in `useUserSettings` consumers (pre-existing, now more visible)

**File:** `src/components/settings/settings-form.tsx` (catch block), `src/hooks/useUserSettings.ts` (catch block)

When the rollback emits on the bus, any `useUserSettings` instance subscribed to the bus applies the same keys a second time. The CAS's `setSettings((curr) => ({ ...curr, ...liveRevert }))` writes the rolled-back value; then the bus subscription's `setSettings((prev) => ({ ...prev, ...picked }))` writes the same values again. This is functionally benign (idempotent) but triggers an extra render per consumer.

This was pre-existing (#414) and is not worsened by this PR — the CAS only narrows which keys get emitted, it doesn't add extra emits. Mentioning for awareness.

---

### Finding 4 — Nit: `calendarWorkingHoursStart` and `weekStartDay` are missing from `UserSettingsBusPayload` but reachable via the CAS cast

**File:** `src/lib/user-settings-bus.ts`, `src/components/settings/settings-form.tsx`

`UserSettingsBusPayload` doesn't include `calendarWorkingHoursStart` or `weekStartDay`, but `UserSettingsData` (in `SettingsForm`) does. The cast `(liveRevert as Record<string, unknown>)[key] = revert[key]` can write those keys into `liveRevert`, which is then passed to `emitUserSettingsChange(liveRevert)`. TypeScript won't catch the mismatch because `liveRevert` is typed as `Partial<UserSettingsData>` and `emitUserSettingsChange` takes `UserSettingsPartial` — this assignment may cause a TypeScript error at the call site, or it may silently widen. At runtime, the bus event will carry the extra key; `pickCalendarFields` in `useUserSettings` handles both fields, so the value propagates correctly. 

This is pre-existing scope (the bus type is acknowledged as needing tightening in the `user-settings-bus.ts` comments), not introduced here. But since this PR adds the cast pattern that bypasses the type check at the assignment site, the call-site type mismatch is worth a look to ensure `pnpm check-types` is actually clean.

---

### Finding 5 — Praise: The single-call regression guard is well-designed

**File:** `src/components/settings/__tests__/settings-form.test.tsx`, new "still emits the rolled-back partial" test

The test correctly verifies that the new CAS guard doesn't suppress rollbacks when no concurrent call has overwritten the key. Using `busHandler.toHaveBeenCalledTimes(1)` is the right assertion here — a bare `toHaveBeenCalledWith` would pass even with an extra emit.

---

### Finding 6 — Praise: `Object.is` is the right comparator for primitives

All current fields in `UserSettingsData` / `UserCalendarSettings` are primitives (string, number, boolean). `Object.is` is the correct comparator — `===` would also work for primitives, but `Object.is` is stricter (`NaN === NaN` where `===` isn't, `-0 !== +0` where `===` is). No current field can hold `NaN` or `-0` in practice, so both work, but `Object.is` is the principled choice.

If any future field is an array or object reference, `Object.is` would silently fail (reference equality). This is called out in the issue prompt — there are no such fields today, and the type system would need to change significantly before one could appear here. Worth a comment in the code to document the assumption.

---

### Summary

| # | Severity | Description |
|---|---|---|
| 1 | Moderate | `settingsRef` async gap — guard relies on UI-event timing, not hermetic. Document the assumption. |
| 2 | Moderate | `useUserSettings` test uses bus-emit to simulate B, not a second `mutate`. Coverage gap, not false positive. |
| 3 | Nit | Double-write on rollback emit (pre-existing). |
| 4 | Nit | `calendarWorkingHoursStart` / `weekStartDay` absent from bus payload type — confirm `check-types` clean. |
| 5 | Praise | Regression guard for single-call failure path is solid. |
| 6 | Praise | `Object.is` is the right comparator; worth a comment that the assumption is primitives-only. |

The core fix is correct and the tests cover the headline acceptance case. Findings 1 and 2 are design-level observations the team should decide whether to address before merging or track as follow-ups.

---

PR #426, opened 2026-06-17
refactor(settings): narrow timeFormat to TTimeFormat (#423)

BenSeymourODB:
## Summary

Pure refactor — no runtime behaviour change. Narrows `UserSettingsData.timeFormat` and `DisplayValues.timeFormat` to the existing `TTimeFormat = "12h" | "24h"` union, mirroring the in-flight `dateFormat: TDateFormat` narrowing in #390 and the long-standing pattern for `weekStartDay` / `calendarTransitionSpeed`.

- Exports a new `isTimeFormat` typeguard from `src/hooks/useUserSettings.ts` that reuses the existing `VALID_TIME_FORMATS` allow-list (no duplication).
- Drops the `as TTimeFormat` cast in `pickCalendarFields` — the typeguard flows the narrowed type through automatically (acceptance criterion).
- Coerces the Prisma `string` at the server-rendered `/settings` boundary via the typeguard, falling back to the schema-default `"12h"` — parallel to the `calendarTransitionSpeed` shape already on the page.
- Retypes `MockUserSettings.timeFormat` so the test fixture stays assignable to the narrower component prop shape without an `as` cast, same as the existing `weekStartDay` / `calendarTransitionSpeed` overrides documented in the fixture.

## Plan

Linked plan: [`.claude/plans/timeformat-narrow-423.md`](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/claude/elegant-newton-ud2bbf/.claude/plans/timeformat-narrow-423.md)
Project: https://github.com/users/BenSeymourODB/projects/1

## Independence from PR #390

#390 narrows the sibling `dateFormat` field. The only files both PRs touch are `interface UserSettingsData` (one-line edit) and the `MockUserSettings` fixture (one-line edit). Whichever lands second has a trivial conflict.

## Deferred follow-ups

Both surfaced in the first-pass review; both intentionally left out of this PR to keep the slice clean.

- `UserSettingsBusPayload.timeFormat` in `src/lib/user-settings-bus.ts:37` is still `string` — tracked by #419 (which covers tightening the whole bus payload to a closed subset of `UserSettings` keys).
- `src/app/api/settings/route.ts:15` duplicates `VALID_TIME_FORMATS` instead of importing from `useUserSettings.ts` — filed as #427.

## Test plan

- [x] `pnpm test` — all 2546 tests pass (155 files).
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean.
[...]

PR #452, opened 2026-06-22
refactor(api): use isTimeFormat typeguard from useUserSettings (#427)

BenSeymourODB:
## Summary

- Drops the duplicated `VALID_TIME_FORMATS = ["12h", "24h"]` array from `src/app/api/settings/route.ts` in favour of the canonical `isTimeFormat` typeguard already exported by `src/hooks/useUserSettings.ts`.
- The PUT handler now calls `isTimeFormat(body.timeFormat)` — same membership check, but the typeguard's own `typeof value === "string"` guard means the previous `body.timeFormat as string` cast is no longer needed.
- **Wire contract is unchanged**: same 400 status and same error body (`"Invalid timeFormat. Must be one of: 12h, 24h"`) for the same invalid input.

## Why

Surfaced by the PR #426 review (issue #427) — the route's local allow-list and the `VALID_TIME_FORMATS` array inside `useUserSettings.ts` were in sync today but maintained in two places. A future "add `24h-padded`" change would have required a coordinated edit; this PR collapses it to one source of truth.

## Plan

No `.claude/plans/` entry — the change is a 2-line production edit + 3 new tests. The original plan for the parent narrowing work lives at `.claude/plans/timeformat-narrow-423.md` (which explicitly flagged this server-side consolidation as out of scope for #423; this PR closes that gap).

## Acceptance criteria (from #427)

- [x] `src/app/api/settings/route.ts` imports the allow-list from `useUserSettings.ts` (no local re-declaration).
- [x] No behaviour change at the API boundary — 400 response and error body identical for the same invalid input.
- [x] Existing `src/app/api/settings/__tests__/*` tests pass unchanged. Added a positive test covering `12h` / `24h` plus an explicit non-string rejection, and tightened the existing negative test to assert the full error string (was `.toContain("Invalid timeFormat")`).
- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types && pnpm test` clean.

## Test plan

- [x] `npx vitest run src/app/api/settings/__tests__/route.test.ts` — 39 / 39 passing (was 36; added 3).
- [x] `pnpm test` — full suite 2662 / 2662 passing.
- [x] `pnpm lint:fix && pnpm format:fix` — no new issues (5 pre-existing warnings unchanged).
- [x] `pnpm build` — production build succeeds; the cross-boundary import from a `"use client"` hook into the server route handler does not regress bundling.

## Deferred follow-ups

- **#453** — Move `isTimeFormat` out of the `"use client"` hook into `src/lib/time-format.ts`. Surfaced in the first-pass review of this PR (the `"use client"` → server-route import works via tree-shaking but is the wrong long-term shape; cleaner pattern already in use for `VALID_DATE_FORMATS` in `src/lib/format-date.ts`). This PR honours the original #427 proposal verbatim; the extraction is the natural next step.

## Out of scope

The same shape applies to `VALID_THEMES` (still inline) and `VALID_DATE_FORMATS` (already in `@/lib/format-date`) in the same file. The issue body explicitly narrows scope to `timeFormat`, so the other allow-lists are tracked separately.

Closes #427

### DIFF
--- a/src/app/api/settings/__tests__/route.test.ts
+++ b/src/app/api/settings/__tests__/route.test.ts
@@ -214,7 +214,46 @@ describe("/api/settings", () => {
       const { status, data } = await parseResponse<ApiErrorResponse>(response);
 
       expect(status).toBe(400);
-      expect(data.error).toContain("Invalid timeFormat");
+      // Locks in the exact error body so the #427 refactor (route ↔
+      // useUserSettings allow-list consolidation) cannot regress the
+      // API surface — the message is part of the wire contract.
+      expect(data.error).toBe("Invalid timeFormat. Must be one of: 12h, 24h");
+    });
+
+    it.each(["12h", "24h"])(
+      "accepts %s as a valid timeFormat",
+      async (timeFormat) => {
+        vi.mocked(getSession).mockResolvedValue(mockSession);
+        const updatedSettings = { ...mockSettings, timeFormat };
+        mockPrisma.userSettings.upsert.mockResolvedValue(updatedSettings);
+
+        const request = createMockRequest("/api/settings", {
+          method: "PUT",
+          body: { timeFormat },
+        });
+
+        const response = await PUT(request);
+        const { status, data } =
+          await parseResponse<typeof updatedSettings>(response);
+
+        expect(status).toBe(200);
+        expect(data.timeFormat).toBe(timeFormat);
+      }
+    );
+
+    it("rejects non-string timeFormat values", async () => {
+      vi.mocked(getSession).mockResolvedValue(mockSession);
+
+      const request = createMockRequest("/api/settings", {
+        method: "PUT",
+        body: { timeFormat: 24 },
+      });
+
+      const response = await PUT(request);
+      const { status, data } = await parseResponse<ApiErrorResponse>(response);
+
+      expect(status).toBe(400);
+      expect(data.error).toBe("Invalid timeFormat. Must be one of: 12h, 24h");
     });
 
     it("validates dateFormat value", async () => {
--- a/src/app/api/settings/route.ts
+++ b/src/app/api/settings/route.ts
@@ -1,3 +1,4 @@
+import { isTimeFormat } from "@/hooks/useUserSettings";
 import {
   ApiError,
   requireUserSession,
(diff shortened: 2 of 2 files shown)


BenSeymourODB (commented):
First-pass review of #452 (refactor: drop duplicated `VALID_TIME_FORMATS`, route to `isTimeFormat`).

**Overall: clean and ready to merge once the draft flag is dropped.** No bugs, no CLAUDE.md violations, no behaviour change. One minor flag and one note below.

---

### What was verified

- **Diff is exactly as described.** One import added, `const VALID_TIME_FORMATS` line removed, one call-site swapped (`VALID_TIME_FORMATS.includes(…as string)` → `isTimeFormat(…)`). No unrelated changes.
- **Wire contract unchanged.** The 400 path, HTTP status, and error string (`"Invalid timeFormat. Must be one of: 12h, 24h"`) are identical before and after.
- **Acceptance criteria from #427 met.** All three checklist items in the issue are satisfied by the diff.
- **The 3 new tests are meaningful, not churn.** The tightened `.toBe()` assertion (was `.toContain()`) locks in the exact error body as a wire contract. The `it.each(["12h", "24h"])` positive test was genuinely absent before (no prior test confirmed the happy path round-trips `timeFormat`). The non-string rejection test (`timeFormat: 24`) proves the `typeof` guard inside `isTimeFormat` does real work that the old `VALID_TIME_FORMATS.includes(… as string)` cast silently skipped — this is the most valuable of the three.
- **No other callers inline `["12h", "24h"]` in production code.** Grep found only `useUserSettings.ts` (the canonical array), `route.ts` (now fixed), and `CalendarProvider.tsx` (string literals used for comparison, not as an allow-list — different concern).
- **`VALID_TIME_FORMATS` inside `useUserSettings.ts` is still unexported** (module-private). The public surface is `isTimeFormat` + `TTimeFormat` only, which is the right shape.

---

### One flag: `"use client"` boundary import

`src/hooks/useUserSettings.ts` starts with `"use client"`. The route handler (`src/app/api/settings/route.ts`) is a Server Component context. Importing `isTimeFormat` from a `"use client"` module into a server module is a known bundling edge case.

The PR description notes that `pnpm build` passes, so Next.js 16's bundler is handling it — it tree-shakes the hook's React/client code and only pulls in the pure function. That said, the cleaner long-term home is `src/lib/time-format.ts` (mirroring how `VALID_DATE_FORMATS` already lives in `src/lib/format-date.ts` and is imported by both the route and the hook). The issue author explicitly scoped this PR to the direct import, so merging as-is is fine; just worth tracking as a follow-up if the boundary causes friction later (e.g. if Jest/Vitest starts requiring special transform config for the `"use client"` directive).

---

### Note on error message hardcoding

`"Invalid timeFormat. Must be one of: 12h, 24h"` is now hardcoded in both `route.ts` (the throw) and the test assertions. If a new format is added, three places need updating: the type union, `VALID_TIME_FORMATS` in the hook, and the error string in the route. That's the same two-place edit as before this PR (previously: the array + the type union), so no regression — just worth noting for the future.

---

**No action required to merge.** Mark as ready for review when CI is green.

---

PR #466, opened 2026-06-24
feat(auth): hash-keyed singleflight for /api/auth/refresh-token (#285)

BenSeymourODB:
## Summary

Closes the dedupe gap left after PR #252:

- The session-callback refresh path (inside `auth()`) has been singleflighted since PR #252.
- The client-driven endpoint `POST /api/auth/refresh-token` still called `refreshGoogleAccessToken` directly, so two concurrent client refreshes for the same user raced — wasting Google rate budget and, if Google rotated the refresh token mid-flight, potentially handing one caller a stale envelope.

This PR adds `getOrStartTokenRefresh` alongside the existing `getOrStartSessionRefresh` in `src/lib/auth/token-refresh-singleflight.ts`. Two sibling `Map`s with disjoint key spaces:

| Path | Key | Return type |
| --- | --- | --- |
| session callback | `google:${providerAccountId}` | `RefreshOutcome` |
| `/api/auth/refresh-token` | `sha256(refreshToken).hex` | `GoogleRefreshedTokens` |

The token-hash key gives per-user dedupe (same token ⇒ same Google account) without a DB lookup. Slot purges in `.finally()` so a hung or failed flight doesn't pin subsequent callers — same discipline as the existing session wrapper.

The route now calls `getOrStartTokenRefresh` instead of `refreshGoogleAccessToken` — a two-line swap. All other behaviour (400 on missing token, 500 on missing OAuth creds, 401 + `requiresReauth: true` on `invalid_grant`/`invalid_request`, upstream status forwarding) is preserved bit-for-bit; errors still propagate via promise rejection so the route's existing `try/catch` owns the HTTP translation.

## Why hash-by-token instead of "DB lookup userId"

The issue body called out two implementation options ("DB or token introspection"). Both are heavier than they need to be:

1. **DB lookup** would add a round-trip per refresh (currently zero DB reads on this route) plus an index on `Account.refresh_token` to query the encrypted column.
2. **Token introspection** doubles the Google call count rather than halving it.
3. **Hash-by-token** has identical dedup semantics — same token ⇒ same Google account by construction — with no DB read, no schema change, and no PII in the key (SHA-256 is one-way).

The plan file (`.claude/plans/api-refresh-token-singleflight-285.md`) documents the trade-off in full.

## Plan

Linked plan: `.claude/plans/api-refresh-token-singleflight-285.md`

## Test plan

- [x] `pnpm test:run src/lib/auth/__tests__/token-refresh-singleflight.test.ts` — 18 cases (8 new for `getOrStartTokenRefresh`).
- [x] `pnpm test:run src/app/api/auth/refresh-token/__tests__/route.test.ts` — adds two new cases (same-token concurrent dedupe, distinct-token non-dedupe) on top of the existing 9 contract cases.
- [x] `pnpm test` — full suite: **2718 tests, 160 files, all green**.
- [x] `pnpm lint:fix` — 0 errors, 5 pre-existing warnings (all in unrelated files).
- [x] `pnpm format:fix` — clean.

New unit cases for `getOrStartTokenRefresh`:
- collapses 5 concurrent calls onto a single upstream call
- does NOT collapse distinct tokens (gated to prove true concurrency, not serial purges)
- delivers the same rejection to every concurrent awaiter
- releases the slot after success / generic rejection / `TimeoutError` (#404 parity)
- `__resetTokenRefreshSingleflightCache()` clears in-flight state for test isolation
- hash isolates near-identical tokens (no collisions for one-character diffs)

## Out of scope (deferred — separate issues exist)

- **Cross-process dedupe** — the in-memory singleflight still allows races across Node processes in a multi-instance deploy. Tracked under #286.
- **Idle-session timeout** — the other half of #216; parent issue stays open.

Closes #285

https://claude.ai/code/session_01U2HfgqRoyNaUm8muURGvZa

---
_Generated by [Claude Code](https://claude.ai/code/session_01U2HfgqRoyNaUm8muURGvZa)_

### DIFF
--- a/.claude/plans/api-refresh-token-singleflight-285.md
+++ b/.claude/plans/api-refresh-token-singleflight-285.md
@@ -12,11 +12,11 @@ This issue closes the gap.
 
 ## Acceptance criteria (from #285)
 
-- [ ] Concurrent calls to `/api/auth/refresh-token` for the same user dedupe to a
+- [x] Concurrent calls to `/api/auth/refresh-token` for the same user dedupe to a
       single upstream Google call.
-- [ ] Failure modes (unknown token, revoked token) handled gracefully — current
+- [x] Failure modes (unknown token, revoked token) handled gracefully — current
       route behaviour preserved bit-for-bit.
-- [ ] Unit tests cover concurrent + failure paths.
+- [x] Unit tests cover concurrent + failure paths.
 
 ## Key choice — what to key the singleflight on
 
@@ -174,10 +174,11 @@ gate/deferred pattern. Cases:
    wrapper case to prove the same `.finally()` discipline.
 7. **Test-only reset clears the cache.**
 
-### `src/app/api/auth/__tests__/refresh-token.route.test.ts` (NEW)
+### `src/app/api/auth/refresh-token/__tests__/route.test.ts` (extend existing)
 
-Integration test for the route at the HTTP boundary. Existing tests for this
-route don't appear to exist; the new test file is small but locks the contract:
+Integration tests at the HTTP boundary. The file already exists from PR #275
+(GOOGLE_CLIENT_ID server-only regression); extend it with the singleflight
+cases below:
 
 1. **400 when `refreshToken` is missing** — route validation preserved.
 2. **500 when OAuth env credentials are missing** — preserved.
--- a/.claude/plans/api-refresh-token-singleflight-285.md
+++ b/.claude/plans/api-refresh-token-singleflight-285.md
@@ -1,0 +1,216 @@
+# `/api/auth/refresh-token` through the singleflight queue (#285)
+
+## Goal
+
+PR #252 added an in-memory singleflight for the **session-callback** refresh path. The
+**client-driven** endpoint `POST /api/auth/refresh-token` still bypasses it, so two
+concurrent client-driven refreshes for the same user race: both POST to Google's
+token endpoint, both pay the rate-limit cost, and either can return a stale token
+to its caller if Google rotates the refresh token mid-flight.
+
+This issue closes the gap.
+
+## Acceptance criteria (from #285)
+
+- [ ] Concurrent calls to `/api/auth/refresh-token` for the same user dedupe to a
+      single upstream Google call.
+- [ ] Failure modes (unknown token, revoked token) handled gracefully — current
+      route behaviour preserved bit-for-bit.
+- [ ] Unit tests cover concurrent + failure paths.
+
+## Key choice — what to key the singleflight on
+
+The issue body suggests "lookup userId from the incoming refresh token (DB or
+token introspection)". I considered three options:
+
+1. **DB lookup by encrypted-refresh-token** — encrypt the incoming plaintext,
+   query `Account.refresh_token`, take `userId` as the key. Cost: a DB round-trip
+   per refresh (currently the route has zero DB reads). Plus a new index on
+   `Account.refresh_token` if the field isn't already indexed. Overkill for what's
+   a dedup concern.
+2. **Token introspection (Google `tokeninfo`)** — defeats the purpose; we'd
+   double the Google call count rather than halve it.
+3. **Hash-by-token (`sha256(refreshToken).hex`)** — same-token-⇒-same-user by
+   construction, so the dedup semantic is identical to keying by userId. Zero DB
+   reads. The hash is one-way, doesn't leak the token if accidentally logged, and
+   has zero PII content. No schema change.
+
+**Chosen: (3) hash-by-token.** It also leaves the two singleflight paths cleanly
+separable: the session-callback wrapper continues to key by
+`google:${providerAccountId}` because it owns DB-write state and needs to dedupe
+per-account regardless of how the token shows up; the client-driven wrapper keys
+by `sha256(refreshToken)` because it's stateless and the token IS the identity
+the caller has.
+
+## Architecture
+
+```
+                            ┌──────────────────────────────────────┐
+                            │  src/lib/auth/                       │
+                            │  token-refresh-singleflight.ts        │
+                            │                                      │
+session callback ─────────▶ │  getOrStartSessionRefresh  (existing)│
+                            │    key: google:${providerAccountId}  │
+                            │    wraps refreshGoogleSessionTokens… │
+                            │    return RefreshOutcome             │
+                            │                                      │
+/api/auth/refresh-token  ─▶ │  getOrStartTokenRefresh   (NEW)      │
+                            │    key: sha256(refreshToken)         │
+                            │    wraps refreshGoogleAccessToken    │
+                            │    return GoogleRefreshedTokens      │
+                            │    rejections propagate              │
+                            └──────────────────────────────────────┘
+```
+
+Two independent `Map<string, Promise<…>>`s in the same module. Each clears its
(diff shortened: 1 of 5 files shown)


src/lib/auth/token-refresh-singleflight.ts
```diff
@@ -64,3 +69,73 @@ export function __resetSessionRefreshSingleflightCache(): void {
...
+  if (existing) return existing;
+
+  const refresh =
+    deps.refreshGoogleAccessToken ?? defaultRefreshGoogleAccessToken;
+  // Slot purges on both fulfilment and rejection — a transient Google outage
+  // must not pin the cache, or the next caller would attach to a long-dead
+  // promise and never see a fresh attempt. Errors are NOT classified here;
+  // the route's existing `catch (error)` block continues to own the
+  // `GoogleTokenRefreshError → HTTP response` translation.
+  const pending = refresh(refreshToken, clientId, clientSecret).finally(() => {
```
BenSeymourODB, line 139: **Correctness: `clientId`/`clientSecret` silently dropped for callers that join an existing flight**

When a second caller arrives with the same `refreshToken` but *different* `clientId` or `clientSecret` (e.g. injected via `deps` in a test or, hypothetically, if env vars are rotated mid-flight), it joins the in-flight promise and its credentials are silently discarded — the first caller's creds are used.

In production this is benign: both callers read from the same `process.env` values, so the creds are always identical. And for the test path, `deps.refreshGoogleAccessToken` is keyed on token equality, not cred equality, so the same observation applies. But the function signature doesn't communicate this invariant; a future caller could reasonably expect their `clientId` to be honoured.

**Not a blocking issue** — production behaviour is correct and the `deps` pattern is test-only. But worth either:
1. Documenting the invariant ("second caller's `clientId`/`clientSecret` are ignored if a flight is already in progress"), or
2. If you want to be defensive, asserting `key !== existing || (clientId === inflightClientId)` in debug/test mode. Option 1 is almost certainly sufficient here.

---

src/app/api/auth/refresh-token/route.ts
```diff
@@ -1,11 +1,15 @@
...
+ * `getOrStartTokenRefresh` (#285) so two simultaneous requests share a single
+ * upstream Google round-trip — important because Google rotates refresh tokens
+ * on occasion, and otherwise the loser of the race would receive a stale
+ * envelope.
  */
-import {
-  GoogleTokenRefreshError,
-  refreshGoogleAccessToken,
-} from "@/lib/auth/refresh-google-token";
+import { GoogleTokenRefreshError } from "@/lib/auth/refresh-google-token";
```
BenSeymourODB, line 11: **Minor: non-string `refreshToken` now throws 500 instead of propagating to Google**

`request.json()` returns `any`, so `refreshToken` is typed `any`. The `!refreshToken` guard (line 20, unchanged) correctly rejects falsy values including `""`, but a truthy non-string — e.g. `{"refreshToken": 123}` — passes the guard and reaches `createHash("sha256").update(refreshToken)` in the new wrapper, which throws `ERR_INVALID_ARG_TYPE` and falls through to the outer `catch` returning 500.

Pre-PR, the same input was coerced to `"123"` by `URLSearchParams` inside `refreshGoogleAccessToken` and sent to Google, which returned `invalid_grant` → 401 `requiresReauth`. So this PR changes the error shape for malformed input from 401 to 500.

The new behaviour is arguably better (fail fast, no wasted Google round-trip), but it is an unintentional behaviour change. Consider tightening the existing guard to also reject non-strings:

```ts
if (!refreshToken || typeof refreshToken !== "string") {
  return NextResponse.json(
    { error: "Refresh token is required" },
    { status: 400 }
  );
}
```

This narrows `refreshToken` to `string` for the rest of the handler, prevents the crash path, and stays consistent with the existing 400 contract for bad input. A test for `{ refreshToken: 123 }` → 400 would lock it in.

---

.claude/plans/api-refresh-token-singleflight-285.md
BenSeymourODB, line 1: **Plan housekeeping: acceptance criteria unchecked + test file path stale**

Two minor inaccuracies to tidy up:

1. **Acceptance criteria are still unchecked.** All three `- [ ]` items are implemented and tested — they should be `- [x]` now that the PR lands.

2. **Test file path in the plan doesn't match reality.** The plan's Tests section heading reads `### src/app/api/auth/__tests__/refresh-token.route.test.ts (NEW)`, but the tests were actually added to the pre-existing file at `src/app/api/auth/refresh-token/__tests__/route.test.ts`. The plan prose lower down correctly says "Route-level tests … already exist", so this is just a stale section heading — but a future reader following the plan to find the test file will be misdirected.

---

BenSeymourODB (commented):
Overall this is a clean, well-reasoned PR. The singleflight semantics are correct, the `.finally()` slot-purge handles all failure modes (success, rejection, timeout, classification-as-outcome), and the test suite is thorough with proper gate/deferred patterns that prove true concurrency rather than serial independence. No `any` types introduced, no manual memoization, no logger calls that expose the hash key or the raw token. The SHA-256 key choice is well justified in both the code comments and the plan.

Three comments below — two are minor issues worth addressing before merge, one is plan housekeeping.

**The one worth acting on:** the `typeof refreshToken !== "string"` gap (see route.ts comment). It introduces a silent behaviour change from the pre-PR path (non-string input now crashes to 500 inside `createHash` instead of reaching Google and getting a 401). The fix is a one-liner in the existing guard plus one test case.

**The other functional note** (singleflight comment on `token-refresh-singleflight.ts`): second caller's `clientId`/`clientSecret` are silently ignored when joining an in-flight slot. This is safe in production (both callers always read the same env vars) but not documented. A JSDoc note is sufficient — no code change needed if you prefer.

**Plan housekeeping** is cosmetic only — feel free to fix in a follow-up or squash it into the PR commit.

---

src/app/api/auth/refresh-token/route.ts
BenSeymourODB: Fixed in a88ff4f. Tightened the existing guard to `!refreshToken || typeof refreshToken !== "string"` (which also narrows the rest of the handler to `string`) and added a `{ refreshToken: 123 }` → 400 case to lock the contract in.

---

src/lib/auth/token-refresh-singleflight.ts
BenSeymourODB: Took option 1 — added a JSDoc block on `getOrStartTokenRefresh` calling out that only `refreshToken` participates in the cache key, and that a joining caller's `clientId` / `clientSecret` / injected `deps.refreshGoogleAccessToken` are silently ignored. Production callers all read from the same `process.env`, so the args are always identical; the doc notes that if a future caller could legitimately pass different credentials, the key would need to include them. a88ff4f.

---

.claude/plans/api-refresh-token-singleflight-285.md
BenSeymourODB: Fixed both in a88ff4f. Ticked the three acceptance criteria and rewrote the stale `### src/app/api/auth/__tests__/refresh-token.route.test.ts (NEW)` heading to `### src/app/api/auth/refresh-token/__tests__/route.test.ts (extend existing)` — which matches where the tests actually live (the file was created in #275 for the GOOGLE_CLIENT_ID server-only regression and extended here).

---

PR #469, opened 2026-06-24
refactor(auth): extract requireGoogleTasksSession across Tasks routes (#441)

### LINKED ISSUE #441: [Code Review] Repeated auth/token/PIN validation patterns duplicated across API routes - src/app/api/
## Finding
**Files:** `src/app/api/tasks/route.ts`, `src/app/api/tasks/[taskId]/route.ts`, `src/app/api/tasks/[taskId]/complete/route.ts`, `src/app/api/auth/account/route.ts`, `src/app/api/profiles/[id]/set-pin/route.ts`, `.../reset-pin/route.ts`, `.../verify-pin/route.ts`
**Category:** separation-of-concerns
**Severity:** Medium
**Effort:** M

## Issue
Several auth/validation patterns are copy-pasted across routes instead of extracted:
- The `session.error === "RefreshTokenError"` check + identical 401 response appears in 5+ task/auth routes.
- Session + Google-Tasks-token acquisition (`requireGoogleTasksAccessToken`) is hand-rolled per task route.
- The three PIN routes (set/reset/verify) duplicate session validation, profile-ownership check, and bcrypt import; `verify-pin` additionally embeds lockout logic (`MAX_FAILED_ATTEMPTS`, `LOCKOUT_DURATION_MS`) directly in the handler.

## Suggestion
- Extend/add `requireAuthenticatedWithGoogleScope` (building on `lib/api/handler.ts`'s `requireUserSession`) to handle session existence + refresh-token-error in one place.
- Add a `withProfileOwnershipCheck` wrapper validating the profile belongs to the session user.
- Move PIN lockout/bcrypt logic into a `pinVerificationService` in `lib/services`; routes become thin delegators. Mirrors the existing `withApiHandler` pattern.

## Context
Found during automated code review on 2026-06-20.

BenSeymourODB:
## Summary

- Extracts the duplicated `getSession → !user → RefreshTokenError → requireGoogleTasksAccessToken` preamble (15+ lines repeated across 4 Google Tasks routes and 1 auth route) into two new helpers in `@/lib/auth/helpers`:
  - `requireAuthenticatedSession()` — session existence + `RefreshTokenError`, throws `AuthError` tagged with the right `requiresReauth` flag.
  - `requireGoogleTasksSession()` — composes the above with the existing `requireGoogleTasksAccessToken`.
- Extends `AuthError` with an explicit `requiresReauth?: boolean` constructor option so callers can disambiguate "plain unauthenticated" (no CTA) from "session needs refresh" (CTA). Existing `AuthError` throw-sites are unchanged — the route catch-block keeps a `status === 401 || 403` fallback when the flag is unset.
- Migrates `/api/tasks` (GET + POST), `/api/tasks/[taskId]` (PATCH), `/api/tasks/[taskId]/complete` (POST), `/api/tasks/lists` (GET), and `/api/auth/account` (GET) onto the helpers — each route sheds 9-13 lines.

Net diff: **5 routes down 9-13 lines each**, 2 new helpers + tests. Existing public contracts (every 401/403/404/200 response shape) preserved.

## Plan

Linked plan: [`.claude/plans/441-google-tasks-session-helper.md`](https://github.com/BenSeymourODB/next-digital-wall-calendar/blob/claude/elegant-newton-3xjtfr/.claude/plans/441-google-tasks-session-helper.md)
Project: <https://github.com/users/BenSeymourODB/projects/1>

Issue #441 calls out **three** duplications. This PR delivers **slice 1 of 3**; the other two are filed as follow-up issues so each can be reviewed independently:

- Slice 2 → #467 (`withProfileOwnershipCheck` wrapper for PIN routes)
- Slice 3 → #468 (`pinVerificationService` for bcrypt + lockout)

## Test plan

- [x] `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (4 pre-existing warnings in unrelated files; one stale warning in `tasks/route.ts` for an unused `GoogleTaskItem` type was cleaned up by the migration)
- [x] `pnpm test` — **2721/2721 pass across 160 files** (no regressions)
- [x] Helper tests added: `requireAuthenticatedSession` happy path + 401/no-session + 401/`RefreshTokenError`; `requireGoogleTasksSession` happy path + each error mode propagation; `AuthError.requiresReauth` shape (default undefined, explicit true/false)
- [x] All route tests pass without assertion changes — only the `vi.mock("@/lib/auth", ...)` factory was updated to wire `requireGoogleTasksSession` through the same `getSession` + `requireGoogleTasksAccessToken` `vi.fn()`s the existing setup already drove

## Behavior preservation

The route catch-block change is observably equivalent for every existing throw-site:

| Failure mode | Today's response | After refactor |
|---|---|---|
| No session | `{ error: "Unauthorized" }`, 401 | `{ error: "Unauthorized" }`, 401 (flag `false` → field omitted) |
| `RefreshTokenError` | `{ error: "Session expired. Please sign in again.", requiresReauth: true }`, 401 | identical |
| Missing Tasks scope | `{ error: "...", requiresReauth: true }`, 403 | identical (catch-block falls back to `status === 403` → `true`) |
| No Google account / no token | `{ error: "...", requiresReauth: true }`, 401 | identical (status fallback) |

The `refresh-token` route's explicit `expect(data.requiresReauth).toBeUndefined()` assertion (for non-401 paths) is preserved because the new catch-block only includes the field in the response body when it's `true`.

## Deferred work (follow-up issues filed)

- #467 — extract `withProfileOwnershipCheck` wrapper for the 3 PIN routes (`set-pin`, `reset-pin`, `verify-pin`)
- #468 — extract `pinVerificationService` (bcrypt + `MAX_FAILED_ATTEMPTS` + `LOCKOUT_DURATION_MS`) out of `verify-pin/route.ts` into `lib/services/`

Both are deliberately out of scope here to keep the auth-helper change isolated from PIN-specific business logic.

Closes #441

https://claude.ai/code/session_017pUX8pk2K3RZDz82EP2Joh

---
_Generated by [Claude Code](https://claude.ai/code/session_017pUX8pk2K3RZDz82EP2Joh)_

### DIFF
--- a/src/app/api/auth/account/route.ts
+++ b/src/app/api/auth/account/route.ts
@@ -2,7 +2,11 @@
  * API endpoint for fetching authenticated user's Google account info
  * Used by client components to display account information
  */
-import { getGoogleAccount, getSession } from "@/lib/auth";
+import {
+  AuthError,
+  getGoogleAccount,
+  requireAuthenticatedSession,
+} from "@/lib/auth";
 import { logger } from "@/lib/logger";
 import { NextResponse } from "next/server";
 
@@ -15,22 +19,8 @@ export interface AccountInfo {
 
 export async function GET() {
   try {
-    const session = await getSession();
-
-    if (!session?.user) {
-      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
-    }
-
-    // Check if token refresh failed
-    if (session.error === "RefreshTokenError") {
-      return NextResponse.json(
-        {
-          error: "Session expired. Please sign in again.",
-          requiresReauth: true,
-        },
-        { status: 401 }
-      );
-    }
+    // Handles session existence + RefreshTokenError in one place (#441).
+    const { session } = await requireAuthenticatedSession();
 
     // Get Google account data
     const googleAccount = await getGoogleAccount();
@@ -60,6 +50,14 @@ export async function GET() {
 
     return NextResponse.json(accountInfo);
   } catch (error) {
+    if (error instanceof AuthError) {
+      const requiresReauth =
+        error.requiresReauth ?? (error.status === 401 || error.status === 403);
+      const body: Record<string, unknown> = { error: error.message };
+      if (requiresReauth) body.requiresReauth = true;
+      return NextResponse.json(body, { status: error.status });
+    }
+
     logger.error(error as Error, {
       endpoint: "/api/auth/account",
       errorType: "fetch_account",
--- a/src/app/api/tasks/route.ts
+++ b/src/app/api/tasks/route.ts
@@ -2,11 +2,7 @@
  * API endpoint for Google Tasks
  * Uses server-side authentication with NextAuth.js
  */
-import {
-  AuthError,
-  getSession,
-  requireGoogleTasksAccessToken,
-} from "@/lib/auth";
+import { AuthError, requireGoogleTasksSession } from "@/lib/auth";
 import { createTask, listTasks } from "@/lib/google/tasks-api";
 import { GoogleTasksApiError } from "@/lib/google/tasks-types";
 import { logger } from "@/lib/logger";
@@ -20,11 +16,6 @@ interface CreateTaskBody {
   due?: string;
 }
 
-interface GoogleTaskItem {
-  id: string;
-  [key: string]: unknown;
-}
-
 /**
  * GET /api/tasks - List tasks from a task list
  * Query params:
@@ -34,25 +25,9 @@ interface GoogleTaskItem {
  */
 export async function GET(request: NextRequest) {
   try {
-    const session = await getSession();
-    if (!session?.user) {
-      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
-    }
-
-    if (session.error === "RefreshTokenError") {
-      return NextResponse.json(
-        {
-          error: "Session expired. Please sign in again.",
-          requiresReauth: true,
-        },
-        { status: 401 }
-      );
-    }
-
-    // Combined scope check + token decryption in a single DB call (#260).
-    // Short-circuits users missing the Tasks scope (#237) before URL
-    // parsing, matching the original assertGoogleTasksScope ordering.
-    const accessToken = await requireGoogleTasksAccessToken(session);
+    // Single helper handles session existence, RefreshTokenError, scope
+    // check, and decrypted-token acquisition (#441).
+    const { session, accessToken } = await requireGoogleTasksSession();
 
     const { searchParams } = new URL(request.url);
     const listId = searchParams.get("listId");
@@ -131,23 +106,7 @@ export async function GET(request: NextRequest) {
  */
 export async function POST(request: NextRequest) {
   try {
-    const session = await getSession();
-    if (!session?.user) {
-      return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
-    }
-
-    if (session.error === "RefreshTokenError") {
-      return NextResponse.json(
-        {
-          error: "Session expired. Please sign in again.",
-          requiresReauth: true,
-        },
-        { status: 401 }
-      );
-    }
-
-    // Combined scope check + token decryption in a single DB call (#260).
-    const accessToken = await requireGoogleTasksAccessToken(session);
+    const { session, accessToken } = await requireGoogleTasksSession();
 
     const body = (await request.json()) as CreateTaskBody;
     const { listId, title, notes, due } = body;
@@ -178,13 +137,13 @@ function handleTasksApiError(
   errorType: "fetch_tasks" | "create_task"
 ): NextResponse {
   if (error instanceof AuthError) {
-    // Both 401 (no/expired session) and 403 (scope missing) are recoverable
-    // by re-authenticating, so the client uses the same `requiresReauth` flow.
-    const requiresReauth = error.status === 401 || error.status === 403;
-    return NextResponse.json(
-      { error: error.message, requiresReauth },
-      { status: error.status }
-    );
+    // Explicit AuthError.requiresReauth wins; otherwise fall back to the
+    // status-based inference so existing throw-sites are unchanged (#441).
+    const requiresReauth =
+      error.requiresReauth ?? (error.status === 401 || error.status === 403);
+    const body: Record<string, unknown> = { error: error.message };
+    if (requiresReauth) body.requiresReauth = true;
+    return NextResponse.json(body, { status: error.status });
   }
 
   if (error instanceof GoogleTasksApiError) {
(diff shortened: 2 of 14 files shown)


src/lib/auth/helpers.ts
```diff
@@ -270,14 +270,90 @@ export async function requireGoogleTasksAccessToken(
...
+ * One Prisma query per call — `getSession()` is the only auth round-trip
+ * outside it.
+ */
+export async function requireGoogleTasksSession(): Promise<{
+  session: Session & { user: { id: string } };
+  accessToken: string;
+}> {
+  const { session } = await requireAuthenticatedSession();
+  const accessToken = await requireGoogleTasksAccessToken(session);
+  return { session, accessToken };
```
BenSeymourODB, line 358: **Inner `RefreshTokenError` guard is now unreachable from this composition path.**

`requireAuthenticatedSession()` (called on line 356) already throws on `RefreshTokenError` before we ever call `requireGoogleTasksAccessToken`. The guard inside `requireGoogleTasksAccessToken` (line 239 of the same file) is dead code when reached through `requireGoogleTasksSession` — it's still correct for callers that invoke `requireGoogleTasksAccessToken` directly (and is the test for it explicitly pins that contract), so removing it would be wrong. But the composition is worth a one-line JSDoc note here so the next auditor traversing throw paths doesn't have to re-derive it.

Non-blocking — pure documentation.

---

src/app/api/tasks/route.ts
```diff
@@ -178,13 +137,13 @@ function handleTasksApiError(
...
-    // by re-authenticating, so the client uses the same `requiresReauth` flow.
-    const requiresReauth = error.status === 401 || error.status === 403;
-    return NextResponse.json(
-      { error: error.message, requiresReauth },
-      { status: error.status }
-    );
+    // Explicit AuthError.requiresReauth wins; otherwise fall back to the
+    // status-based inference so existing throw-sites are unchanged (#441).
+    const requiresReauth =
+      error.requiresReauth ?? (error.status === 401 || error.status === 403);
```
BenSeymourODB, line 143: **Response-shape change for the "no session" 401 is subtly different — `requiresReauth` is absent rather than literal `false`.**

The old route returned `{ error: "Unauthorized" }` (no `requiresReauth` key) directly from `NextResponse.json`. The new path throws `AuthError(401, { requiresReauth: false })`; the catch block (line 137-145) then omits the key because `if (requiresReauth)` is falsy for `false`. Net response is **observably identical** — same body, same status — and the existing tests assert only `data.error === "Unauthorized"` so they stay green.

The one place this difference would matter is a client doing a strict `data.requiresReauth === false` check (rather than `!data.requiresReauth`). A grep over `src/components/` and `src/hooks/` for that pattern would be a quick sanity-check; if any client does the strict form, it would now misclassify a missing-session response as "no re-auth needed" via a different code path. The current client uses the truthy form everywhere I can see, so this is informational.

The same shape change applies across all 5 migrated routes — flagging here once.

---

src/app/api/auth/account/route.ts
```diff
@@ -60,6 +50,14 @@ export async function GET() {
```
BenSeymourODB, line 50: **This catch arm is net-new behavior for `/api/auth/account` — worth keeping in mind for future `getGoogleAccount()` refactors.**

The pre-refactor route had no `AuthError` catch — `RefreshTokenError` was a direct `return` inside the `try`, and `getGoogleAccount()` returns `null` (never throws) when no account is linked. So the only `AuthError` throw-site that reaches this new catch is `requireAuthenticatedSession()`, and it correctly renders the 401 with the right `requiresReauth` flag.

The reason this is worth a comment: if `getGoogleAccount()` is ever refactored to throw `AuthError` (e.g. on decryption failure), the new catch arm here will absorb it. That's probably correct behavior, but it's a non-obvious widening of the catch scope. Worth keeping in mind during future audits.

Non-blocking — current behavior is correct.

---

BenSeymourODB (commented):
**First-pass review (Sonnet sub-agent + Opus driver) — `ready-after-minor-fixes` → fixes pushed in `907b194`.**

I ran a thorough first-pass review across every changed file (full diff, all 14 files, every failure mode traced end-to-end). Verdict was `ready-after-minor-fixes` — the two actionable items are already addressed in the follow-up commit `907b194`:

**Fixed in `907b194`:**
1. **Pin `user.id === ""` edge case** (`src/lib/auth/__tests__/helpers.test.ts`). The new helper guard uses `!session?.user?.id`, which is stricter than the old route guard's `!session?.user`. Added a regression test so a future refactor can't silently relax it to a strict `undefined` comparison and slip an empty-id session through.
2. **Document the guard-order assumption** (`src/lib/auth/helpers.ts` JSDoc on `requireAuthenticatedSession`). The `user.id` check runs before the `RefreshTokenError` check; in practice they never collide because the NextAuth session callback only sets `RefreshTokenError` on sessions with a populated `user.id`. Documented inline so the next auditor doesn't have to re-derive it.

**Verified locally on the merged branch (`907b194`):**
- `pnpm lint:fix && pnpm format:fix && pnpm check-types` — clean (4 pre-existing warnings in unrelated files: `profile-avatar.tsx`, `MockCalendarProvider.tsx`, `account-section.tsx`).
- `pnpm test` — **2722/2722 passing across 160 files** (+1 from the new edge-case test; no regressions).

**Behavior preservation — every failure mode traced:**

| Failure mode | Today's response | After refactor |
|---|---|---|
| No session | `{ error: "Unauthorized" }`, 401 | `{ error: "Unauthorized" }`, 401 (helper sets flag `false` → catch omits the field) |
| `RefreshTokenError` | `{ error: "Session expired...", requiresReauth: true }`, 401 | identical (helper sets flag `true`) |
| Missing Tasks scope | `{ error: "...", requiresReauth: true }`, 403 | identical (no flag on `requireGoogleTasksAccessToken`'s throws → catch falls back to `status === 403`) |
| No Google account / no token | `{ error: "...", requiresReauth: true }`, 401 | identical (status fallback) |
| Upstream Google 401/403 | unchanged — handled separately in the `GoogleTasksApiError` arm | unchanged |

**Inline comments below** are three informational notes flagging non-obvious aspects of the change (dead-but-correct inner guard, field-absent-vs-`false` shape nuance for the no-session 401, widened catch scope on `auth/account`). All non-blocking — verifying for the record. No more code changes required.

**CLAUDE.md compliance:** clean — standard Tailwind palette (N/A here), no `useMemo`/`useCallback`/`React.memo`, no `any` types, no test weakening, JSDoc explains WHY not WHAT.

**Other Tasks-pattern duplications not in this PR's scope:** `grep -rn 'session.error === "RefreshTokenError"' src/app/api/` shows 6 remaining hits in `src/app/api/calendar/` — Calendar uses a different Google token (calendar scope) and is its own duplication pattern, correctly out of scope here.

**Follow-up issues confirmed open:** #467 (`withProfileOwnershipCheck`), #468 (`pinVerificationService`). Both linked from the PR body. Closes #441.

Marking ready-for-review.

---

